PR #18, opened 2018-10-24
DOC: links to dependency module homepages

juhi24:
Listed dependencies in README.rst are now links to homepages of the modules. Just so it's easier to find e.g. installation instructions of these modules.


dnerini (approved):
Very good, thanks for the helpful commit.

PR #22, opened 2018-11-24
Add JSON pystepsrc configuration file

aperezhortal:
Load default parameters from a JSON configuration file.

* All the parameters in config are now included in the pystepsrc file that uses the JSON format.
* The installation instruction where updated to include the correct setup of the pystepsrc file for each platform.
* The config module interface was maintained for backward compatibility, but it returns the parameters in the *pystepsrc** file.
* A JSON schema validator was implemented. If the **pystepsrc** has an error, an exception is raised during the import of the pysteps module. One more dependency was added for this (jsonschema).
* The requirements.txt file was added and the environment.yml was updated with the new dependencies.
* The README.rst , setup.py and MANIFEST.IN files where updated.


dnerini (approved):
This is a truly excellent contribution, thanks @aperezhortal! 
It looks like you took care of all details, including future deprecation warnings for the get_specification method. The documentation is clear and also covers windows users. 
This new configuration approach is reasonably simple and flexible enough to accomodate future developments of pysteps.
I tested the changes and found no issue, the pull request can be merged.

Issue #16, opened 2018-09-17
Specify dependency package versions in environment.yml


pkars:
Currently the conda environment file (environment.yml) does not specify which versions of dependencies should be installed. It would be a good idea to specify package versions at least for required dependencies to prevent issues arising from different versions used.

I suggest we use current (latest) stable versions as our required versions, unless there's a reason to use an older version. As of writing this (Sept. 2018), the following versions are available from conda:

  - git 2.18.0
  - dask 0.19.1
  - numpy 1.15.1 
  - matplotlib 2.2.3
  - opencv 3.4.2
  - pillow 5.2.0
  - pyproj 1.9.5.1
  - scipy 1.1.0

dnerini:
We need to revise the environment.yml before the next release.

For example, now jsmin and attrdict are listed, but these two packages are not available in the anaconda distribution and nees to be installed with pip. Hence, the current version of the environment.yml cannot be used with `conda env create -f environment.yml`.

aperezhortal:
I was testing pysteps in a minimal environment, with out the optional dependencies listed below:

* opencv-python_ (for the Lucas-Kanade optical flow)
* pillow_ (for importing gif data)
* pyproj_ (for cartographic transformations and needed to import data)

The package functionality is quite minimal without these dependencies and the examples scripts provided can not run with out them.

Since many of them are already in the environment.yml, wouldn't be better to consider them as dependencies instead as optional packages? By doing so, the full functionality is available after the installation. 

This is not the case with **dask, toolz and pyfftw **,  they are indeed optional, since they help to improve the computational efficiency.

dnerini:
OK, good idea, I've now listed opencv, pillow and pyproj as main dependencies (commit f157da9e223ba41f154ec238086476e0bd739e7d).

aperezhortal:
Going back to the @pkars suggestion about the version numbers. The **environment.yml** file is used to create a the pysteps environment. When doing so, the anaconda installer will resolve the dependencies of the different packages listed in the **environment.yml** file and it will it install the latest versions available in anaconda channels that are configured.

dnerini:
@aperezhortal 's comment seems to close this issue.

Issue #13, opened 2018-09-03
Use specific exception types instead of generic Exception


pulkkins:
Generic Exception classes are used in many places where a more specific exception type would be appropriate. Find ones from the builtin Python exception classes or implement own.

aperezhortal:
Some of the generic exception were replaced by two custom exception: MissingOptionalDependency and DataModelError.

The rest of the generic exception were replaced by build-in Exceptions.

Commits 28d7d7c25d17843df2c18b4213be51e476ee0381 5100b85fd59b89846b7c150964e8d3786c6f080d 5100b85fd59b89846b7c150964e8d3786c6f080d close this issue.

PR #25, opened 2018-12-06
Remove import checks for dependencies.

aperezhortal:
Since several *optional* packages are now listed as **dependencies**. Therefore, several runtime import checks are not needed anymore (following discussion in issue #16). 

This also removes several generic exceptions, which contributes to solve the issue #13.

### DIFF
--- a/pysteps/io/exporters.py
+++ b/pysteps/io/exporters.py
@@ -56,16 +56,8 @@ implements the following interface:
 
 import numpy as np
 from datetime import datetime
-try:
-    import netCDF4
-    netcdf4_imported = True
-except ImportError:
-    netcdf4_imported = False
-try:
-    import pyproj
-    pyproj_imported = True
-except ImportError:
-    pyproj_imported = False
+import netCDF4
+import pyproj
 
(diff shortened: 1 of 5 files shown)


dnerini:
Thanks @aperezhortal. I perfectly see the point of this pull request. However, I do wonder if import checks can be still useful also in case of listed dependencies. 

An advanced user might want to use only a specific module of pysteps, say the VET. With the current version, he/she just needs to install only a limited set of dependencies and will still be able to import pysteps and use the VET module. Without such checks, the import will result in an error instead.

aperezhortal:
> An advanced user might want to use only a specific module of pysteps, say the VET. 

Good point! 

Another alternative could be create a custom Exception (like MissingOptionalDependency) to move on with issue #13, leaving the import checks as they are now.

dnerini:
Yes, this seems a very reasonable approach. How would you like to proceed? Can you directly adapt this pull request or do we need to make a new one?

aperezhortal:
I will close this request. I can push the changes directly to the repository since the motivation of this pull request was to discuss the idea.

dnerini:
Excellent, many thanks!

PR #31, opened 2019-01-11
Fixed 'referenced before assignment' errors when running ensemble nowcasts

tjniemi:
Hey guys,

I ran into problems when trying to run the run_ensemble_nowcast.py example. 
pysteps crashes in nowcasts/steps.py when computing the noise adjustment factors due to  noise.utils.compute_noise_stddev_adjs(...) calling for undeclared variables. 

Here is a quick fix where I moved filter initialization, decomp_method selection and array size
(M, N) declarations before noise_std_coeffs generation to avoid
'referenced before assignment' errors.


pulkkins:
Thanks for the fix. However, the noise adjustment is still an experimental feature that has not been properly tested. Therefore we do not recommend using it for now.

dnerini:
Thanks @tjniemi! @pulkkins do you want to have a look at tero's fix and merge it into the base branch?

PR #36, opened 2019-02-11
Update BoM dataset importer

cvelascof:
Hi All,
New versions of BoM datasets do not include valid_min, valid_max attributes and therefore a new approach is used here to find xmin,xmax,ymax,ymin.
Also more metadata is extracted from datasets rather to assign expected values, for example, accum_time is now estimated from actual start_time and valid_time in the datasets.
Also added subtitle to rainfall plots to make them the same size than ensemble rainfall plots.
Let me know any inconvenient.

Carlos

### DIFF
--- a/pysteps/io/importers.py
+++ b/pysteps/io/importers.py
@@ -95,6 +95,7 @@ try:
 except ImportError:
     pyproj_imported = False
 
+
 def import_bom_rf3(filename, **kwargs):
     """Import a NetCDF radar rainfall product from the BoM Rainfields3.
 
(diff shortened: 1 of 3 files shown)


dnerini (commented):
Hi @cvelascof 

It's great to see you back to pysteps! You pull request looks perfect, can I just ask you to have a quick look at the default entry for BOM data in our [pystepsrc file](https://github.com/pySTEPS/pysteps/blob/c331a5b50cee32b1db180a5c6687c63eb65809d3/pysteps/pystepsrc#L14)? My guess is that you'll have to clean up a bit that part, too, as this was my attempt to find a structure in your filenames, but honestly I'm not sure that I fully understood it... thanks!

cvelascof:
The pystepsrc file looks OK for the BoM datasets on the pysteps/data. The changes here may work as well with these datasets pysteps/data and the newer versions of BoM netcdfs.

cvelascof:
new commit to solve a problem with SSEPS. Now assigning num_workers as per in STEPS code, and also fixing a confusion of the extrapolation method name variable and the actual extrapolation method function variable.

PR #59, opened 2019-03-19
Add codecov to store coverage results

cvelascof:
https://codecov.io is a service to store and visualize coverage results. It is free for open-source projects as pysteps. 
@dnerini please give it a visit and if happy connect the pysteps to codecov


dnerini:
Thanks @cvelascof, I wasn't aware of codecov, it seems like a very helpful tool!

PR #62, opened 2019-03-20
New detcatscores

dnerini:
Reorganization of the detcatscores module so that it uses the same interface as other verification modules (init, accum, compute).

New option to specify the axis along which the scores have to be accumulated in order to allow the computation of scores over the whole domain.


loforest (approved):
This is a great contribution! 
The next step would be to do the same also for the continuous verification scores. This way we could iteratively update the forecast error fields, as in Foresti and Seed (2015). An open question is whether we need to update the mean error fields online or if we should simply accumulate the errors and divide by N offline at the end. The latter option seems simpler.

PR #64, opened 2019-03-22
New detcontscores

dnerini:
* Use same _init, _accum, _compute interfaces as in the other verification methods.
* Implement online update of error mean, variance and covariances.
* Ignore NaNs (also for detcatscores).
* Return results as a dictionary instead of as a list.
* Include argument axis.
* Include argument conditioning.
* Distinguish between online and offline scores.
* Update test suite.


pysteps/verification/detcontscores.py
```diff
@@ -111,64 +409,124 @@ def get_iterable(x):
...
-        # pearson corr
-        if score == 'corr_p':
-            corr_p = pearsonr(pred, obs)[0]
+        if score in ["mae"]:
+            MAE = err["mae"]
+            result.append(MAE)
+
+        # mean squared error
+        if score in ["mse"]:
+            MAE = err["mse"]
```
cvelascof: This line should be 
```suggestion
            MSE = err["mse"]
```

pysteps/verification/detcontscores.py
```diff
@@ -111,64 +409,124 @@ def get_iterable(x):
...
+            result.append(MAE)
```
cvelascof:
```suggestion
            result.append(MSE)
```

cvelascof (commented):
Great improvement for pySTEPS.
Because conditioning options are new in this branch, @dnerini would you mind to add new tests to check both options.
I added some comments in some lines as well

pysteps/verification/detcontscores.py
cvelascof: Also MSE score needs a test in` test_detcontscores.py`

dnerini:
Thank @cvelascof for the review! 
It's a good point, I will add the new tests as suggested, perhaps using a larger and more realistic data set.

cvelascof (approved):
Thanks @dnerini  for adding the tests.
I think that all this is good to be merged.

PR #66, opened 2019-03-25
Add tests for ensemblestats

cvelascof:
Add tests for mean and excprob in ensemblestats psotprocessing.


dnerini (approved):
This is looking good!

Issue #70, opened 2019-04-03
Using pysteps-data for tests and examples


dnerini:
Currently we have a set of sample data in a separate repository, pysteps-data. This is well motivated by the necessity to keep the main library repository as slim as possible.

The problem is that more and more we need sample data within the library, as in the case of the tests run with Travis or the sphinx-gallery that we are preparing and will be eventually built by  Read the Docs.

To avoid having to include data in the library, we could think of using the approach of wradlib. Fundamentally, wradlib uses a shell script to [install the library through Travis](https://github.com/wradlib/wradlib/blob/2ca43e5a15740fd2a48bbebe22ec2081bab1e6c2/.travis.yml#L48). This script will also clone the wradlib-data repository and set an environment variable WRADLIB_DATA containing the path to the data ([here](https://github.com/wradlib/wradlib/blob/2ca43e5a15740fd2a48bbebe22ec2081bab1e6c2/scripts/install.sh#L77)). Finally, a [util method](https://github.com/wradlib/wradlib/blob/2ca43e5a15740fd2a48bbebe22ec2081bab1e6c2/wradlib/util.py#L714) is used to get the path to data directory when need, as for example when running a test for importing a file ([here](https://github.com/wradlib/wradlib/blob/2ca43e5a15740fd2a48bbebe22ec2081bab1e6c2/wradlib/tests/test_io.py#L71)).

Personally, I think we should adopt something similar so that we can store all our sample data in one place and fetch them only when needed (tests, sphinx-gallery,...).

What do you think?

dnerini:
#71 closes this issue.

PR #71, opened 2019-04-03
Load pysteps-data

dnerini:
* Address #70 
* Follow the same wradlib approach
* Git clone the pysteps-data repository with Travis
* Export the path to the data as PYSTEPS_DATA
* Add utility function to read the PYSTEPS_DATA path
* Update the existing io test routines where necessary
* Remove the sample data


dnerini:
We still need to implement the same approach when building the sphinx-gallery with Read the Docs.

dnerini:
I also wonder whether we could better integrate this approach within the pystepsrc file, by directly setting all the "root_path" variables there after cloning the pysteps-data repository. 
What do you think, @aperezhortal ?

aperezhortal:
@dnerini I like the idea. I pushed an implementation of what you propose. The travis.py script reads the default pystepsrc file, update the root_path, and create the pystepsrc.travis file. Then, on travis, the PYSTEPSRC is set with the full path to the new rc file. Maybe it needs more test, but if it work well, when pysteps is loaded it fetch the configuration from the new rc file.

dnerini:
@aperezhortal This looks great!

It works very nicely with Travis:
``` 
python -c "import pysteps as stp;"
Pysteps configuration file found at: /home/travis/build/pySTEPS/pysteps/pystepsrc.travis
```
And also the test of the bom importer went through!

I will therefore remove the other method to avoid clutter in the library.

Edit: I moved travis.py to scripts/travis.py, since I'll have to create a similar scripts for Read the Docs, too

PR #75, opened 2019-04-05
Add tests for FMI and MCH importers.

cvelascof:


aperezhortal:
@cvelascof ,  under OSX, the tests were giving errors when asserting equality for float numbers. I pushed a new helper function that check for equality within a precision or by equality if needed (for strings for example) and modified the tests_io* files to use that function.

dnerini (approved):

cvelascof:
thanks

PR #79, opened 2019-04-15
Add tests for motion module

aperezhortal:
Test the convergence to the actual solution of the optical flow method used.

Test that  inputs with no precipitation give a zero motion field.

Add examples to the user guide where the convergence of the optical flow methods is tested using idealized motion fields.

Fix bug with vet mask. When masked arrays where used as input, the original mask was modified unintentionally.


examples/optical_flow_methods_convergence.py
```diff
@@ -0,0 +1,363 @@
...
+
+    cmap = get_cmap('jet')
+    cmap.set_under('grey', alpha=0.25)
+    cmap.set_over('none')
+
+    # Compare retrieved motion field with the ideal one
+    plt.figure(figsize=(9, 4))
+    ax = plot_precip_field(input_precip, geodata=metadata,
+                           title="Reference motion",
+                           cartopy_subplot=(1, 2, 1), map='cartopy')
```
dnerini: Here `map='cartopy'` causes the doc build to fail, since cartopy is not an actual requirement, but only an optional dependency. I faced a similar problem with other examples. 
A simpler figure without any basemap will fix the issue, but we could also consider installing cartopy when building the documentation.

dnerini: Here `map='cartopy'` causes the doc build to fail, since cartopy is not an actual requirement, but only an optional dependency. I faced a similar problem with other examples. 
A simpler figure without any basemap will fix the issue, but we could also consider installing cartopy when building the documentation.

examples/optical_flow_methods_convergence.py
```diff
@@ -0,0 +1,363 @@
...
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+root_path = data_source["root_path"]
+path_fmt = data_source["path_fmt"]
+fn_pattern = data_source["fn_pattern"]
+fn_ext = data_source["fn_ext"]
+importer_name = data_source["importer"]
+importer_kwargs = data_source["importer_kwargs"]
+
+# Find the input files from the archive
```
dnerini:
```suggestion
# Find the reference field in the archive
```

examples/optical_flow_methods_convergence.py
```diff
@@ -0,0 +1,363 @@
...
+fns = io.archive.find_by_date(date, root_path, path_fmt, fn_pattern, fn_ext,
+                              timestep=5, num_prev_files=0)
+
+# Read the radar composites
```
dnerini:
```suggestion
# Read the reference radar composite
```

examples/optical_flow_methods_convergence.py
```diff
@@ -0,0 +1,363 @@
...
+                                                   metadata,
+                                                   threshold=0.1,
+                                                   zerovalue=-15.0)
+
+print("Precip. pattern shape: " + str(reference_field.shape))
+
+# This suppress nan conversion warnings in plot functions
+reference_field.data[reference_field.mask] = np.nan
+
+plot_precip_field(reference_field, geodata=metadata, title="Reference field")
```
dnerini: This won't work out, since `plot_precip_field`assumes that the input field is in units of mm/h by default. Unfortunately the method doesn't support dBR, so you need to plot the field before the log transformation.

dnerini (changes requested):
Great contribution, @aperezhortal!
The new gallery tutorial is brilliant and sets a much higher standard for our examples ;-)

I just included few suggestions and comments, in particular there are two issues that need to be addressed before merging the PR.

aperezhortal:
Thanks @dnerini. I'm working on your suggestions. I will remove the dependency on cartopy and make plots without a base map. The map is not really important for this example.

I also found an unexpected behavior on the VET tests. The errors vary between different runs of the scripts. My guess is that it is not converging to the solution so I need have to play a little bit more with the vet options (number of iterations or minimization methods). 

**[UPDATE]**: In commit aa1c177, the VET maximum iterations was increased to give more stability to the tests.

examples/optical_flow_methods_convergence.py
aperezhortal: Fix in 8b6eeef

examples/optical_flow_methods_convergence.py
aperezhortal: Fix in 8b6eeef

examples/optical_flow_methods_convergence.py
```diff
@@ -0,0 +1,361 @@
...
+        num_times = 9
+
+    ideal_motion, precip_obs = create_observations(input_precip,
+                                                   motion_type,
+                                                   num_times=num_times)
+
+    oflow_method = motion.get_method(optflow_method_name)
+
+    computed_motion = oflow_method(precip_obs, verbose=False)
+
```
dnerini, line 272: Add additional step to store precip field in mm/h so that we can plot it later.

```suggestion

precip_obs, _ = stp.utils.dB_transform(precip_obs, inverse=True)

```

examples/optical_flow_methods_convergence.py
```diff
@@ -0,0 +1,361 @@
...
+    precip_data, _ = stp.utils.dB_transform(precip_obs.max(axis=0),
```
dnerini:
```suggestion
    precip_data = precip_obs.max(axis=0)
```

examples/optical_flow_methods_convergence.py
```diff
@@ -0,0 +1,361 @@
...
+    cmap.set_over('none')
+
+    # Compare retrieved motion field with the ideal one
+    plt.figure(figsize=(9, 4))
+    plt.subplot(1, 2, 1)
+    ax = plot_precip_field(input_precip, title="Reference motion")
+    quiver(ideal_motion, step=25, ax=ax)
+
+    plt.subplot(1, 2, 2)
+    ax = plot_precip_field(input_precip, title="Retrieved motion")
```
dnerini:
```suggestion
    ax = plot_precip_field(precip_obs[0], title="Retrieved motion")
```

examples/optical_flow_methods_convergence.py
```diff
@@ -0,0 +1,361 @@
...
+                   & ~precip_obs.mask.any(axis=0))
+
+    cmap = get_cmap('jet')
+    cmap.set_under('grey', alpha=0.25)
+    cmap.set_over('none')
+
+    # Compare retrieved motion field with the ideal one
+    plt.figure(figsize=(9, 4))
+    plt.subplot(1, 2, 1)
+    ax = plot_precip_field(input_precip, title="Reference motion")
```
dnerini:
```suggestion
    ax = plot_precip_field(precip_obs[0], title="Reference motion")
```

examples/optical_flow_methods_convergence.py
```diff
@@ -0,0 +1,361 @@
...
+    ideal_motion, precip_obs = create_observations(input_precip,
+                                                   motion_type,
+                                                   num_times=num_times)
+
+    oflow_method = motion.get_method(optflow_method_name)
+
+    computed_motion = oflow_method(precip_obs, verbose=False)
+
+    precip_data, _ = stp.utils.dB_transform(precip_obs.max(axis=0),
+                                            inverse=True)
```
dnerini:
```suggestion
```

dnerini (changes requested):
Looks great to me, I have just included a last suggestion in order to correctly plot the reference field in the results (it is the same problem you had with the first figure before).

dnerini (approved):

dnerini:
Brilliant!

PR #81, opened 2019-04-17
Add test noise motion

cvelascof:
Add simple test to pysteps.noise.motion. 
Also fixed a bug in initialize_bps when p_par and p_perp are None.


dnerini (approved):
Looks great to me!

PR #82, opened 2019-05-09
Improves management of time variables in bom_importer

cvelascof:
Add some minor improvements to the bom_rf3_importer. Better management of time variables.


cvelascof:
Also it solves latest compilation problems in macOS by using now gcc-9.

PR #88, opened 2019-06-05
Add more details for minimum python version

Fangyh09:
Python3.6 is the minimum python version to support F-string, which needs attention. 
I tried to use python3.5 and failed very late. It will save time to reinstall python3.6 at the beginning.

### DIFF
--- a/doc/source/user_guide/install_pysteps.rst
+++ b/doc/source/user_guide/install_pysteps.rst
@@ -8,7 +8,7 @@ Dependencies
 
 The pysteps package needs the following dependencies
 
-* `python >=3.6 <http://www.python.org/>`_
+* `python >=3.6 (Need F-string support) <http://www.python.org/>`_
 * `attrdict <https://pypi.org/project/attrdict/>`_
 * `jsmin <https://pypi.org/project/jsmin/>`_
 * `jsonschema <https://pypi.org/project/jsonschema/>`_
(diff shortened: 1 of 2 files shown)


aperezhortal:
Hi @Fangyh09, thanks for the PR. The minimum python version was already specified in the documentation. 

> Python3.6 is the minimum python version to support F-string, which needs attention.

Although the f-strings supported by python 3.6 is one of the main reasons why that version is needed, there are other functionalities that can be used during the development of the library  (see https://www.python.org/downloads/release/python-360/).

> I tried to use python3.5 and failed very late. It will save time to reinstall python3.6 at the beginning.

You are right, it will be better if we make clear the python version needed if you are installing from sources or using pip.

I suggest the following suggested changes:

line 11 as it was originally:
* `python >=3.6 <http://www.python.org/>`_

line 91 to:
**IMPORTANT**: installing from source requires python=>=3.6 and numpy already installed.

line 138:
**IMPORTANT**: the installation using pip requires python=>=3.6 already installed.

Sorry for suggesting the changes in the old way. Github does not allow suggestions on not modified  lines.

aperezhortal:
@Fangyh09, no need to close the PR. Y If you want, you can push new commits to your branch (patch-1) with the suggested modifications and then we can merge the PR.

Fangyh09:
Okay~

dnerini:
any update on this pull request, @Fangyh09 ?

Fangyh09:
Nope

dnerini:
:-)
Right, I'll close it for the time being, if you agree.
Feel free to reopen it in case you decide to include @aperezhortal 's suggestions.

Issue #116, opened 2019-08-30
Implement GeoTIFF exporter


pulkkins:
Some users have expressed their interest to read the pysteps output in GeoTIFF format. This is because GeoTIFF is the standard format for GIS applications (e.g. QGIS). Being able to easily import pysteps output into GIS applications would be especially helpful for visualization purposes.

kmuehlbauer:
@pulkkins You might not want to pull in wradlib as a dependency, but maybe your users (eg. me) already have it installed. Then you might get some inspiration from [this wradlib-notebook](https://docs.wradlib.org/en/stable/notebooks/fileio/wradlib_gis_export_example.html). This uses GDAL as another big dependency (which QGIS users normally also have installed). It can output to all available GDAL raster formats.

aperezhortal:
PR #118 closes this issue.

Afshinshafei:
Does somebody know that, how can I import fmi_geotiff dataset? when I use "fmi" it uses the pgm. but how can use geotiff dataset?

PR #118, opened 2019-09-05
Improved exporters

pulkkins:
Several improvements to the exporters module:
- A more general interface that allows saving forecasts into multiple files instead of just one.
- Use output path and output file prefix instead of just output file name.
- Implement a GeoTIFF exporter.
- Modify the NetCDF exporter so that for deterministic forecasts, a three-dimensional array is written and the ens_number dimension is dropped.
- Add documentation about the names of the output files.
- Set n_ens_members to 1 by default (i.e. a deterministic forecast).


dnerini (approved):
Nice contribution!

aperezhortal (approved):

aperezhortal:
Nice work! 
This PR also closes issue #116.

Issue #120, opened 2019-09-09
NetCDF exporter broken with netcdf4-python >= 1.5.1


marack:
The netcdf exporter is throwing an exception when used with recent versions of the netCDF4 python module.  This is due to the following lines (similar for `lat`):

```
lon, lat = pr(x_2d.flatten(), y_2d.flatten(), inverse=True)
...
var_lon[:] = lon
var_lat[:] = lat
```

The `lon` and `lat` variables are flat 1D arrays, but they are being assigned into 2D NetCDF variables.

In earlier versions of netCDF4 this was permissible.  As of version 1.5.1 this behaviour is now illegal and throws an exception.  See https://github.com/Unidata/netcdf4-python/issues/939.

The `lon` and `lat` arrays should be reshaped back to 2D before assignment.

marack:
Note for others, I'm currently working around this by explicitly downgrading the `netcdf4` module to version 1.5.0 (`conda install netcdf=1.5.0`).

RubenImhoff:
I noticed this too yesterday. Downgrading the netCDF4 module is a good quick-and-dirty fix. With netCDF4 < 1.5.1, assigning a 2D array to the netCDF variables works too, right? In that case, I would propose to just fix the code appropriately and thus reshape lat and lon to 2D arrays. This should be a relatively easy fix.

cvelascof:
working on this, and adding a test as well.

aperezhortal:
Hi @marack, PR #122 closes this issue. Feel free to re-open it if you experience any issues.

PR #122, opened 2019-09-10
Add tests to exporters

cvelascof:
Solve #120 and add test to netcdf exporter.
Also it fixes another issues to make more CF complaint the saved netCDFs

### DIFF
--- a/pysteps/tests/test_exporters.py
+++ b/pysteps/tests/test_exporters.py
@@ -55,6 +55,4 @@ def test_io_export_netcdf_one_member_one_time_step():
         export_forecast_dataset(precip[np.newaxis, :], exporter)
         close_forecast_files(exporter)
         # assert if netcdf file was saved and file size is not zero
-        import shutil
-        shutil.copy(file_path, '/home/cvelasco/test_netcdf_pysteps.nc')
         assert (os.path.exists(file_path) and os.path.getsize(file_path) > 0)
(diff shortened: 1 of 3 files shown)


RubenImhoff:
@cvelascof, looks good! Thanks for the quick update.

pysteps/tests/test_exporters.py
```diff
@@ -14,7 +23,36 @@ def test_get_geotiff_filename():
...
+    rel_path = os.path.join("prcp-cscn", "2", "2018", "06", "16")
+    filename = os.path.join(root_path, rel_path,
+                            "2_20180616_100000.prcp-cscn.nc")
+    precip, _, metadata = pysteps.io.import_bom_rf3(filename)
+
+    # save it back to disk
+    with tempfile.TemporaryDirectory() as outpath:
+        outfnprefix = 'test_netcdf_out'
+        file_path = os.path.join(outpath, outfnprefix+'.nc')
+        startdate = pd.to_datetime('2018-06-16T10:00:00')
```
dnerini: can this be done with datetime instead of pandas to avoid having that dependency?

`startdate = datetime.strptime("2018-06-16 10:00:00", "%Y-%m-%d %H:%M:%S")`

dnerini (approved):
Nice work, @cvelascof !
I just included a minor comment, otherwise looks good to go!

pysteps/tests/test_exporters.py
cvelascof: Yes. It makes sense. I have just pushed a new version without pandas as per your suggestion.

PR #127, opened 2019-10-14
Add importer for EUMETSAT SAF Files

cvelascof:
@juanpablosimarro did write an importer for SAF CRRI netcdf files, updated pysteps visualization scripts to display these datasets and create new examples using pySTEPS with Satellite data. 
https://github.com/juanpablosimarro/pysteps/tree/saf
 
This pull request includes only the CRRI importer and an additional test for the CRRI importer.
@juanpablosimarro and I will continue working to add more of his updates into pySTEPS.

This branch passed all tests in Travis for Linux but for some reason MacOS env seems broken at Travis.

### DIFF
--- a/pysteps/io/importers.py
+++ b/pysteps/io/importers.py
@@ -366,13 +366,12 @@ def _import_crri_eu_geodata(filename):
     ymax = ds_rainfall.getncattr("gdal_ygeo_up_left")
     xpixelsize = abs(geotable[1])
     ypixelsize = abs(geotable[5])
-    factor_scale = 1.0
-    geodata["x1"] = xmin * factor_scale
-    geodata["y1"] = ymin * factor_scale
-    geodata["x2"] = xmax * factor_scale
-    geodata["y2"] = ymax * factor_scale
-    geodata["xpixelsize"] = xpixelsize * factor_scale
-    geodata["ypixelsize"] = ypixelsize * factor_scale
+    geodata["x1"] = xmin
+    geodata["y1"] = ymin
+    geodata["x2"] = xmax
+    geodata["y2"] = ymax
+    geodata["xpixelsize"] = xpixelsize
+    geodata["ypixelsize"] = ypixelsize
     geodata["yorigin"] = "upper"
 
     # get the accumulation period
--- a/pysteps/io/importers.py
+++ b/pysteps/io/importers.py
@@ -297,10 +297,11 @@ def import_saf_crri(filename, **kwargs):
     -------
 
     out : tuple
-        A three-element tuple containing the rainfall field in mm/h imported
-        from the CRRI SAF netcdf, the quality field and the metadata. The
-        quality field is currently set to None.
-
+        A three-element tuple containing the rainfall field in mm/h, the quality
+        field and the metadata imported from the CRRI SAF netcdf file.
+        The quality field includes values [1, 2, 4, 8, 16, 24, 32] meaning
+        "nodata", "internal_consistency", "temporal_consistency", "good",
+        "questionable", "bad", and "interpolated", respectively.
     """
     if not NETCDF4_IMPORTED:
         raise MissingOptionalDependency(
@@ -333,9 +334,7 @@ def import_saf_crri(filename, **kwargs):
         idx_x = None
         idx_y = None
 
-    precip = _import_saf_crri_data(filename, idx_x, idx_y)
-
-    # TODO(import_saf_crri): Add missing georeferencing data.
+    precip, quality = _import_saf_crri_data(filename, idx_x, idx_y)
 
     metadata["transform"] = None
     metadata["zerovalue"] = np.nanmin(precip)
@@ -344,7 +343,7 @@ def import_saf_crri(filename, **kwargs):
     else:
         metadata["threshold"] = np.nan
 
-    return precip, None, metadata
+    return precip, quality, metadata
 
 
 def _import_saf_crri_data(filename, idx_x=None, idx_y=None):
@@ -352,14 +351,17 @@ def _import_saf_crri_data(filename, idx_x=None, idx_y=None):
     if "crr_intensity" in ds_rainfall.variables.keys():
         if idx_x is not None:
             data = np.array(ds_rainfall.variables["crr_intensity"][idx_y, idx_x])
+            quality = np.array(ds_rainfall.variables["crr_quality"][idx_y, idx_x])
         else:
             data = np.array(ds_rainfall.variables["crr_intensity"])
+            quality = np.array(ds_rainfall.variables["crr_quality"])
         precipitation = np.where(data == 65535, np.nan, data)
     else:
         precipitation = None
+        quality = None
     ds_rainfall.close()
 
-    return precipitation
+    return precipitation, quality
 
 
 def _import_saf_crri_geodata(filename):
--- a/pysteps/io/importers.py
+++ b/pysteps/io/importers.py
@@ -73,7 +73,7 @@ Available Importers
     :toctree: ../generated/
 
     import_bom_rf3
-    import_crri_eu
+    import_saf_crri
     import_fmi_geotiff
     import_fmi_pgm
     import_mch_gif
@@ -275,7 +275,7 @@ def _import_bom_rf3_geodata(filename):
 
     return geodata
 
-def import_crri_eu(filename, **kwargs):
+def import_saf_crri(filename, **kwargs):
     """Import a NetCDF radar rainfall product from the Convective Rainfall Rate
     Intensity (CRRI) product from the Satellite Application Facilities (SAF).
 
@@ -300,12 +300,12 @@ def import_crri_eu(filename, **kwargs):
             "but it is not installed"
         )
 
-    precip = _import_crri_eu_data(filename)
+    precip = _import_saf_crri_data(filename)
 
-    geodata = _import_crri_eu_geodata(filename)
+    geodata = _import_saf_crri_geodata(filename)
     metadata = geodata
 
-    # TODO(import_crri_eu): Add missing georeferencing data.
+    # TODO(import_saf_crri): Add missing georeferencing data.
 
     metadata["transform"] = None
     metadata["zerovalue"] = np.nanmin(precip)
@@ -336,7 +336,7 @@ def import_crri_eu(filename, **kwargs):
     return precip, None, metadata
 
 
-def _import_crri_eu_data(filename):
+def _import_saf_crri_data(filename):
     ds_rainfall = netCDF4.Dataset(filename)
     if "crr_intensity" in ds_rainfall.variables.keys():
         data = np.array(ds_rainfall.variables["crr_intensity"])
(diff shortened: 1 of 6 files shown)


aperezhortal:
Hi @cvelascof, there was a problem with the brew installation. Running **brew update-reset** and **brew update** before installing gcc-9 solved the problem (commits 
0c3db1e  and 11490fc).

cvelascof:
Fantastic @aperezhortal. Thanks for fixing the macOs env.

juanpablosimarro:
Carlos and Andres, thank you.

pysteps/io/importers.py
```diff
@@ -135,6 +135,119 @@
...
+
+    metadata["transform"] = None
+    metadata["zerovalue"] = np.nanmin(R)
+    if np.any(np.isfinite(R)):
+        metadata["threshold"] = np.nanmin(R[R > np.nanmin(R)])
+    else:
+        metadata["threshold"] = np.nan
+
+    # cut a window, to be defined here and only here
+    if True:
```
aperezhortal: This "if" condition is always executed. Is this there for a future keyword? If it is so, the "if" check may be removed and replaced by a TODO comment.

aperezhortal (changes requested):

juanpablosimarro:
Hello Andres,

Sorry for the delay.

Yes, one possibility is to define some geographical windows here through a
keyword (Europe, East Europe, West Europe, countries, etc). If you agree, I
can set a list of suitable windows here.

The keyword would select the geographical window or, alternatively let the
user to define its own. Nevertheless, those windows would be only valid for
those NETCDF SAF products I have used. The SAF software can produce the
same products with a different geographical extension (the METEOSAT
provides an image of almost one whole hemisphere).

Please tell me what you think.

King regards,

Juan

pysteps/io/importers.py
dnerini: What is the motivation to include the domain clipping in the importer?
Personally I would prefer to keep this sort of data manipulation outside of the importers module. If this was motivated by memory limits, perhaps we could add a general option in the readers module, in order to apply some data processing operations (eg clipping, upscaling,..) while reading the data.

Besides, there is a utils method already implemented in pysteps to clip rasters based on geographical coordinates ([clip_domain()](https://pysteps.readthedocs.io/en/latest/generated/pysteps.utils.dimension.clip_domain.html#pysteps.utils.dimension.clip_domain)). Probably worth to have a look at it to see if it can already perform the operation you need.

juanpablosimarro:
Hello Daniele,

The projection and area of the original NETCDF file is that in the attached
image. It is very inconvenient for spectral methods, because pixels varies
a lot from one to another in shape and size, there are a lot of void space,
etc.

One possibility is the one you mention, that is, to use external tools to
cut and reproject this image to the user desired one. I tried to do this
and I did not find the way. I'm not expert in this kind of software (for
instance https://proj.org), could be you can help me. So, after trying some
options I found this simple, however not optimal: to cut the window in the
import sequence.

Another possibily you point out, is to use clip_domain(). I did know it
exists. Then if I'm not mistaken, the sequence would be: import the whole
NETCDF SAF files, and after that use clip_domain()
<https://pysteps.readthedocs.io/en/latest/generated/pysteps.utils.dimension.clip_domain.html#pysteps.utils.dimension.clip_domain>
before continuing with the calculations. I will try to use it.

Possibly the best option is this second one.

Best regards

Juan

pysteps/io/interface.py
```diff
@@ -42,51 +43,55 @@ def get_method(name, method_type):
...
+        | mch_hdf5     | HDF5 file format used by MeteoSiss (MCH).            |
+        +--------------+------------------------------------------------------+
+        | mch_metranet | metranet files in the MeteoSwiss (MCH) archive       |
+        |              | containing precipitation composites.                 |
+        +--------------+------------------------------------------------------+
+        | opera_hdf5   | ODIM HDF5 file format used by Eumetnet/OPERA.        |
+        +--------------+------------------------------------------------------+
+        | knmi_hdf5    |  HDF5 file format used by KNMI.                      |
+        +--------------+------------------------------------------------------+
+        | crri_eu      |  NetCDF SAF CRRI files                               |
```
dnerini: Maybe we should try to keep this in alphabetical order. 
(can you also adjust the position of the knmi_hdf5 entry, please?)

pysteps/io/importers.py
```diff
@@ -135,6 +135,119 @@
     PYPROJ_IMPORTED = False
 
 
+def import_crri_eu(filename, **kwargs):
```
dnerini: I would prefer to keep an alphabetical order for the importers in this module.

Also, don't forget to add your method's name in the list of available importers in the docstrings at the beginning of the module (lines 75-82).

dnerini (changes requested):
Excellent contribution, thanks!
I just included few minor comments and, as already discussed, I think we should move the windowing outside of the importer.

pysteps/io/importers.py
```diff
@@ -135,6 +135,119 @@
...
+    """Import a NetCDF radar rainfall product from the CRRI SAF.
```
aperezhortal: Can you also add what SAF and CRRI stand form?

dnerini:
> Hello Daniele, The projection and area of the original NETCDF file is that in the attached image. It is very inconvenient for spectral methods, because pixels varies a lot from one to another in shape and size, there are a lot of void space, etc. 

Hi @juanpablosimarro , thanks for your answer. Most of the methods in pysteps assume that pixels in your grid have constant size, hence the need to reproject your data.

> One possibility is the one you mention, that is, to use external tools to cut and reproject this image to the user desired one. I tried to do this and I did not find the way. I'm not expert in this kind of software (for instance https://proj.org), could be you can help me. 

Of course! This PR is probably not the right place to discuss this though. Maybe we can discuss this aspect further together on the pysteps slack.

> So, after trying some options I found this simple, however not optimal: to cut the window in the import sequence. Another possibily you point out, is to use clip_domain(). I did know it exists. Then if I'm not mistaken, the sequence would be: import the whole NETCDF SAF files, and after that use clip_domain() before continuing with the calculations. I will try to use it. Possibly the best option is this second one. Best regards 

I'm not sure to understand how clipping the domain is helping with the reprojection of the data. Perhaps because you focus on a subset of your domain where the pixels are reasonably regular?

Personally, I usually first read the raw data in and then adjust the domain at once with clip_domain(). However, this is not necessarily the most efficient way, particularly if you need to limit memory usage when working with large domains. 

In this case, I would suggest to include the option to apply clip_domain() to each individual raster while looping the timestamps in the read_timeseries() method. 

It is also true that it could be a good idea to directly extract a subset of the raster by leaving the option to provide lat lon bounds to the _import_crri_eu_data()  and then simply using indexes, i.e.

`data = np.array(ds_rainfall.variables["crr_intensity"][idx_lat, idx_lon])`.

This kind of contradicts my comment about removing the windowing from the importer, but I guess it would provide an efficient solution, especially with these large satellite rasters.

pysteps/io/importers.py
```diff
@@ -135,6 +135,119 @@
...
+    px = ds_rainfall.variables['nx'][:]
+    py = ds_rainfall.variables['ny'][:]
+    geotable = ds_rainfall.getncattr('gdal_geotransform_table')
+    xmin = px.min()
+    xmax = px.max()
+    ymin = py.min()
+    ymax = py.max()
+    xpixelsize = abs(geotable[1])
+    ypixelsize = abs(geotable[5])
+    factor_scale = 1.0
```
RubenImhoff: Is this 'factor_scale' a changeble factor? If so, I would recommend to add it as e.g. '**kwargs' in the function (unless it's already somewhere defined and I've missed it..). If not, we won't need this variable at all, I assume.

RubenImhoff (changes requested):

RubenImhoff:
Thanks for the useful contribution by the way!

pulkkins:
I sorted the methods in io/interface.py and io/importers.py in alphabetical order.

pulkkins (approved):

pysteps/io/importers.py
dnerini: I can't find any SAF product named "CRRI", maybe you meant to write simply CRR (Convective Rain Rate)?

edit: uh maybe I got it: it's the CRR_INTENSITY product.. doh!

aperezhortal: Added!

juanpablosimarro: Your are right, inside the same file there are other products, for instance hourly accumulation

Screenshot from 2019-11-08 12-03-51

pysteps/io/importers.py
dnerini: removed in cd52933

dnerini:
After a (long expected) closer look, I realized that the SAF CRRI data are actually already projected, my apologies to @juanpablosimarro. This means that the precipitation data are on a regular grid and, in principle, ready to be used by pysteps!

I added few changes to the pull request, mainly added the option to specify the spatial extent when reading the data, instead of hard-coding it in the importer (99d6b65).

I also fixed the definition of the x1, x2, y1, y2 metadata (7cab798). These were specified as the coordinates of the lower-left and upper-right _pixels_, instead of the lower-left and upper-right _corners_ of the domain. 

From my side, this pull request is now ready to be merged into master.

Edit: the build for this branch is failing because of 4e63e93 (in fact, the build passes for the PR)

dnerini (approved):

pysteps/io/importers.py
dnerini: solved by 99d6b65

RubenImhoff:
Nice work, Daniele, this looks good now! I'm glad we're finally able to add this.

pysteps/io/importers.py
```diff
@@ -314,33 +343,17 @@ def import_saf_crri(filename, **kwargs):
...
+def _import_saf_crri_data(filename, idx_x=None, idx_y=None):
     ds_rainfall = netCDF4.Dataset(filename)
     if "crr_intensity" in ds_rainfall.variables.keys():
-        data = np.array(ds_rainfall.variables["crr_intensity"])
-        precipitation = np.where(data == 65535, np.nan, data)
+        if idx_x is not None:
+            data = np.array(ds_rainfall.variables["crr_intensity"][idx_y, idx_x])
+        else:
+            data = np.array(ds_rainfall.variables["crr_intensity"])        precipitation = np.where(data == 65535, np.nan, data)
+            precipitation = np.where(data == 65535, np.nan, data)
```
RubenImhoff: I've quickly scanned through most of the changes, and I'm not sure if this is correct, but it may also be a result of reading through it online.
Shouldn't it be:

```
if idx_x is not None:
    data = np.array(...)
else:
    data = np.array(...)

precipitation = np.where(...)
```

RubenImhoff (approved):

pysteps/io/importers.py
dnerini: you're absolutely right, the error was fixed by a4352e8

PR #132, opened 2020-01-14
Make use of linspace compatible with NumPy 1.18

dnerini:
https://numpy.org/devdocs/release/1.18.0-notes.html 
> np.linspace parameter num must be an integer. Deprecated in NumPy 1.12. (gh-14620)


dnerini:
Obviously failing because of #131.

PR #133, opened 2020-01-22
Spectral

pulkkins:
This branch contains implementations of S-PROG and STEPS in the spectral domain. These implementations have reduced computation time and memory footprint (~20-40%).

The main changes include:
- cascade.decomposition:
  - add a method for recomposing the cascade
  - add option to normalize the cascade levels
   - make computation of the cascade level statistics (mean and std) optional
  - add options to choose the input and output domains: spatial or spectral
  - in addition, implement option to use compact output in the spectral domain
   - use lowercase names for the field and mask arguments
- cascade.interface:
  - modify get_method so that two functions are returned: one for decomposing and one for recomposing the cascade
- noise.fftgenerators:
  - add option to choose the output domain for the noise: spatial or spectral
- nowcasts.sprog and nowcasts.steps:
  - add option to choose the domain for the computations: spatial or spectral
- nowcasts.utils:
  - remove computation of mean and std from stack_cascades
- timeseries.autoregression:
  - use lowercase names for variables and revise docstrings
- timeseries.correlation:
  - implement computation of correlation coefficients in the spectral domain
- utils.spectral:
  - add methods for computing mean, standard deviation and correlation coefficient in the spectral domain

### DIFF
--- a/pysteps/nowcasts/steps.py
+++ b/pysteps/nowcasts/steps.py
@@ -40,9 +40,9 @@ def forecast(R, V, n_timesteps, n_ens_members=24, n_cascade_levels=6,
              noise_stddev_adj=None, ar_order=2, vel_pert_method="bps",
              conditional=False, probmatching_method="cdf",
              mask_method="incremental", callback=None, return_output=True,
-             seed=None, num_workers=1, fft_method="numpy", extrap_kwargs=None,
-             filter_kwargs=None, noise_kwargs=None, vel_pert_kwargs=None,
-             mask_kwargs=None, measure_time=False):
+             seed=None, num_workers=1, fft_method="numpy", domain="spectral",
+             extrap_kwargs=None, filter_kwargs=None, noise_kwargs=None,
+             vel_pert_kwargs=None, mask_kwargs=None, measure_time=False):
     """Generate a nowcast ensemble by using the Short-Term Ensemble Prediction
     System (STEPS) method.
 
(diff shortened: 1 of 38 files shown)


dnerini:
Impressive work, @spulkkin ! I will definitely look into it and review the pull request, I hope to be able to do it in the coming days.

dnerini:
Why not having the spectral version as default domain, @spulkkin ?

pulkkins:
@dnerini: I would not yet make "spectral" the default implementation. Looking at the S-PROG nowcasts, it produces slightly different results, although the implementations should be theoretically equivalent.

For STEPS, the verification is more difficult, as the random generators work in a different way, and it is not possible to reproduce the results using the same seed number. This could be tried with a very large ensemble (>500 members) to see if they converge to the same distribution.

Also note that the localized noise generators and sseps have not been implemented in the spectral domain. They should throw a NotImplementedError if the user chooses domain="spectral".

dnerini:
OK, I understand, thanks.

> Also note that the localized noise generators and sseps have not been implemented in the spectral domain. They should throw a NotImplementedError if the user chooses domain="spectral".

At the moment the sseps method doesn't accept the "domain" argument. Should I add it together with the NotImplementedError in case "spectral" is passed? Note that the interface to this method is already a bit different to the one of the steps method.

pysteps/nowcasts/steps.py
```diff
@@ -137,6 +137,11 @@ def forecast(R, V, n_timesteps, n_ens_members=24, n_cascade_levels=6,
       A string defining the FFT method to use (see utils.fft.get_method).
       Defaults to 'numpy' for compatibility reasons. If pyFFTW is installed,
       the recommended method is 'pyfftw'.
+    domain : {"spatial", "spectral"}
```
dnerini, line 140: Add reference to publication on spectral implementation of STEPS?

dnerini (approved):
This looks good!

All tests are passing and the documentation also builds without problems:
https://pysteps.readthedocs.io/en/spectral/index.html

We are good to go!

From my side:
- updated the unit tests for the steps and sprog methods using the spectral domain
- added a new test for the sseps method
- fixed few errors caused by changes in the interface of the decomposition methods

pysteps/nowcasts/steps.py
pulkkins: Done.

pulkkins:
> OK, I understand, thanks.
> 
> > Also note that the localized noise generators and sseps have not been implemented in the spectral domain. They should throw a NotImplementedError if the user chooses domain="spectral".
> 
> At the moment the sseps method doesn't accept the "domain" argument. Should I add it together with the NotImplementedError in case "spectral" is passed? Note that the interface to this method is already a bit different to the one of the steps method.

No need to add the "domain" argument if sseps doesn't support this functionality.

loforest:
Congrats for including this development!

PR #134, opened 2020-02-12
Fix for a crash when importing a OPERA HDF5 file

pkars:
The ODIM specification allows for including metadata for data in two different places: either under "/dataset/what" group or under "/dataset/data/what" group. If the metadata is included in the latter location, the first location would be empty. However, pysteps doesn't take this into account and tries to read the metadata from "/dataset/what", resulting in a KeyError.

This fix makes it so that pysteps will ignore the "/dataset/what" group if it exists, but doesn't include "quantity" attribute.


dnerini (approved):
Looks all fine to me, thank you for the contribution!

PR #135, opened 2020-02-12
Test cartopy, fix pyproj, somerc projection

dnerini:
- Update to crs.is_geographic (pyproj >= 2.2.0)
- Add basic tests for plot_precip_field method
- Use equivalent projection for Swiss data

### DIFF
--- a/pysteps/tests/test_plt_cartopy.py
+++ b/pysteps/tests/test_plt_cartopy.py
@@ -1,0 +1,49 @@
+# -*- coding: utf-8 -*-
+
+import pytest
+
+from pysteps.visualization import plot_precip_field
+from pysteps.utils import to_rainrate
+from pysteps.tests.helpers import get_precipitation_fields
+import matplotlib.pyplot as pl
+
+pytest.importorskip("cartopy")
+
+plt_arg_names = ("source", "map", "drawlonlatlines", "lw")
+
+plt_arg_values = [
+    ("mch", "cartopy", False, 0.5),
+    ("mch", "cartopy", True, 1.0),
+    ("bom", "cartopy", True, 0.5),
+    ("fmi", "cartopy", True, 0.5),
+    # TODO: why isn't the knmi projection workin? ("knmi", "cartopy", True, 0.5),
+    ("opera", "cartopy", True, 0.5),
+]
+
+@pytest.mark.parametrize(plt_arg_names, plt_arg_values)
+def test_visualization_plot_precip_field(
+        source, 
+        map, 
+        drawlonlatlines, 
+        lw
+    ):
+
+    field, metadata = get_precipitation_fields(0, 0, True, True, None, source)
+    field = field.squeeze()
+    field, __ = to_rainrate(field, metadata)
+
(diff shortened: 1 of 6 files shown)


pysteps/tests/test_plt_cartopy.py
```diff
@@ -0,0 +1,49 @@
...
+pytest.importorskip("cartopy")
+
+plt_arg_names = ("source", "map", "drawlonlatlines", "lw")
+
+plt_arg_values = [
+    ("mch", "cartopy", False, 0.5),
+    ("mch", "cartopy", True, 1.0),
+    ("bom", "cartopy", True, 0.5),
+    ("fmi", "cartopy", True, 0.5),
+    # TODO: why isn't the knmi projection workin? ("knmi", "cartopy", True, 0.5),
```
dnerini, line 19: @RubenImhoff, tests fail when plotting KNMI data with cartopy, can you maybe have a look into it ?

RubenImhoff: I guess it has to do with the somewhat odd projection of the KNMI data. I'll try to have a look at it soon!

RubenImhoff: I hope to have some time tomorrow! :)

cvelascof: Hi @dnerini. We are here just testing if pysteps can plot. We are not assessing if what it was plotted is what was expected to be plotted, right?

dnerini: Hi @cvelascof, that's right, currently we test only if the plotting functions produce no errors, a sort of first basic test...

RubenImhoff: Problem found, the problem is in the KNMI importer (or better said, the fact that the projection is expecting kilometers instead of meters, while the data is meant to be in meters). However, changing the meters to kilometers (so the pixelsize too) fixes it. I will do a little more testing and then I'll change it tomorrow. Maybe it's better if I do that in a different PR, as it changes some of the tests too etc.?

dnerini: Well done @RubenImhoff !
Yes, it probably makes more sense to open a different PR as you say.
I just wonder about changing the projection definition into km wil cause problems somewhere else, as we always assumed that the unit would be meters. need to think a bit more about that...

dnerini:
I think I'm going to merge this PR as it is, knowing that there will be some additional work to do as some point to have better testing of the visualisation module.

pysteps/tests/test_plt_cartopy.py
RubenImhoff: I was thinking the same.. So far I haven't seen any differences, but I'm not sure enough to affirm that it should work. Option B would be to change the x- and y-values of the metadata in the plotting/plot-testing scripts (should work too).

RubenImhoff (approved):
Nice work, Daniele! I think it looks good. 

As discussed, the KNMI changes will follow in a different PR (that will changes a couple of other scripts too in that case).

PR #139, opened 2020-02-19
KNMI visualization io

RubenImhoff:
Changes to allow for correct plotting with cartopy and basemap. Also the testing scripts were adjusted accordingly. I found a small mistake in the y1- and y2-values of the KNMI-importer, which only impacts the visualization tools (should all be good now).

The KNMI projection works best with km values instead of meters. However, I decided to stick to the pysteps standard of using meters. Hence, the plotting scripts that require geodata have a small adjustment for the KNMI-data now to fix the unit issue.


RubenImhoff:
I used Black to change the codes lines to PEP8, but I saw so many changes that I'm wondering whether I did the right thing. Are you also using Black now to make the code PEP8-compliant? If not, I can still change the code. ;)

dnerini:
Hi @RubenImhoff , many thanks for the PR! I'll review it as soon as possible ;-)

dnerini:
> 
> 
> I used Black to change the codes lines to PEP8, but I saw so many changes that I'm wondering whether I did the right thing. Are you also using Black now to make the code PEP8-compliant? If not, I can still change the code. ;)

I quite like Black for code formatting! But it's probably better next time to include large code reformatting changes from Black in a separate PR, otherwise it's a bit tricky to distinguish the actual changes from simpler code reformatting during the review ;-)

dnerini:
> The KNMI projection works best with km values instead of meters. However, I decided to stick to the pysteps standard of using meters. Hence, the plotting scripts that require geodata have a small adjustment for the KNMI-data now to fix the unit issue.

I was actually wondering whether we actually need to assume that it is always in meters.
I need to check another time, but perhaps we can leave it open to have x and y coordinates in any arbitrary unit, as far as these are consistent among them (and the remaining metadata, e.g. xpixelsize and ypixelsize). The units of the projection should be specified in the proj string, as for example with the [Swiss projection](https://github.com/pySTEPS/pysteps/blob/dce3b8e7acffeffedcdddb7dd3cfaa497f758446/pysteps/io/importers.py#L995). 

If this is true, we can leave the KNMI metadata in km and not have to convert the units when plotting the basemap, which is a bit of a hack...

RubenImhoff:
Haha yes, seeing the amount of changes with Black, it would have been smarter to make two different PRs out of this. Next time! ;) 

If it works with kilometers as units, then that's good too (in fact even better, seeing this 'hack'). I'll leave the PR open for now until we make a decision on this. It should be easily changed if we decide to do it that way.

RubenImhoff:
I'm closing this PR, because we've decided to continue with the kilometer-based approach.

PR #141, opened 2020-03-02
Add datasets module

aperezhortal:
At the moment, the user needs to manually download the example data and create the configuration file. This PR adds helpers functions for those tasks.

These functions are cross-platform since they use modules from the Python standard library. 
In particular, they can be used by new users to quickly set up the pysteps's environment and run the examples easily. 

Also, they can be used to replace the bash scripts used in Travis and tox to create the test environment. 

Since the automatic creation of the pystepsrc file is prone to override existing configuration files, existing files are backed up (up to 5 backups are kept). 

PR changes:
- [x] Add the *datasets* module
- [x] Add function to download pysteps data from github.
- [x] Add function to create a default configuration file pointing to the pysteps data directory.
- [x] Add function to quickly load the example data.
- [x] Use these functions in Travis and tox, replacing the bash scripts. (Note: This will break the pypi's tox tests until a new release with this PR is uploaded to pypi).
- [x] Add helper functions to easily load the example data for each data source (fmi, mch, opera, etc).
- [x] Add the dataset module to the documentation and update the installation instructions.

See the updated documentation [here](http://www.meteo.mcgill.ca/~aperez/_pysteps_datasets/): 
[Edited]


pysteps/datasets.py
```diff
@@ -0,0 +1,353 @@
...
+            os.path.join(pysteps_data_dir, value["root_path"])
+        )
+
+    if config_dir is None:
+        home_dir = os.path.expanduser("~")
+        if os.name == "nt":
+            subdir = "pysteps"
+        else:
+            subdir = ".pysteps"
+        config_dir = os.path.join(home_dir, subdir, file_name)
```
dnerini: This will create a `$HOME/.pysteps/pystepsrc/` directory where to save the new config file, which will have path `$HOME/.pysteps/pystepsrc/pystepsrc`.
By doing so, the new pystepsrc won't be found the next time pysteps is imported, since it will look for a `$HOME/.pysteps/pystepsrc` file instead. 
I suggest changing line 259 into `config_dir = os.path.join(home_dir, subdir)`.

dnerini (changes requested):
Hi @aperezhortal, apologies for taking so long to review your PR.
I went through your new code, tested the branch and tried the new datasets module. Everything works smoothly, I really like the new features and how it simplifies the handling of the example data! 

I only found a minor change that I think it should be implemented before merging into master, otherwise looks good to go! 

And many thanks for the excellent contribution!

doc/source/user_guide/example_data.rst
```diff
@@ -3,17 +3,70 @@
...
+    # By default it will place the configuration file in the
+    # $HOME/.pysteps (unix and Mac OS X) or $USERPROFILE/pysteps (windows).
+    config_file_path = create_default_pystepsrc(pysteps_data_dir_path)
+
+Note that for these changes to take effect you need to restart the python interpreter or
+use the :func:`pysteps.load_config_file` function as follows::
+
+    # Load the new configuration file and replace the default configuration
+    import pysteps
+    pysteps.load_config_file(path_to_my_config_file, verbose=True)
```
dnerini: Just a detail: using `config_file_path` as argument here instead of `path_to_my_config_file` would make a direct and logic link to the code above (line 47)

pysteps/datasets.py
aperezhortal: Solved in 689dc01.

doc/source/user_guide/example_data.rst
aperezhortal: Solved in 77cd9c8.

PR #142, opened 2020-03-05
Use logging instead of print

### LINKED ISSUE #140: Use logging instead of print
Are you interested in migrating all log messages from the simple `print` to the more standard `logging` module?

Given the high number of logs that the outputs are written I think it may be useful to many to have a common method to manage them.

Let me know if you want my contribution to introduce the use of the `logging` module in this fantastic framework!

fox91:
This is the PR with the `logging` module implemented in the `pysteps.nowcasts` module.
if it seems okay I can proceed with the rest of the modules.

Fix #140 when is completed.

### DIFF
--- a/examples/plot_steps_nowcast.py
+++ b/examples/plot_steps_nowcast.py
@@ -8,187 +8,215 @@ radar data.
 
 """
 
-from pylab import *
+import logging
+import sys
+
 from datetime import datetime
-from pprint import pprint
+from pprint import pformat
+
+import matplotlib.pyplot as plt
+import numpy as np
+
+# TODO: actually you need to add logging handler before pysteps import because
+# of the use of logging in the __init__ file.
+logging.basicConfig(
+    format='%(asctime)s.%(msecs)03d %(levelname)s %(name)s: %(message)s',
+    datefmt='%H:%M:%S',
+    level=logging.DEBUG,
+    stream=sys.stderr,
+)
+
 from pysteps import io, nowcasts, rcparams
 from pysteps.motion.lucaskanade import dense_lucaskanade
 from pysteps.postprocessing.ensemblestats import excprob
 from pysteps.utils import conversion, dimension, transformation
 from pysteps.visualization import plot_precip_field
 
-# Set nowcast parameters
-n_ens_members = 20
-n_leadtimes = 6
-seed = 24
 
-###############################################################################
-# Read precipitation field
-# ------------------------
-#
-# First thing, the sequence of Swiss radar composites is imported, converted and
-# transformed into units of dBR.
+logging.getLogger('matplotlib').setLevel(logging.WARNING)
+
+_LOGGER = logging.getLogger(__name__)
 
 
-date = datetime.strptime("201701311200", "%Y%m%d%H%M")
-data_source = "mch"
+def main():
+    # Set nowcast parameters
+    n_ens_members = 20
+    n_leadtimes = 6
+    seed = 24
 
-# Load data source config
-root_path = rcparams.data_sources[data_source]["root_path"]
-path_fmt = rcparams.data_sources[data_source]["path_fmt"]
-fn_pattern = rcparams.data_sources[data_source]["fn_pattern"]
-fn_ext = rcparams.data_sources[data_source]["fn_ext"]
-importer_name = rcparams.data_sources[data_source]["importer"]
-importer_kwargs = rcparams.data_sources[data_source]["importer_kwargs"]
-timestep = rcparams.data_sources[data_source]["timestep"]
-
-# Find the radar files in the archive
-fns = io.find_by_date(
-    date, root_path, path_fmt, fn_pattern, fn_ext, timestep, num_prev_files=2
-)
-
-# Read the data from the archive
-importer = io.get_method(importer_name, "importer")
-R, _, metadata = io.read_timeseries(fns, importer, **importer_kwargs)
-
-# Convert to rain rate
-R, metadata = conversion.to_rainrate(R, metadata)
-
-# Upscale data to 2 km to limit memory usage
-R, metadata = dimension.aggregate_fields_space(R, metadata, 2000)
-
-# Plot the rainfall field
-plot_precip_field(R[-1, :, :], geodata=metadata)
-
-# Log-transform the data to unit of dBR, set the threshold to 0.1 mm/h,
-# set the fill value to -15 dBR
-R, metadata = transformation.dB_transform(R, metadata, threshold=0.1, zerovalue=-15.0)
-
-# Set missing values with the fill value
-R[~np.isfinite(R)] = -15.0
-
-# Nicely print the metadata
-pprint(metadata)
-
-###############################################################################
-# Deterministic nowcast with S-PROG
-# ---------------------------------
-#
-# First, the motiong field is estimated using a local tracking approach based
-# on the Lucas-Kanade optical flow.
-# The motion field can then be used to generate a deterministic nowcast with
-# the S-PROG model, which implements a scale filtering appraoch in order to
-# progressively remove the unpredictable spatial scales during the forecast.
-
-# Estimate the motion field
-V = dense_lucaskanade(R)
-
-# The S-PROG nowcast
-nowcast_method = nowcasts.get_method("sprog")
-R_f = nowcast_method(
-    R[-3:, :, :],
-    V,
-    n_leadtimes,
-    n_cascade_levels=8,
-    R_thr=-10.0,
-    decomp_method="fft",
-    bandpass_filter_method="gaussian",
-    probmatching_method="mean",
-)
-
-# Back-transform to rain rate
-R_f = transformation.dB_transform(R_f, threshold=-10.0, inverse=True)[0]
-
-# Plot the S-PROG forecast
-plot_precip_field(
-    R_f[-1, :, :],
-    geodata=metadata,
-    title="S-PROG (+ %i min)" % (n_leadtimes * timestep),
-)
-
-###############################################################################
-# As we can see from the figure above, the forecast produced by S-PROG is a
-# smooth field. In other words, the forecast variance is lower than the
-# variance of the original observed field.
-# However, certain applications demand that the forecast retain the same
-# statistical properties of the observations. In such cases, the S-PROG
-# forecasts are of limited use and a stochatic approach might be of more
-# interest.
-
-###############################################################################
-# Stochastic nowcast with STEPS
-# -----------------------------
-#
-# The S-PROG approach is extended to include a stochastic term which represents
-# the variance associated to the unpredictable development of precipitation. This
-# approach is known as STEPS (short-term ensemble prediction system).
-
-# The STEPES nowcast
-nowcast_method = nowcasts.get_method("steps")
-R_f = nowcast_method(
-    R[-3:, :, :],
-    V,
-    n_leadtimes,
-    n_ens_members,
-    n_cascade_levels=6,
-    R_thr=-10.0,
-    kmperpixel=2.0,
-    timestep=timestep,
-    decomp_method="fft",
-    bandpass_filter_method="gaussian",
-    noise_method="nonparametric",
-    vel_pert_method="bps",
-    mask_method="incremental",
-    seed=seed,
-)
-
-# Back-transform to rain rates
-R_f = transformation.dB_transform(R_f, threshold=-10.0, inverse=True)[0]
+    ###############################################################################
+    # Read precipitation field
+    # ------------------------
+    #
+    # First thing, the sequence of Swiss radar composites is imported, converted and
+    # transformed into units of dBR.
 
 
-# Plot the ensemble mean
-R_f_mean = np.mean(R_f[:, -1, :, :], axis=0)
-plot_precip_field(
-    R_f_mean,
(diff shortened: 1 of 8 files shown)


pysteps/__init__.py
```diff
@@ -21,6 +21,10 @@
 from . import verification as vf
 from . import visualization as plt
 
+logging.getLogger(__name__).addHandler(logging.NullHandler())
```
aperezhortal, line 24: Maybe we can define the default handler set at INFO level to be equivalent to the print statement. 
That is, no additional formatting. _logger.info(msg) equivalent to print(msg)

That will reduce the verbosity of the outputs.  

We can add additional information (like time, level, etc) for the DEBUG level.

pysteps/__init__.py
```diff
@@ -21,6 +21,10 @@
...
+
+_LOGGER = logging.getLogger(__name__)
```
aperezhortal, line 26: _LOGGER should be lowercase ("_logger").  Uppercase variables normally used for constants.

pysteps/__init__.py
```diff
@@ -149,8 +153,7 @@ def __getitem__(self, key):
         rcparams = json.loads(jsmin(f.read()))
 
     if not rcparams.get("silent_import", False):
-        print("Pysteps configuration file found at: " + params_file
-              + "\n")
+        _LOGGER.debug("Pysteps configuration file found at: %s", params_file)
```
aperezhortal, line 156: Here the logic could be:

If silent_import is True:
    _logger.debug()  
else:
    _logger.info()  # we want the behavior to be equivalent to print. This is important for runtime information.




If silent_import is True,  the

pysteps/nowcasts/extrapolation.py
```diff
@@ -79,7 +81,7 @@ def forecast(precip, velocity, num_timesteps,
 
     if measure_time:
         computation_time = time.time() - start_time
-        print(f"{computation_time:.2f} seconds.")
+        _LOGGER.debug("%.2f seconds.", computation_time)
```
aperezhortal, line 84: Better use INFO level here.

pysteps/nowcasts/sseps.py
```diff
@@ -346,7 +349,7 @@ def forecast(
         }
     )
 
-    print("Estimating nowcast parameters...", end="")
+    _LOGGER.debug("Estimating nowcast parameters...")
```
aperezhortal, line 352: INFO level. This is indicating that the program is running correctly.

pysteps/nowcasts/sseps.py
```diff
@@ -516,9 +519,9 @@ def estimator(R, parsglob=None, idxm=None, idxn=None):
         pars = None
 
     if measure_time:
-        print("%.2f seconds." % (time.time() - starttime))
+        _LOGGER.debug("%.2f seconds.", time.time() - starttime)
```
aperezhortal, line 522: INFO level. This is indicating that the program is running correctly.

pysteps/nowcasts/sseps.py
```diff
@@ -555,14 +558,14 @@ def estimator(R, parsglob=None, idxm=None, idxn=None):
 
     R = R[-1, :, :]
 
-    print("Starting nowcast computation.")
+    _LOGGER.debug("Starting nowcast computation.")
```
aperezhortal, line 561: INFO

pysteps/nowcasts/sseps.py
```diff
@@ -764,9 +767,9 @@ def worker(j):
         res = None
 
         if measure_time:
-            print("%.2f seconds." % (time.time() - starttime))
+            _LOGGER.debug("%.2f seconds.", time.time() - starttime)
```
aperezhortal, line 770: INFO level. This is indicating that the program is running correctly.

examples/plot_steps_nowcast.py
```diff
@@ -8,187 +8,215 @@
 
 """
```
aperezhortal, line 9: Maybe is a good idea to leave the examples untouched before merging the PR. In that way they show the default behavior of the library. 

But this was very useful to see how the logging module worked. :+1:  Perhaps, we can add a new short example showing how the logging module can be used.

examples/plot_steps_nowcast.py
```diff
@@ -8,187 +8,215 @@
...
+        type="prob",
+        units="mm/h",
+        probthr=0.5,
+        title="Exceedence probability (+ %i min)" % (n_leadtimes * timestep),
+    )
+
+    # sphinx_gallery_thumbnail_number = 5
+
+
+if __name__ == '__main__':
```
aperezhortal, line 219: Avoid using if __name__ == 'main' if possible. The examples should run sequentially as a script. 

This will be useful in the future if we use jupyter notebooks for the examples.

pysteps/nowcasts/utils.py
```diff
@@ -13,10 +13,14 @@
...
 
+import logging
+
 import numpy as np
 
+_LOGGER = logging.getLogger(__name__)
+
 
-def print_ar_params(PHI):
+def log_ar_params(PHI):
```
aperezhortal, line 23: Just in case, in addition to the log_ar_params function,  leave the **print_ar_params** for backward compatibility. We can inform to the users with a Deprecation Warning that this function will be removed in a future release.

pysteps/nowcasts/utils.py
```diff
@@ -25,34 +29,34 @@ def print_ar_params(PHI):
...
 
     for k in range(PHI.shape[0]):
-        print(fmt_str % ((k + 1,) + tuple(PHI[k, :])))
-        print(hline_str)
+        _LOGGER.debug(fmt_str, (k + 1,) + tuple(PHI[k, :]))
+        _LOGGER.debug(hline_str)
 
 
-def print_corrcoefs(GAMMA):
+def log_corrcoefs(GAMMA):
```
aperezhortal, line 59: Just in case, in addition to the log_corrcoefs function, leave the print_corrcoefs for backward compatibility. We can inform to the users with a Deprecation Warning that this function will be removed in a future release.

examples/plot_steps_nowcast.py
```diff
@@ -8,187 +8,215 @@
...
-    geodata=metadata,
-    drawlonlatlines=False,
-    type="prob",
-    units="mm/h",
-    probthr=0.5,
-    title="Exceedence probability (+ %i min)" % (n_leadtimes * timestep),
-)
 
-# sphinx_gallery_thumbnail_number = 5
+logging.getLogger('matplotlib').setLevel(logging.WARNING)
```
aperezhortal, line 36: Just out of curiosity, do you know what happens if you remove this line? Does matplotlib clutters the output with debug messages?

aperezhortal (changes requested):
Hi @fox91, thanks for the contribution! I really like the addition of the logging module. 

I made a few comments about the changes implemented.  


About the logging levels, DEBUG can be used to give information about details that are of interest for diagnosing purposes. While INFO to display information about the operation of a program (only high-level things that shows that the program operates normally). 

Finally, what do you think if we can start with minimal formatting (especially for INFO levels): 

**[LEVEL] message**

or just

**message**

pysteps/__init__.py
dnerini: I agree with @aperezhortal on this, we should try to mimic the current behaviour. 

As a rule of thumb, we could also say that we use the INFO level for all print statements that are now unconditionally executed, while DEBUG can be used for those print statements that are currently conditional to a verbose argument or similar (e.g. the measure_time argument in the nowcast.steps() method).

pysteps/nowcasts/extrapolation.py
dnerini: Or we could remove the measure_time argument and log the computation time at DEBUG level, what do you think?

aperezhortal: That's a good point. But we should leave the keyword argument and add a deprecation warning to maintain backward compatibility (at least for a few releases).

dnerini (changes requested):
I am also really happy about the implementation of the logging module in pysteps, thank you @fox91 for this really nice contribution! 

I don't have much to add on top of what @aperezhortal already mentioned. I would say that 
now we mainly need to find out the right setup that we can use for all modules.

examples/plot_steps_nowcast.py
dnerini: I also support the idea of a dedicated example script showing how to use the logging module.

pysteps/nowcasts/extrapolation.py
dnerini: Agreed ;-)

fox91: So a `warnings.warn` with `PendingDeprecationWarning` for one versione, `DeprecationWarning` for the next versione and total remove with the third version?
Or is best to use a `DeprecationWarning` for the next two version and then remove?

pysteps/__init__.py
fox91: I'm following `pylint` suggestions but I'm probably wrong.
I'm changing to lowercase with the next commit.

pysteps/__init__.py
fox91: I'm changing `DEBUG` to `INFO` in the next commit, thanks for the clarifications.

examples/plot_steps_nowcast.py
fox91: Using the logging module instead of print you don't see any output in the console if you don't add the proper config inside the example.
I can revert the improvements (ex: `main` function) but at minimum I need to add the `logging.basicConfig` at the top of the examples.

examples/plot_steps_nowcast.py
fox91: In this specific example about a ten debug lines only related to internal matplotlib. Not too many but useless, so I preferred to hide them.

pysteps/__init__.py
aperezhortal: Pylint expects that global variables should be uppercase.  However, we try to follow [a different naming convention](https://pysteps.readthedocs.io/en/latest/developer_guide/contributors_guidelines.html#coding-style-summary).

fox91: Maybe this can be another issue but: why not implement a `pylintrc` file in the root of the project, add to travis and enforce all code to use it?
You mention _Google’s python style guide_: it give you a good start point with [pylintrc](https://github.com/google/styleguide/blob/gh-pages/pylintrc).

pysteps/nowcasts/extrapolation.py
aperezhortal: Good point. According to [PEP-565](https://www.python.org/dev/peps/pep-0565/):
- `PendingDeprecationWarning` is hidden by default for all code.
- `DeprecationWarning` is reported by default for code that runs directly in the `__main__` module (as such code is considered relatively unlikely to have a dedicated test suite), but hidden by default for code in other modules.
- `FutureWarning` is reported by default for all code. The intended audience is users of applications written in Python, rather than other Python developers 

So, since we want to target the users, it seems that [FutureWarning](https://www.python.org/dev/peps/pep-0565/#additional-use-case-for-futurewarning) is the way to go.

examples/plot_steps_nowcast.py
aperezhortal: We can define a default configuration in the `pysteps.__init__.py` that prints the INFO messages by default and add a set_loglevel function to pysteps [as in matplotlib](https://matplotlib.org/3.1.1/faq/troubleshooting_faq.html#getting-help). 

This is an example using [matplotlib's logging implementation](https://github.com/aperezhortal/logging_test/tree/64837349e89e822af78c6f17d6d3f04aa098ee1a) that logs the info messages by default. I was thinking of something along those lines.

fox91: About that there is an open issue: https://github.com/matplotlib/matplotlib/issues/13264 and a PR: https://github.com/matplotlib/matplotlib/pull/13275
Personally I dislike and I'm not sure it's a good way to do the work.

fox91: I have not found any other library that uses `logging` like matplotlib.
It's also true that I don't find any library that uses so many output messages like pySTEPS to be able to compare the code and evaluate how they implement the output during a normal run.

aperezhortal: Although it is not the same as matplotlib, TensorFlow seems to implement a default handler and an interface to control the logging level.

https://github.com/tensorflow/tensorflow/blob/e5bf8de410005de06a7ff5393fafdf832ef1d4ad/tensorflow/python/platform/tf_logging.py#L93-L143

```
import logging
logger = tf.get_logger()
logger.setLevel(logging.ERROR)
```

dnerini:
Hi @fox91, do you plan to work on this PR again ? I think it would make an excellent contribution to pysteps!

fox91:
Hi @dnerini,
in the last period I have moved on to other projects.
In the meantime I came up with ideas on how to solve the problems highlighted by @aperezhortal.
I should be able to get back on this project in the coming weeks.

dnerini:
Very glad to hear that you'll be able to continue on this, @fox91, thanks a lot!

dnerini:
Im eventually closing this PR as stale, but the linked issue (#140) has been included to the pystepsV2 project.

PR #143, opened 2020-04-01
Changes to KNMI importer, test and visualization tools

RubenImhoff:
In contrast to the other PR, this is a version with the KNMI data in kilometers (hopefully in time for the next release). As far as I could see, it does not effect the pysteps runs, but it is more elegant than the application in the other PR. Hence, it depends on if we can live with this, or if we require the grid to be in meters. In addition (and also present in the other PR), the y1- and y2-values are reversed. This does not effect the nowcast, but it helps a lot when plotting the data. ;)


aperezhortal (approved):
Thanks @RubenImhoff! The PR looks good to go. Should we close #139  in favor of this one?

RubenImhoff:
Thanks @aperezhortal! If @dnerini agrees too, we can merge this one and close the other PR. :)

dnerini (approved):
Thanks for the PR and the nice contribution, @RubenImhoff !
I only pushed some changes to remove all references to 'meters' in the docstrings, now that we know that these were not necessary. Hopefully it'll be self explanatory that all geographical units have to be in the same units of the data projection. 
I found only one routine where we assume that the pixel size is in meter (nowcasts.sseps), but that code is anyway a bit messy. I'll need to refactor it at some point. For the time being, I just specified in the docstrings that the routine makes that assumption.

The PR looks good to go for me, too! Well done!

PR #145, opened 2020-04-09
Add importer for NSSL's MRMS grib files

aperezhortal:
This PR adds an importer to read the Importer for NSSL's Multi-Radar/Multi-Sensor System  ([MRMS](https://www.nssl.noaa.gov/projects/mrms/)) rain rate product (grib format).

[pygrib](https://github.com/jswhit/pygrib) was added as a new optional dependency. There are two more alternative grib readers for python: [NCAR's pynio](https://www.pyngl.ucar.edu/Nio.shtml) and [ECMWF's cfgrib](https://github.com/ecmwf/cfgrib). However, `pygrib` allowed more control over the data to be retrieved, which reduces the reading time considerably.

Since this product has a fine grid (1km x 1km) and a large domain (7000 x 3500, [lon, lat]), optional preprocessing options were added to make the data easier to handle during prototyping. 
By default, the data is load using a  float32 data type and its original resolution is downscaled by 4     (giving a 4 km grid spacing). Also, since the domain is quite large, an option to extract the precipitation data for a sub-region was added.

Another point that worth discussing is the new function `pysteps.utils.block_reduce` which merges the  [skimage.measure.block_reduce](https://scikit-image.org/docs/dev/api/skimage.measure.html#skimage.measure.block_reduce) and [skimage.util.view_as_blocks](https://scikit-image.org/docs/dev/api/skimage.util.html#view-as-blocks) functions to avoid an extra dependency in pysteps. However, if we add Scikit-image as a dependency, our implementation of `block_reduce` can be reduced to a few lines of codes or even be removed.


pysteps/datasets.py
```diff
@@ -149,6 +158,71 @@ def end(message="Download complete"):
         sys.stdout.write("\n" + message + "\n")
 
 
+def download_mrms_data(dir_path=None, frames=35):
```
dnerini: The example data for MRMS are downloaded from an external source instead of having them stored in pysteps-data like the other datasets. This is perfectly fine, we just need to be sure that this source is active. It seems to me that you're already testing it in `test_datasets.py`, is that right?

pysteps/io/importers.py
```diff
@@ -131,6 +133,230 @@
...
+    """
+    Importer for NSSL's Multi-Radar/Multi-Sensor System
+    ([MRMS](https://www.nssl.noaa.gov/projects/mrms/)) rainrate product
+    (grib format).
+
+    The rainrate values are expressed in mm/h, and the dimensions of the data
+    array are [latitude, longitude]. The first grid point (0,0) corresponds to
+    the upper left corner of the domain, while (last i, last j) denote the
+    lower right corner.
+
```
dnerini, line 177: Well done! So far in pysteps we have somewhat overlooked the dtype when importing data (but this applies to the other methods, too). Maybe we should apply the same approach you're using here to all importers to finally introduce an explicit control of it.

pysteps/io/importers.py
```diff
@@ -131,6 +133,230 @@
...
+        Name of the file to import.
+
+    fillna : float or np.nan
+        Value used to represent the missing data ("No Coverage").
+        By default, np.nan is used.
+
+    dtype : str
+        Data-type to which the array is cast.
+        Valid values:  "float32", "float64", "single", and "double".
+
```
dnerini, line 207: The `lat_range` and `lon_range` arguments have somewhat the same functionality as the `extent` argument in the `import_saf_crri` method. If possible, I suggest using the same approach for both methods to provide some consistency among the importers.

pysteps/utils/arrays.py
```diff
@@ -11,6 +11,81 @@
 """
 
 import numpy as np
+from numpy.lib.stride_tricks import as_strided
+
+
```
dnerini, line 16: This new method seems very similar to the already existing `utils.dimensions.aggregate_fields_space` method, although with improved functionalities (e.g., data are trimmed if not perfectly divisible).

Do we need to keep both? Can we have only one method to upscale/reduce/aggregate fields in space?

dnerini (commented):
Hi Andres, this is a really exciting PR, I look forward to include it in the next release! 
I included some first rather high level comments in your code that I'd like to discuss with you before merging into master.
Since I suspect this new feature will interest many people, it may be interesting to also include a short example to show it (we could also just change the dataset in one of the already existing examples). What do you think?

pysteps/io/importers.py
```diff
@@ -131,6 +133,230 @@
...
+    lat_range : tuple, list, or None
+        Latitude range (min_lat, max_lat) of the data to be retrieved.
+        By default, the entire latitude range is retrieved.
+
+    lon_range : tuple, list, or None
+        Longitude range (min_lon, max_lon) of the data to be retrieved.
+        By default, the entire longitude range is retrieved.
```
dnerini: The `lat_range` and `lon_range` arguments have somewhat the same functionality as the `extent` argument in the `import_saf_crri` method. If possible, I suggest using the same approach for both methods to provide some consistency among the importers.

pysteps/io/importers.py
```diff
@@ -131,6 +133,230 @@
...
+    The rainrate values are expressed in mm/h, and the dimensions of the data
+    array are [latitude, longitude]. The first grid point (0,0) corresponds to
+    the upper left corner of the domain, while (last i, last j) denote the
+    lower right corner.
+
+    Due to the large size of the dataset (3500 x 7000), a float32 type is used
+    by default to reduce the memory footprint. However, be aware that when this
+    array is passed to a pystep function, it may be converted to double
+    precision, doubling the memory footprint.
+    To change the precision of the data, use the *dtype* keyword.
```
dnerini, line 218: Well done! So far in pysteps we have overlooked a bit the dtype when importing data (but this applies to the other methods, too). Maybe we should apply the same approach you're using here to all importers to finally introduce an explicit control of it.

pysteps/io/importers.py
```diff
@@ -131,6 +133,230 @@
...
+    lats = np.linspace(ul_lat, lr_lat, grib_msg["Nj"])
+    lons = np.linspace(ul_lon, lr_lon, grib_msg["Ni"])
+
+    # NOTE:
+    # Values equal to -3 represents "No Coverage" and they are considered
+    # missing values. This reader replace those values by np.nan.
+    # Note that Missing values are not the same as zero precipitation values.
+    # Missing values indicates regions with no valid measures.
+    # While zero precipitation indicates regions with valid measurements,
+    # but with no precipitation detected.
```
dnerini: This is an important comment, perhaps worth moving it in the docstrings as a note?

pysteps/utils/arrays.py
```diff
@@ -11,6 +11,81 @@
...
+def block_reduce(image, block_size, func=np.mean, **kwargs):
+    """Downsample image by applying a function to local blocks
+    (average by default).
```
dnerini: This new method seems very similar to the already existing `utils.dimensions.aggregate_fields_space` method, although with improved functionalities (e.g., data are trimmed if not perfectly divisible).

Do we need to keep both? Can we have only one method to upscale/reduce/aggregate fields in space?

RubenImhoff:
Great contribution, @aperezhortal! I looked through the files and other than @dnerini's comments, I have nothing to add.

pysteps/datasets.py
aperezhortal: Yes. However, the archive where the data is stored warns that:

>The IEM is a volunteer effort and receives no funds for facilities or staff from Iowa State University or the State of Iowa. Users of the IEM must therefore recognize that the IEM may be discontinued at any time with little or no notice.
https://mesonet.agron.iastate.edu/disclaimer.php

pysteps/io/importers.py
aperezhortal: That is a good idea. Probably, an elegant solution is to use a decorator that changes the output datatype. To avoid modifying other importers in this PR, after this PR is merged, we can prepare a new PR implementing that change.

pysteps/io/importers.py
aperezhortal: Good point. I made the change but the way the `extent` is used in the MRMS importer is different from the `import_saf_crri` function. `import_mrms` uses the extent in degrees while `import_saf_crri` uses the data coordinates units (km). 

The `extent` keyword can be also applied to the other importers, allowing the user to select the units of the extent (km or degrees). This is a good candidate for a separate PR (including the data type casting).

pysteps/io/importers.py
aperezhortal: Change made in commit c07d0cc

pysteps/utils/arrays.py
aperezhortal: Oops, my bad. I totally forgot about the functions in the `utils.dimensions` module. Commit 7e493f6 refactors the 'aggregate_fields' function to include the functionality of the  `block_reduce` function (now removed).  Also, I added a few more tests to check the `aggregate_fields` function under different scenarios.

pysteps/datasets.py
dnerini: Alright, thanks for the info. We just know that when that happens, the test_datasets will fail and we'll probably need to deprecate that method. 
But if it's only about having a small example set, then perhaps having it in our pysteps-data repo is not a bad idea. What do you think?

pysteps/io/importers.py
dnerini: Excellent, thanks!

I don't see it as a different use of `extent` actually: both use the (left, right, bottom, top) in *data coordinates* to specify bounding box for a regular grid. Whether this is in km or degrees doesn't really matter.

Yes, good idea, extent could be implemented for all importers.

pysteps/utils/arrays.py
dnerini: Nice job!

pysteps/utils/dimension.py
```diff
@@ -206,71 +213,125 @@ def aggregate_fields_space(R, metadata, space_window, ignore_nan=False):
...
 
-def aggregate_fields(R, window_size, axis=0, method="mean"):
-    """Aggregate fields.
-    It attemps to aggregate the given R axis in an integer number of sections
+def aggregate_fields(data, window_size, axis=0, method="mean", trim=False):
+    """Aggregate fields along a given direction.
+
+    It attempts to aggregate the given R axis in an integer number of sections
     of length = window_size.
     If such a aggregation is not possible, an error is raised.
```
dnerini: "unless `trim=True`, in which case the axis is trimmed to make it perfectly divisible".

pysteps/io/importers.py
aperezhortal: You are right. The actual difference is in the units. We can add another keyword to specify this, like extend_units='latlon' or 'km'. 

I will add it to this PR since it will become handy if we want this functionality in all the importers.

pysteps/datasets.py
aperezhortal: Good point, it is only 30Mb. I will move them to the pysteps-data and remove the download_mrms_data function. It will be more stable in the long term.

dnerini: Ok, but I wouldn't just remove your original function! Can we leave it as additional method in the dataset module? Maybe you could just extend it to accept an arbitrary timestamp and number of frames. Although maybe this would go a bit behind the scope of module...

pysteps/io/importers.py
dnerini: I wonder whether this is actually needed, since extent is to be specified in the same units of the dataset. In most use cases, one would want to subset the field using the same coordinates. Or am I missing something here?

pysteps/datasets.py
aperezhortal: I was too drastic :laughing:. With minor refactoring, `download_mrms_data` could be used to retrieve data from the server between two dates. I already have the code snippet needed for that. I would only add a warning in case they try to retrieve a large period, let's say, one year (to be gentle on the server).

pysteps/io/importers.py
aperezhortal: So far, the importers retrieve the data coordinates in "km" or "m" with respect to the origin. However, we can also use the lat/lon associated with the grid points to define the extent of the data that we want to retrieve.  The domain of the MRMS quite large (the contiguous US), if you want to focus on a single state, you simply specify the lat-lon bounding box. On the other hand, to use the data coordinates, you first need to find the bounding box coordinates measured from the origin, which will require an additional step. 

We can have as default behavior the coordinates specified in the same units of the dataset, and leave the lan/lon units as a particular use-case. What do you think?

aperezhortal: It worth mentioning that the `extend units` are unnecessary for the MRMS importer, since the data coordinates [x, y] are the same as the [lon, lat] (a regular lat/lon projection is used).

However, it may be useful for the other importers where a projection that transforms the lat/lon coordinates to x/y native map projection coordinates (in meters).

dnerini: I see your point, but the `extent` argument as it is defined now (i.e. [left, right, bottom, top] coordinates) can only work with a regular grid. To be able to accept bounding boxes in lat lon for any projection, we would need 4 pairs of coordinates (the 4 corners of the bounding box). For the MRMS data, the regular grid is defined using a geographical projection in lat lon, so the extent can be specified in degrees.

I suggest keeping the same approach as it is now, which closely mimic the behavior of the `extent` argument in the pyplot.imshow() method. In brief, the extent is specified in whatever coordinates a given dataset is based on. This will ensure that the extent is always with respect to a regular grid.

aperezhortal: > To be able to accept bounding boxes in lat lon for any projection, we would need 4 pairs of coordinates (the 4 corners of the bounding box). 

I see the possible issue.

> I suggest keeping the same approach as it is now

Agree, let's keep the actual approach.

aperezhortal:
Another issue worth discussing before merging the PR. 

This importer uses a regular lat-lon projection and the native coordinates (x,y) of the projection are expressed in degrees. This projection is seamlessly handled by pysteps' functions as the other projections since we may have an ambiguous interpretation of the `xpixelsize` and `ypixelsize` in different routines. 

#### xpixelsize and ypixelsize assumed in meters:

- `scripts/run_vel_pert_analysis.py` 
- `pysteps.nowcasts.sseps.forecast(*args,**kwargs)`
This may cause issues with the KMNI importer that uses km.  

#### xpixelsize and ypixelsize in the same units as the native projection (m, km, or degrees)

- aggregate_fields_space(R, metadata, space_window, ignore_nan=False):
- clip_domain(*args,**kwargs)  
- pysteps.visualization.utils.reproject_geodata(geodata, t_proj4str, return_grid=None)
- pysteps.visualization.motionfields

The MRMS grid has a 0.01°x0.01° constant grid spacing.  However, the distances in km (arc along the earth surface) are not constant. They range from 640 to 1046 meters in the west-east direction, and from 1046 to 1113 meters in the north–south direction.

How we should tackle this issue? Perhaps adding a `pixelsize_units` keyword? 
Any thoughts are welcome.

dnerini:
We briefly discussed this issue in #143. As you point out, almost all methods and routines do not assume any particular spatial unit nor projection, just a regular grid in whatever units.
 
I reckon that the simplest solution is to assume that `xpixelsize` and `ypixelsize` (and any other variable related to space, e.g. `x1`, `x2`, `y1`, `y2`, or `extent` where needed) use the same units as in the native projection specified in the `projection` metadata. This would not require any short-term change in our codes.

For the MRMS dataset, this means using degrees latitude and longitude, since these define its regular grid. As you mentioned, reprojecting the lat lon coordinates will result in irregular `xpixelsize` and `ypixelsize`, as well as in a irregular extent, i.e., not in a rectangle that can be expressed by  `x1`, `x2`, `y1`, `y2`. I suggest that in your MRMS importer you always refer to units of degrees. You can then specify the equivalent spacing in km with formulas like "which correspond to roughly 1 km".

The `nowcast.sseps` method is a bit of an unfortunate exception at the moment, but I would not change everything else just because of this small inconsistency. As already said, I plan anyway to refactor it, at that point I will be able to fix this problem. For the time being, I specified in the docstrings that the method assumes that `xpixelsize` and `ypixelsize` to be in meters. 

Similar discussion for `scripts/run_vel_pert_analysis.py`. I think we can live with another small exception to the general rule. Perhaps @spulkkin will be able at some point to change his script to be able to accept any unit for `xpixelsize` and `ypixelsize`.

pysteps/datasets.py
aperezhortal: ac37cc2 refactored  `download_mrms_data()` to allow downloading the data over a time range. Tretrieve the data from the server. Since this relies on an external server, tests of this function were removed.

aperezhortal:
>I think we can live with another small exception to the general rule.

:thumbsup: The pixelsizes in the metadata are now expressed in degrees, the same units as the data coordinates. 

Also, I tried to add an example of a nowcast using the MRMS data (commit 9b899f9), but I had to revert the changes because it was causing issues in ReadTheDocs that were not trivial to solve (4315d27). Maybe is a good idea to leave the example for a separate PR.

dnerini (approved):

dnerini:
@aperezhortal with the last commits, this PR looks good to go! 
Maybe a last detail, you could add MRMS to the pystepsrc file, along with the other available data sources?

I'm sorry for the MRMS example. We are a bit on the limit with resources on RDT, but it shouldn't be too hard to obtain more memory if needed. Also, did you try installing miniconda instead of conda? This might considerably reduce the memory usage. But I agree with you: let's take this to a  new PR.

aperezhortal:
Read the docs has conda available in its docker image. The conda support is activated using the yaml configuration file. The memory issue is caused during the dependencies solving in conda. In the RTD github page, there are many issues about this. Most of them were solved by the RTD people,  increasing the maximum memory for the projects to 2GB.
https://docs.readthedocs.io/en/stable/config-file/v2.html#conda

> Maybe a last detail, you could add MRMS to the pystepsrc file

Added in commit 5ea61d376610a233df7294960ecaae41a222a0a4

dnerini:
OK, thanks! I propose to make a request to RDT for more memory together with the PR implementing the new example with MRMS data.

aperezhortal:
Merged. Thanks @dnerini for reviewing this PR!

PR #146, opened 2020-04-09
Remove AttrDict dependency

### LINKED ISSUE #144: AttrDict package is inactive
Running the pysteps tests I found a few warnings related to the AttrDict package:

```
python3.8/site-packages/attrdict/mapping.py:4: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working

    from collections import Mapping

python3.8/site-packages/attrdict/mixins.py:5: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working

    from collections import Mapping, MutableMapping, Sequence
```

As the warning says, this will cause an error in python 3.9 versions. Since the package has been tagged as inactive and the GitHub repository has been archived, there is a minor chance that a new version of the package that fixes this issue will be released.

The AttrDict is used in pysteps to access the default parameters using the dictionary key or as attributes. This was introduced to maintain backward compatibility with the first versions of pysteps.

A solution that will be helpful in the long-term is removing the ``DocDictify`` class from the ``__init__.py`` module, and the AttrDict dependency with it. However, This dual type of access to the parameters is used in the examples and it may cause problems for the users that take advantage of the duality dictionary-attributes of the current interface.

aperezhortal:
This PR removes the AttrDict package as a dependency and closes #144. 

It turns out that inheriting AttrDict in the `DotDictify` class was not needed for the common *set* and _get_ operations used in pysteps.

With this PR, `DotDictify` inherits a `dict` class.  To make sure that the rcparams can be accessed as items or as attributes, a series of tests were added.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: python
 
 env:
   global:
-    - MINIMAL_DEPENDENCIES="cython numpy attrdict jsmin jsonschema matplotlib netCDF4 opencv pillow pyproj scipy dask"
+    - MINIMAL_DEPENDENCIES="cython numpy jsmin jsonschema matplotlib netCDF4 opencv pillow pyproj scipy dask"
     - OPTIONAL_DEPENDENCIES="dask pyfftw cartopy h5py PyWavelets"
     - TEST_DEPENDENCIES="pytest pytest-cov codecov"
 
--- a/doc/source/user_guide/install_pysteps.rst
+++ b/doc/source/user_guide/install_pysteps.rst
@@ -9,8 +9,6 @@ Dependencies
 The pysteps package needs the following dependencies
 
 * `python >=3.6 <http://www.python.org/>`_
-* `attrdict <https://pypi.org/project/attrdict/>`_
-* `jsmin <https://pypi.org/project/jsmin/>`_
 * `jsonschema <https://pypi.org/project/jsonschema/>`_
 * `matplotlib <http://matplotlib.org/>`_
 * `netCDF4 <https://pypi.org/project/netCDF4/>`_
@@ -163,12 +161,12 @@ Setting up the user-defined configuration file
 ----------------------------------------------
 
 .. _JSON: https://en.wikipedia.org/wiki/JSON
-.. _AttrDict: https://pypi.org/project/attrdict/
 
 The pysteps package allows the users to customize the default settings
 and configuration.
 The configuration parameters used by default are loaded from a user-defined
-JSON_ file and then stored in the **pysteps.rcparams** AttrDict_.
+JSON_ file and then stored in the **pysteps.rcparams**, a dictionary-like object
+that can be accessed as attributes or as items.
 
 .. toctree::
     :maxdepth: 1
--- a/doc/source/user_guide/set_pystepsrc.rst
+++ b/doc/source/user_guide/set_pystepsrc.rst
@@ -4,15 +4,12 @@ The pySTEPS configuration file (pystepsrc)
 ==========================================
 
 .. _JSON: https://en.wikipedia.org/wiki/JSON
-.. _AttrDict: https://pypi.org/project/attrdict/
 
 The pysteps package allows the users to customize the default settings
 and configuration.
 The configuration parameters used by default are loaded from a user-defined
-JSON_ file and then stored in the **pysteps.rcparams** AttrDict_.
-
-The configuration parameters can be accessed as attributes or as items
-in a dictionary.
+JSON_ file and then stored in `pysteps.rcparams`, a dictionary-like object
+that can be accessed as attributes or as items.
 For example, the default parameters can be obtained using any of the following ways::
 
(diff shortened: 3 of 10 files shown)


dnerini (approved):
Really nice to see that you found a very elegant solution to the issue with the attrdict dependency!

I have only one small comment.

Currently, `__init__.py` imports a bunch of external modules that end up together with the actual pysteps methods:
```
>>> import pysteps
>>> pysteps.
pysteps.DotDictify(        pysteps.extrapolation      pysteps.nowcasts           pysteps.timeseries
pysteps.Draft4Validator(   pysteps.io                 pysteps.os                 pysteps.utils
pysteps.cascade            pysteps.jsmin(             pysteps.plt                pysteps.verification
pysteps.config_fname(      pysteps.json               pysteps.postprocessing     pysteps.vf
pysteps.datasets           pysteps.load_config_file(  pysteps.rcparams           pysteps.visualization
pysteps.decorators         pysteps.motion             pysteps.stat               pysteps.warnings
pysteps.exceptions         pysteps.noise              pysteps.sys
```
I suggest renaming the external methods with a leading underscore to hide them, for example
```
import json as _json
import os as _os
import stat as _stat
import sys as _sys
import warnings as _warn

from jsmin import jsmin as _jsmin
from jsonschema import Draft4Validator as _Draft4Val
```
which produces the following (note that I also declared the `DocDotify` class as private):
```
>>> import pysteps
>>> pysteps.
pysteps.cascade            pysteps.extrapolation      pysteps.nowcasts           pysteps.utils
pysteps.config_fname(      pysteps.io                 pysteps.plt                pysteps.verification
pysteps.datasets           pysteps.load_config_file(  pysteps.postprocessing     pysteps.vf
pysteps.decorators         pysteps.motion             pysteps.rcparams           pysteps.visualization
pysteps.exceptions         pysteps.noise              pysteps.timeseries
```

What do you think?

I was also thinking that at some point we might want to avoid importing all pysteps submodules at once, so to allow a user to quickly load only the needed modules. But this is another story.

In case case, well done with the PR and thanks for the nice contribution!

aperezhortal:
> Currently, __init__.py imports a bunch of external modules that end up together with the actual pysteps methods:

> I suggest renaming the external methods with a leading underscore to hide them, for example

This also happens in every sub-module in pysteps. For example `print(dir(pysteps.io.importers))` will show that **['imread', 'netCDF4', 'np', 'os', 'pyproj']** are also attributes of the `importers` module. 

I'm in favor of leaving the imports as they are to maintain style consistency across all the modules and avoid refactoring them (add the underscores to hide imports).
Also adding an underscore to hide the imports will make the code looks less "standard". E.g.,  `np` is commonly used for numpy, instead of `_np`. 
I found an interesting point of view about this issue on StackOverflow (most voted answer): 
https://stackoverflow.com/questions/26516405/hiding-module-imports-in-package


-------------------------------------------------------


>  (note that I also declared the DocDotify class as private)

That is a good point. DocDotify should be a private class. Changed in commit 1947906.

-------------------------------------------------------

>I was also thinking that at some point we might want to avoid importing all pysteps submodules at once, so to allow a user to quickly load only the needed modules. But this is another story.

There is no difference in terms of importing time. The full pysteps module is always loaded, even if a single package is imported (https://softwareengineering.stackexchange.com/questions/187403/import-module-vs-from-module-import-function). 
Importing the submodules allows them to be accessed as attributes of pysteps. But I think that we are not handling this in the same way for all the submodules, with some of them import the downstream submodules and others that don't. I will check this.

dnerini:
> 
> 
> > Currently, **init**.py imports a bunch of external modules that end up together with the actual pysteps methods:
> 
> > I suggest renaming the external methods with a leading underscore to hide them, for example
> 
> This also happens in every sub-module in pysteps. For example `print(dir(pysteps.io.importers))` will show that **['imread', 'netCDF4', 'np', 'os', 'pyproj']** are also attributes of the `importers` module.

Right, good point!

> I'm in favor of leaving the imports as they are to maintain style consistency across all the modules and avoid refactoring them (add the underscores to hide imports).
> Also adding an underscore to hide the imports will make the code looks less "standard". E.g., `np` is commonly used for numpy, instead of `_np`.

Agreed!

> I found an interesting point of view about this issue on StackOverflow (most voted answer):
> https://stackoverflow.com/questions/26516405/hiding-module-imports-in-package

Thanks for the reference, very interesting indeed! I really liked the quote at the end 
"Remember, in Python, `We're all consenting adults here.`": it kind of summarizes well the whole spirit :-) 

> > (note that I also declared the DocDotify class as private)
> 
> That is a good point. DocDotify should be a private class. Changed in commit [1947906](https://github.com/pySTEPS/pysteps/commit/19479064b57b97f7d0336dfe7149ff06af9062e5).

👍

PR #147, opened 2020-04-14
Fix mask arrays in LK

dnerini:
This PR fixes few issues in the LK motion method concerning the handling of MaskedArrays, namely
- wrong if statements
- inaccurate documentation 
- copy of MaskedArray returned ndarray instead
- unsupported MaskedArrays in `images.morph_opening`

These caused LK motion  routine to recompute a new masked array on the input data, thus ignoring the mask that was passed, unless the masked values where nans. 

A simple test to check whether using MaskedArrays or ndarrays produces the same results was added.

### DIFF
--- a/pysteps/utils/images.py
+++ b/pysteps/utils/images.py
@@ -187,10 +187,16 @@ def morph_opening(input_image, thr, n):
     """Filter out small scale noise on the image by applying a binary
     morphological opening, that is, erosion followed by dilation.
 
+    .. _MaskedArray:\
+        https://docs.scipy.org/doc/numpy/reference/maskedarray.baseclass.html#numpy.ma.MaskedArray
+
+    .. _ndarray:\
+    https://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.html
+
     Parameters
     ----------
 
-    input_image : array_like
+    input_image : ndarray_ or MaskedArray_
         Array of shape (m, n) containing the input image.
 
     thr : float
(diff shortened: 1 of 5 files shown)


dnerini:
I change this PR to draft, as I was thinking to also include few tests for MaskedArrays.

pysteps/utils/images.py
```diff
@@ -208,8 +217,18 @@ def morph_opening(input_image, thr, n):
             "routine but it is not installed"
         )
 
+    input_image = input_image.copy()
+
+    # masked array
```
RubenImhoff: Perhaps to make the comment a little clearer:
`#  Check if a MaskedArray is used. If not, mask the ndarray`

pysteps/utils/images.py
```diff
@@ -223,4 +242,7 @@ def morph_opening(input_image, thr, n):
     # Filter out small isolated pixels based on mask
     input_image[mask] = np.nanmin(input_image)
 
+    if to_ndarray:
```
RubenImhoff, line 245: Again, just for the details ;) 
`# If an ndarray is used as input, return an ndarray`

RubenImhoff (commented):
Nice work, @dnerini! I have nothing important to add so far, just two proposed line comments to 'document' it a little better. 
Do you know if we can encounter the same problems with the other optical flow algorithms?

RubenImhoff:
Seems so far like the other scripts in pysteps.motion don't or cannot have this issue. :)

dnerini:
Thanks @RubenImhoff for the review!
I made the changes you suggested to the inline comments in a8e5c5d.

RubenImhoff:
Looks good, @dnerini!

PR #148, opened 2020-04-15
Add decorator to importers for postprocessing the precipitation data.

aperezhortal:
This PR adds a decorator to handle simple postprocessing operations for the importers:
* Allow explicitly define the precipitation data type (dtype keyword)
* Set invalid or missing data to a predefined value (fillna keyword)
* Add "Other parameters" docstrings with the documentation of the keywords used by the decorator.

It also makes the keywords arguments used in each importer explicit (with the exception of the keywords handled by the decorator). Hence, in the API documentation, the default values of those keywords are shown. 

- [x] Add `postprocess_import` decorator to the all the data and nowcast importers
- [x] Add more tests for the `postprocess_import` decorator
- [x] Add support for MaskedArrays in `postprocess_import`


RubenImhoff:
Nice work, @aperezhortal! Looks good so far to me, I made some small textual changes, but that's all for now. :)

pysteps/tests/helpers.py
```diff
@@ -193,3 +204,22 @@ def smart_assert(actual_value, expected, tolerance=None):
                                              abs=tolerance,
                                              nan_ok=True,
                                              )
+
+
+def get_invalid_mask(input_array, fillna=np.nan):
+    """
+    The the mask indicating the invalid values.
```
RubenImhoff: Maybe good to elaborate a bit on this explanation, so that it is directly clear what this function does.

RubenImhoff (approved):

PR #150, opened 2020-04-18
Anvil

pulkkins:
Implementation of the ANVIL method developed in Pulkkinen et al.: Nowcasting of Convective Rainfall Using Volumetric Radar Observations, IEEE Transactions on Geoscience and Remote Sensing, doi: 10.1109/TGRS.2020.2984594, 2020.


aperezhortal:
Nice job @pulkkins! 

It seems that the branch's commit history is out of sync with the latest commit in master and the **files changes** tab shows modifications from other PRs. 
Can you merge the latest master into anvil to see the only changes made by this PR?

pulkkins:
> Nice job @pulkkins!
> 
> It seems that the branch's commit history is out of sync with the latest commit in master and the **files changes** tab shows modifications from other PRs.
> Can you merge the latest master into anvil to see the only changes made by this PR?

Done. The "files changed" tab should make more sense now.

pysteps/nowcasts/anvil.py
```diff
@@ -0,0 +1,370 @@
...
+    mask = np.logical_and(mask, stdx * stdy > 1e-8)
+    mask = np.logical_and(mask, n > 1e-3)
+    corr = np.empty(x.shape)
+    corr[mask] = cov[mask] / (stdx[mask] * stdy[mask])
+    corr[~mask] = 0.0
+
+    return corr
+
+
+def _R_VIL_regression(vil, r, window_radius):
```
aperezhortal: This function name can be lowercase to comply with pep8.

pysteps/nowcasts/anvil.py
```diff
@@ -0,0 +1,370 @@
...
+
+try:
+    import dask
+
+    DASK_IMPORTED = True
+except ImportError:
+    DASK_IMPORTED = False
+
+
+def forecast(vil, rainrate, velocity, n_timesteps, n_cascade_levels=8,
```
aperezhortal, line 36: This function seems to be missing in `pysteps.nowcast.interface`.

pysteps/nowcasts/anvil.py
```diff
@@ -0,0 +1,370 @@
...
+             extrap_method="semilagrangian", ar_order=2, ar_window_radius=50,
+             r_vil_window_radius=5, fft_method="numpy", num_workers=1,
+             extrap_kwargs=None, filter_kwargs=None, measure_time=False):
+    """Generate a nowcast by using the ANVIL method.
```
aperezhortal: Perhaps a very short description of the method could be added to the docstrings.

pysteps/nowcasts/anvil.py
```diff
@@ -0,0 +1,370 @@
...
+    print("Starting nowcast computation.")
+
+    if measure_time:
+        starttime_mainloop = time.time()
+
+    r_f = []
+    dp = None
+    for t in range(n_timesteps):
+        print("Computing nowcast for time step %d... " % (t + 1), end="")
+        sys.stdout.flush()
```
aperezhortal: `sys.stdout.flush()` can be removed by using print(..., flush=True) in the previous line.

pysteps/nowcasts/anvil.py
RubenImhoff: I think it's good to put some comments in this function. Perhaps good to refer to parts of your paper here! :) Same for line 157 where this function is used.

dnerini (changes requested):
Very nice, @pulkkins, congratulations on this new method !

Here you can find the documentation built for this PR
https://pysteps.readthedocs.io/en/anvil/

~~You'll notice that `nowcasts.anvil` is missing not only in `nowcasts.inferface.get_method`, as already pointed out by @aperezhortal, but from the `nowcasts` module altogether. This is because it needs to be manually added in the relevant documentation page, here:
https://github.com/pySTEPS/pysteps/blob/master/doc/source/pysteps_reference/nowcasts.rst~~
edit: 26a3186 includes anvil into the documentation

I still think that it would be nice to valorize your work by adding an example script of anvil in the pysteps gallery. There you could briefly explain the novelties of your method and show how to use it. What do you think?  Shouldn't be much work, as you can easily modify one of the existing examples.

~~Finally, I think that we need at least a basic test for the anvil method. Again, you could simply follow one of the existing tests for the nowcasting routines, which consist in running an example forecast and making sure that the forecast skill/error is above/below a certain threshold (which could simply be the current result on that set of data, so that we make sure that future modifications won't degrade the quality of the forecast).~~
edit: 1b7daaa introduces a first very basic test for anvil, feel free to expand/change as you like.
The current test only uses rain rate fields as input (since we do not have any VIL example data) and at only test for `ar_order=2` (different orders caused errors, as described in my comments below).

pysteps/nowcasts/anvil.py
```diff
@@ -0,0 +1,370 @@
...
+
+    phi = []
+    for i in range(n_cascade_levels):
+        if ar_order > 2:
+            phi_ = autoregression.estimate_ar_params_yw_localized(gamma[i, :], d=1)
+        elif ar_order == 2:
+            phi_ = _estimate_ar2_params(gamma[i, :])
+        else:
+            pass
+        phi.append(phi_)
```
dnerini, line 243: When `ar_order=1`, `phi_` is not defined and an UnboundLocalError is thrown.

pysteps/nowcasts/anvil.py
```diff
@@ -0,0 +1,370 @@
...
+    """
+    if len(vil.shape) != 3:
+        raise ValueError("vil.shape = %s, but a three-dimensional array expected" % str(vil.shape))
+
+    if rainrate is not None:
+        if len(rainrate.shape) != 2:
+            raise ValueError("rainrate.shape = %s, but a two-dimensional array expected" % str(rainrate.shape))
+
+    if vil.shape[0] != ar_order + 2:
+        raise ValueError("vil.shape[0] = %d, but vil.shape[0] = ar_order = %d required" % (vil.shape[0], ar_order))
```
dnerini: Looks like something is wrong here: the if statement checks whether `vil.shape[0] != ar_order + 2`, but the error message says that `vil.shape[0]` must be equal to `ar_order`.

pysteps/nowcasts/anvil.py
```diff
@@ -0,0 +1,370 @@
...
+    dp = None
+    for t in range(n_timesteps):
+        print("Computing nowcast for time step %d... " % (t + 1), end="")
+        sys.stdout.flush()
+        
+        if measure_time:
+            starttime = time.time()
+
+        for i in range(n_cascade_levels):
+            vil_dec[i, :] = autoregression.iterate_ar_model(vil_dec[i, :], phi[i])
```
dnerini, line 266: calling `anvil.forecast` with `ar_order=3` fails here with the following error message:

```
Computing ANVIL nowcast:
------------------------

Inputs:
-------
input dimensions: 320x355

Methods:
--------
extrapolation:   semilagrangian
FFT:             numpy

Parameters:
-----------
number of time steps:        3
parallel threads:            1
number of cascade levels:    8
order of the ARI(p,1) model: 3
ARI(p,1) window radius:      50
R(VIL) window radius:        5
Starting nowcast computation.
Computing nowcast for time step 1... Traceback (most recent call last):
  File "pysteps/tests/test_nowcasts_anvil.py", line 73, in <module>
    test_anvil_rainrate(**dict((x, y) for x, y in test_args))
  File "pysteps/tests/test_nowcasts_anvil.py", line 58, in test_anvil_rainrate
    ar_window_radius=ar_window_radius,
  File "/users/ned/pySTEPS/pysteps/pysteps/nowcasts/anvil.py", line 243, in forecast
    vil_dec[i, :] = autoregression.iterate_ar_model(vil_dec[i, :], phi[i])
  File "/users/ned/pySTEPS/pysteps/pysteps/timeseries/autoregression.py", line 975, in iterate_ar_model
    raise ValueError("dimension mismatch between x and phi: x.shape[0]=%d, len(phi)=%d" % (x.shape[0], len(phi)))
ValueError: dimension mismatch between x and phi: x.shape[0]=3, len(phi)=5
```

RubenImhoff:
Nice work! :)

pysteps/nowcasts/anvil.py
pulkkins: It could be a good idea to disallow using a higher-order AR model (ar_order>2). Not sure if the nowcasts are reliable because the stationarity of the AR process can only be guaranteed when ar_order<=2.

dnerini: If stationarity is the problem, maybe it is enough to specify it somewhere in the docstrings.
But is this the cause for the error above?

aperezhortal: Seems that is caused by line 241: `vil_dec = vil_dec[:, -3:, :]`, that keeps only the last 3 frames of the 4 passed in the tests. Does ANVIL always use only 3 frames?

pulkkins: The error is caused by a different problem. I'll fix that.

pulkkins: Line 241 was indeed incorrect.

pulkkins:
Thanks for implementing the test module. I added a gallery example illustrating how ANVIL can predict growth and decay patterns.

dnerini (approved):
I really like the new example script, @pulkkins !
To me everything now seems ready for merging this very nice PR into master, well done!

RubenImhoff (approved):
I agree with @dnerini, super nice work!

aperezhortal (approved):

pulkkins:
Thanks for finding the bugs and suggesting improvements!

PR #152, opened 2020-04-19
Fix warnings shown during tests

aperezhortal:
This PR fixes some of the warnings shown when the test suite is run. 
- `pysteps.motion.vet`: Fix bug in `vet` keywords handling. Previously, the `method` argument passed to the minimizer was ignored.
- `pysteps.io.importers`: Add a function to safely get the rain/no rain threshold that handles NaN correctly.
- Fix collections imports. Remove DeprecationWarning for using the ABCs from 'collections' instead
    of from 'collections.abc'.
- `pysteps.timeseries.autoregression`: Make some docstrings raw literals, avoiding the interpretation of math equations as invalid escape sequences.

### DIFF
--- a/pysteps/io/importers.py
+++ b/pysteps/io/importers.py
@@ -215,7 +215,12 @@ def _get_threshold_value(precip):
     valid_mask = np.isfinite(precip)
     if valid_mask.any():
         _precip = precip[valid_mask]
-        return np.min(_precip[_precip > _precip.min()])
+        min_precip = _precip.min()
+        above_min_mask = _precip > min_precip
+        if above_min_mask.any():
+            return np.min(_precip[above_min_mask])
+        else:
+            return min_precip
     else:
         return np.nan
 
--- a/pysteps/io/importers.py
+++ b/pysteps/io/importers.py
@@ -200,6 +200,26 @@ def _get_grib_projection(grib_msg):
     return projparams
 
 
+def _get_threshold_value(precip):
+    """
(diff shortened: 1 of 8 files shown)


pysteps/io/importers.py
```diff
@@ -200,6 +200,26 @@ def _get_grib_projection(grib_msg):
...
+
+    Returns
+    -------
+
+    threshold : float
+    """
+    valid_mask = np.isfinite(precip)
+    if valid_mask.any():
+        _precip = precip[valid_mask]
+        return np.min(_precip[_precip > _precip.min()])
```
dnerini: Line 218 will fail if precip is a uniform field (e.g. all zeros).
I think we need to add an additional test to line 216, something like:
`if valid_mask.any() and (precip[valid_mask] > precip[valid_mask].min()).any():`
or
`if valid_mask.any() and not (precip[valid_mask] == precip[valid_mask][0]).all():`
or whatever is more elegant/efficient, you get the point :-)

dnerini (changes requested):
Turns out there was a bit of a mess on how we defined the rainfall threshold in the various importers, thanks for cleaning this up! 
As explained in my in-line comment, I think we just need to handle cases when the field has uniform value (e.g. all zeros), otherwise looks good to go!

pysteps/io/importers.py
aperezhortal: Commit tf4614c0 fixes that bug :bowtie:

dnerini (approved):
excellent, looks good to go!

Issue #27, opened 2018-12-13
Make code pep8 compilant


aperezhortal:
Many parts of the source code do not follow good coding practices recommended for python ([pep8 style guide](https://www.python.org/dev/peps/pep-0008/)).

Although this requires some effort, it will increase the readability of the package library.
A good summary of the standard can be found [here](https://development.robinwinslow.uk/2014/01/05/summary-of-python-code-style-conventions/).

Some good candidates to tackle first are:

**Whitespace and newlines, Indentation, line-length & code wrapping** 

* Always use 4 spaces for indentation (don’t use tabs)
* Write UTF-8
* Max line-length: 79 characters
* Always indent wrapped code for readability
* 2 blank lines before top-level function and class definitions
* 1 blank line before class method definitions
* Use blank lines in functions sparingly
* Avoid extraneous whitespace
* Don’t use whitespace to line up assignment operators (=, :)
* Spaces around = for assignment
* No spaces around = for default parameter values
* Spaces around mathematical operators, but group them sensibly
* Multiple statements on the same line are discouraged

Useful tools
=========

* IDEs : Many development environments show warnings in non-compilant code (like pyCharm). 
* [pylint](https://www.pylint.org/) : code analysis tool. 
* [pep8](https://pep8.readthedocs.io/en/release-1.7.x/) code analysis tool. 
* [autopep8](https://pypi.org/project/autopep8/) : automatically formats Python code to conform to the PEP 8 style guide.

dnerini:
has anyone used [black](https://black.readthedocs.io/) before?

It looks pretty good.
Here it's all it took to reformat the `pysteps/verification/detcontscores.py` module:
```
$>pip install black 
$>black pysteps/verification/detcontscores.py
reformatted pysteps/verification/detcontscores.py
All done! ✨ 🍰 ✨
1 file reformatted.
```
You can see the result in commit 4850ad00baa75c997d92dcccc6cf73b18efede66.

Definitely something to consider for all of us!

Edit: [here](https://news.ycombinator.com/item?id=17151813) you find an interesting discussion about black and its formatting choices.

RubenImhoff:
I think I have missed this post, but definitely worth considering! 
Black seems to work well. Good suggestion, Daniele.

aperezhortal:
The refactoring of the build system (PR #102) includes a configuration for black in the [pyproject.toml](https://github.com/pySTEPS/pysteps/blob/refactor_build_system/pyproject.toml) file.

[...]

PR #155, opened 2020-04-27
Run black and create pre-commit hooks 

aperezhortal:
In #27 we proposed to use `black` to autoformat the code and reduce some of the code-formatting in the code. This PR takes one step further and runs `black` in all the source code.

Although there are other autoformatters, `black` is simple and opinionated, allowing (intentionally) very little configuration. Although the little configuration capabilities may look like a limitation, it limits possible discussions about certain aspects of the coding style (:+1:).  

The main benefit of autoformatting all the sources now is that it will minimize the differences in future commits and enforce (some) uniformity in the coding style.

This PR also changes the maximum line length from 79 to 88 (black defaults) that produces smaller sources and deals better with many indentations levels ([here](https://youtu.be/esZLCuWs_2Y?t=1286) is a nice justification about this, min. 21:28)

It also proposes using pre-commit hooks as part of the development workflow to prevent commits that do not comply with the `black` style. This is described in an [updated version of the developers' guidelines](https://pysteps.readthedocs.io/en/black_precommit/developer_guide/contributors_guidelines.html). 

Any comments are welcome!


CONTRIBUTING.rst
```diff
@@ -99,6 +100,29 @@ If you are new to GitHub, below you can find a list of useful tutorials:
...
 
+Install pre-commit hooks
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Once you cloned the repository and the development environment is installed (and active),
+install the git pre-commit hook by executing the following command in the repository's
+root::
+
+    pre-commit hook install
+
```
RubenImhoff, line 111: This seems like a useful addition! How did you run Black to set it to a line length of 88? --> Good to add this here too, to make sure everyone does it that way. :)

CONTRIBUTING.rst
```diff
@@ -99,6 +100,29 @@ If you are new to GitHub, below you can find a list of useful tutorials:
...
+The pre-commit hooks are scripts that are executed in every commit automatically to
+identify simple issues with the code. If an issue is identified (the pre-commit script
+exits with non-zero status), the hook aborts the commit and print the error.
+Currently, Pysteps only test the code to be committed complies with black's autoformat.
+In case that the commit is aborted, you only need to run black in the entire source code.
+This can be done by running `black .` or `pre-commit run --all-files`.
```
dnerini: I think the commands in this line are not rendering as expected in the documentation: 
"This can be done by running *black .* or *pre-commit run –all-files*. "
instead of
"This can be done by running `black .` or `pre-commit run --all-files`. "

Just prepending `:code:` might solve the issue.

dnerini (approved):
I very much agree with all your points, @aperezhortal. Although black takes some pretty strong decisions on code formatting, it does provides fully reproducible code styling. And we can only profit from having a uniform code style across the repo. So, a big thumbs-up to this PR from my side!

RubenImhoff (approved):
Thanks @aperezhortal! It's a good idea to include Black (or anything similar) in our formatting process! I haven't checked all files, as you can imagine, but it looks good. I like the line length of 88, even though it is not 100% conform PEP8.

CONTRIBUTING.rst
aperezhortal: Good point. The black configuration is stored in the `pyproject.toml` file. When black is run in any pysteps folder, it loads that configuration file. Since the line-length was not specified, its default value is used (88).  The pre-commit application only installs black, but with no associated configuration. 

I updated the developers' guidelines briefly explaining this and made the line-length explicit in the `pyproject.toml` file.

aperezhortal:
> I haven't checked all files, as you can imagine, but it looks good. 

@RubenImhoff, one big advantage of black is that it checks that the reformatted code is equivalent to the original code (from an execution point of view). :tada:

pulkkins:
Thumbs up for good work! I'm ready to approve the pull request as well.

pulkkins (approved):

Issue #163, opened 2020-08-05
Padding bug in VET motion code


fionaRust:
Whilst using the VET motion code we uncovered a bug when using the padding option:

- When using a masked array as input the padding isn't added to the array, but it is still removed before returning the calculated velocity fields, resulting in velocity fields the is smaller in the x and y dimensions than the input arrays.
- When using an array without a mask the padding is added and removed correctly. 

I think the problem is with the if statement here: https://github.com/pySTEPS/pysteps/blob/master/pysteps/motion/vet.py#L517

aperezhortal:
Hi @fionaRust, welcome to PySteps! 

Indeed the portion of code that you pointed out is causing the issue since the padding is not been handled correctly. Thanks for finding the issue! I will fix it shortly.

aperezhortal:
PR #164 should close this issue. @fionaRust feel free to re-open it if you experience any problems.

PR #164, opened 2020-08-05
Fix vet padding bug (#163)

aperezhortal:
This PR fixes issue #163. 

- Fix the issue with the **vet function's** padding keyword when MaskedArrays are used. 
- Add a short test to check that the function now works correctly, and correctly pad Masked and Non-Masked input images.

[Edited]

### DIFF
--- a/pysteps/tests/test_motion.py
+++ b/pysteps/tests/test_motion.py
@@ -18,10 +18,10 @@ the retrieval.
 """
 
 from contextlib import contextmanager
-from functools import partial
 
 import numpy as np
 import pytest
+from functools import partial
 from scipy.ndimage import uniform_filter
 
 import pysteps as stp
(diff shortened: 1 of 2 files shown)


pysteps/tests/test_motion.py
```diff
@@ -324,6 +325,35 @@ def test_input_shape_checks(
...
+        motion.get_method("vet"),
+        verbose=False,
+        sectors=((16, 4, 2), (16, 4, 2)),
+        options=dict(maxiter=5)
+        # We use only a few iterations since
+        # we don't care about convergence in this test
+    )
+
+    assert precip_obs.shape == vet_method(precip_obs).shape
+    assert precip_obs.shape == vet_method(np.ma.masked_invalid(precip_obs)).shape
```
dnerini: not sure here, isn't padding = 0 by default?

dnerini (approved):
Nicely done! In essence, you make sure that both the data and the mask are padded, correct? I also appreciate a lot the addition of the test!

Just one small question from my side, otherwise looks good to go!

pysteps/tests/test_motion.py
aperezhortal: Oops, the test was not testing the padding code :confounded:
Commits 1dc4eee and 35ca32a add the tests and fix a bug with the mask padding.

aperezhortal:
Thanks for the review! 

>  In essence, you make sure that both the data and the mask are padded, correct?

That's right. I edited the PR description since it was not accurate.

dnerini:
Super, thanks! I guess this can now be merged.

PR #166, opened 2020-08-07
Fix plt motionfields

### LINKED ISSUE #165: Incorrect plot_geography call when plotting motion field over map
The plotting function in `pysteps.visualization.precipfields` and `pysteps.visualization.motionfields` are very different, and the one in `motionfields.py` doesn't match documented API. 

`motionfields.py:136` requests key `xpixelsize` and key `ypixelsize` from `geodata` but not documented, and implementation on `precipfields` is different (not using it). Later when it calls `plot_geography`, it puts 9 positional args. but `plot_geography` only accepts 6.

Overall, it looks like plotting in motionfields is broken.

dnerini:
Addresses issue #165, namely

* update interface to plot_geography()
* update interface for quiver() and streamplot() functions for consistency with plot_precip_fields()
* xpixelsize and ypixelsize are not required anymore
* update docstrings
* add tests


pysteps/visualization/motionfields.py
```diff
@@ -238,11 +209,9 @@ def streamplot(
     map=None,
     geodata=None,
     drawlonlatlines=False,
-    basemap_resolution="l",
-    cartopy_scale="50m",
     lw=0.5,
```
aperezhortal, line 222: **lw** keyword is not used in streamplot. Perhaps it should be added to the plot_geography call in streamplot? (like in quiver).

pysteps/visualization/motionfields.py
```diff
@@ -26,11 +25,10 @@ def quiver(
     map=None,
```
aperezhortal, line 25: Using **map** directly as keyword masked the build-in `map` python function. Although the code works (since the reference to map is redefined) maybe we can move map to kwargs and assign its value to a different variable name `plot_map=kwargs.pop("map")`. By doing this we don't need to change the interface and we avoid the shadowing of the python `map` function. What do you think?

aperezhortal:
Great job! I added two minor comments in the file changes. 

Also, I think that the following statement In motionfields.py, lines 332-333, have no effect, since X,Y are redefined later on. Could be? 
```
if not reproject:
    X, Y = np.meshgrid(x, y)
```

pysteps/visualization/motionfields.py
dnerini: well spotted! corrected in aaee5d1

pysteps/visualization/motionfields.py
dnerini: It's an excellent point, but unfortunately this is true not only for the quiver() function, but also for all our plotting methods.

We could implement your workaround for most cases,  but I feel that this would somewhat complicate the code. 
Alternatively we could explicitly rename the `map` argument to `plot_map`, but this will change the interface.

I understand that the current situation is not ideal, but do you see any major problem in leaving it as it is? Our local `map` argument shadows the built-in `map` method, which is not used in there. Do you think that something can go wrong by doing it?

dnerini:
Thanks for the review! 

(indeed, lines 332-333 were unnecessary, see 0ec44ee )

pysteps/visualization/motionfields.py
aperezhortal: > Do you think that something can go wrong by doing it?

I don't think so. We can leave it as it is. But in the long term, maybe we can change the interface slowly, first adding a warning that the argument name will change, and a few releases down the road, change the implementation. 
Either way, this is not an important issue to solve right now, but we should keep it in mind for future releases :)

aperezhortal (approved):

pysteps/visualization/motionfields.py
dnerini: Right, good idea. In cc505e2  I introduced a FutureWarning and internally renamed `map` to `plot_map`, do you mind having a quick look to check whether this is OK?

aperezhortal: Great! I also added some minor changes to the `plot_geography` interface in commits 9714c3f and 9d0ed8a.

PR #172, opened 2020-09-01
HDF5 OPERA IMPORTER FAILS READING QUALITY DATA

### LINKED ISSUE #171: Impossible importing quality data by importer_opera_hdf5
Hi! I tried importing quality data from hdf5 files by Italian DPC ( Dipartimento della Protezione Civile). 
These are OPERA HDF5 files conforming to the ODIM specification. 
I read rainrate data using the importer 'opera_hdf5', but not quality data.
Actually, in the implementation file of the importers (importers.py), the path where quality dataset is searched does not correspond to the correct position in the hdf5 hierarchy where quality data are placed in these files.
Consequently the quality field results a list of None, long as the number of files hdf5 wanted to read.
If I change the code in the implementation of importer 'opera_hdf5', modifying the path pointing at the quality dataset, I can read correctly quality data.
I will pull a new request about these modifies soon.
Hoping your agreement.

chiara-arpae:
Hi!

I changed the code of the implementation file for pysteps importers in the data importer from OPERA HDF5 files, 'opera_hdf5'. 
Actually the path where quality dataset is placed in OPERA HDF5 files by DPC didn't correspond to the path used in the previous code.

Consequently quality data couldn't be read and a list of None, long as the number of timesteps wanted to read, was returned instead of the expected nd-array of shape (t,x,y), where t is the number of timesteps, x and y the grid points coordinates of the domain.
Before the quality dataset was searched looking if the quantity defined in the group attributes was 'QIND', but looking in a higher level of the hierarchy. Instead quality data are in the dataset inside a group in the underlying level respect to the group containing rainrate data.

In this way, I search for quality data, looking for the group containing 'quality' in its name and then it is verified it contains 'QIND' as the quantity attribute.

I hope you appreciate the issue and receive your ouput soon. 

Fix #171


aperezhortal:
Hi @chiara-arpae, welcome to pysteps! Thanks a lot for the PR. 

I'll take a look at it shortly. Are the DPC's HDF files available online? They would be helpful to test the code and to compare them with the other OPERA files we have in the pysteps-data repo that are being used to test the library.

dnerini:
Hi @chiara-arpae, thanks a lot for the PR!!

This week unfortunately I won't be able to look into it, but I see that @aperezhortal is already on it, (thanks!). Perhaps also @pulkkins would like to have a look, since he wrote the opera importer in the first place?

We might want to add a test for reading the quality data, as I don't think we currently test for them.

chiara-arpae:
Hi!

DPC's HDF files aren't available online.

Anyway I got the permission to share with you a little dataset, relative to
entire 4/6/2020  and 5/6/2020 00:00 up to 5/6/2020 11:00 UTC.

You can access them on directories:

https://drive.google.com/drive/folders/1lkptGy1C2g48qSphowlnCq_czJ-97k6M?usp=sharing

https://drive.google.com/drive/folders/1Nt-abSl_zVr9Lim7hj_dfNNS5yS3dtGL?usp=sharing

Let me know if there are any problems in accessing these files.

Hoping these would be helpful.

Chiara






Il giorno mar 1 set 2020 alle ore 17:03 Andres Perez Hortal <
notifications@github.com> ha scritto:

aperezhortal:
@chiara-arpae That is more enough. Thanks!

The modifications work well with the DPC and the OPERA files that we have in the pysteps-data repo (PAAH opera product I think). As you pointed out, the files in the pysteps-data repo organize the data in a slightly different way. 

Below there is an overview of the context of each opera files for reference (blue square: the PAAH opera product in the pysteps-data repo, below in red: the files from the DPC). 

image

Right now, the importer looks for the quality data in two places: inside the "data" group, and in a separate dataset. However, if the quality data is found in both places, it returns the quality data from the separate dataset in the case this one is placed after the "precipitation" dataset. I am wondering if we should leave this as the default behavior, or add a keyword to control which quality flag is prioritized? Although this is not strictly a problem for the PAAH and the DPC files, I don't know if there are other Opera files using a different data layout.  What do you think @chiara-arpae, @pulkkins, @dnerini?

chiara-arpae:
Hi!

I think your proposal of introducing a keyword to optimize the search of
the quality data is a good idea, even if I imagine it would not add a great
improvement to the present modification of the code in terms of
generalization.

Actually I can't find a way to make the implementation of 'opera importer'
more general than the current one, in order to provide compatibility with
any eventual different OPERA file in importing its data.
Moreover incompatibility could not occur, we can't say.

Waiting for your decisions and news,
Good weekend!
Chiara


Il giorno gio 3 set 2020 alle ore 20:22 Andres Perez Hortal <
notifications@github.com> ha scritto:

dnerini:
In pysteps, so far,  there was no real ambition to provide general data importers: each importer was implemented to read a specific radar product and format only. The existing importers are simply a consequence of the datasets that are used by  pysteps contributors (e.g., I work with MeteoSwiss radar composites and therefore implemented the importers for those data).

If we continue to follow this same logic, the `import_opera_hdf5` importer needs to be able to read OPERA composites (i.e. the pan-european composite) in the HDF5 format only, while the DPC HDF5 files, which use a slightly different structure, would be using a separate importer called `import_dpc_hdf5` or even `import_arpae_hdf5`, for example. This means that this PR should not change the existing ìmport_opera_hdf5`, but instead implement a new one.

Please note that this is only true if the contribution by @chiara-arpae is a consequence of a legit difference between the OPERA and DPC datasets, and not because of a bug in the `import_opera_hdf5` function. 

Finally, I would be in favor of implementing a more general importer for datasets following the [ODIM specifications](https://www.eumetnet.eu/wp-content/uploads/2019/01/ODIM_H5_v23.pdf), i.e. a `import_odim_hdf5` function. 

To sum up:
- if there is nothing wrong with the current `import_opera_hdf5` importer, I suggest not modifying it (again, here we consider that "opera" refers to the pan-european composite product only, while "odim" is the data model used by opera).
- instead, we can either include a new `import_dpc_hdf5` importer or a more general `import_odim_hdf5` one, if this proves to be possible.

@chiara-arpae what do you think? @aperezhortal  @pulkkins any thoughts?

chiara-arpae:
I think including a new importer according to ODIM specifications would be
great and very helpful.

Thank you very much for your kind attention.

Good day,
Chiara



Il giorno dom 6 set 2020 alle ore 14:32 Daniele Nerini <
notifications@github.com> ha scritto:

dnerini:
Good, thanks for the feedback, @chiara-arpae.
What you suggest for this PR? Do you want to refactor you changes into the new ODIM importer?

chiara-arpae:
Ok for the refactoring.
I can write you when I'll load the changes.

Thanks,
Chiara

Il giorno mer 9 set 2020 alle ore 07:23 Daniele Nerini <
notifications@github.com> ha scritto:

pulkkins:
My suggestion is to replace the old importer (import_opera_hdf5) with the new one and rename it as import_odim_hdf5 (and modify the docstring accordingly). When I wrote import_opera_hdf5, I tried to make it as general as possible and not to include anything that is specific to the HDF5 files provided by OPERA.

I totally agree that someone needs to implement an importer that fully supports the ODIM specification. Now it only supports a small subset of it. So, the restrictions should be clearly described in the docstring if we decide to rename it as import_odim_hdf5. The main restriction is that it currently reads only one dataset from the file.

chiara-arpae:
Hi!

Thank-you very much for your suggestion.

Actually, I was about to push the modifies, consisting of one only
additional importer hdf5_opera, suitable for reading DPC hdf5 files and
separated from the existing one.
So, if you prefer, I can wait and make other modifies in order to replace
the old opera_importer with a new one, searching for quality data in
different places for example according to the filename that contains
information about the institution and using the quantity attribute
information.

Otherwise, If you prefer providing a generalization of the code by your
own, I thank you for your availability and attention to this PR anyway.

Waiting for news,
Thank you all.
Chiara



Il giorno ven 11 set 2020 alle ore 14:58 Seppo Pulkkinen <
notifications@github.com> ha scritto:

aperezhortal:
> To sum up:
> - if there is nothing wrong with the current import_opera_hdf5 importer, I suggest not modifying it (again, here we consider that "opera" refers to the pan-european composite product only, while "odim" is the data model used by opera).
> - instead, we can either include a new import_dpc_hdf5 importer or a more general import_odim_hdf5 one, if this proves to be possible.

Along these lines,  to maintain backward compatibility, we should leave the import_opera_hdf5 function for a while, and add also in parallel a general odim importer. Since the pan-european composite is a particular case of the ODIM standard, the general odim imported could have keywords that control which particular type of product is read (pan-european, DPC, generic, etc.). 

@chiara-arpae, do you want to push the changes? We can use that implementation as a starting point for the more general implementation. I have the impression that a more general ODIM implementation will take a considerable effort and it may be outside the scope of this PR.

Perhaps we can start small, merging this PR with an odim reader that deals with the pan-european and DPC format, and then, incrementally make it more general.

chiara-arpae:
Push done.

Hoping it would be useful,
thank you for your kind attention,
Chiara

Il giorno ven 11 set 2020 alle ore 15:56 Andres Perez Hortal <
notifications@github.com> ha scritto:

dnerini:
Hi @chiara-arpae 

My apologies for this lengthy process, unfortunately I had little time to dedicate to pysteps lately.

Can we all try to quickly summarize the point at which we are now? Also, I believe @aperezhortal had an idea on how to proceed with this PR? 

From what I gathered so far, we want to
- implement first version of general ODIM importer named `import_odim_hdf5` which can deal with both OPERA and DCP -> how? optional keywords?
- we keep the old `import_opera_hdf5` for backward compatibility -> make it a wrapper to `import_odim_hdf5`?

Agreed ?

pysteps/io/importers.py
```diff
@@ -1432,6 +1452,183 @@ def import_opera_hdf5(filename, qty="RATE", **kwargs):
 
     return precip, quality, metadata
 
+@postprocess_import()
+def import_DPC_hdf5(filename, qty="RATE", **kwargs):
```
dnerini: Should we use this as base for the more general ODIM importer and rename it `import_odim_hdf5` ?

pysteps/io/importers.py
```diff
@@ -1349,6 +1349,26 @@ def import_opera_hdf5(filename, qty="RATE", **kwargs):
...
+                                ) = _read_opera_hdf5_what_group(dgg[1]["what"])
+                            if qty_.decode() == "QIND":
+                                arr = dgg[1]["data"][...]
+                                mask_n = arr == nodata
+                                mask_u = arr == undetect
+                                mask = np.logical_and(~mask_u, ~mask_n)
+                                quality = np.empty(arr.shape)#, dtype=float)
+                                quality[mask] = arr[mask] * gain + offset
+                                quality[~mask] = np.nan  # a qui -----------------------------
+
```
dnerini, line 1375: Instead of adding this new block here, can we make `import_opera_hdf5` a wrapper to the new method  below?

aperezhortal:
Hi @chiara-arpae, sorry for stalling on this. 

I agree with @dnerini comments. 

The original goal of this PR was to fix the problem with the opera importer reading the quality data from the DPC files. Looking backward, commit b5aa2fe3c99a2db4f34349d64e8674b22497cbae solved that issue and `import_opera_hdf5` deals with the OPERA and DCP files. 

Although we all agreed that a general ODIM importer is a nice addition to pysteps, I think that it is better to maintain it outside the scope of this PR. 

Following on @daniele comments, a possible way to proceed with this PR is:

> implement first version of general ODIM importer named import_odim_hdf5, which can deal with both OPERA and DCP. 

The following steps can address this point: 
1) Revert to commit b5aa2fe3c99a2db4f34349d64e8674b22497cbae 
2) Rename `import_opera_hdf5` to `import_odim_hdf5`
3) Update the doc-strings to mention that the OPERA and the DPC files are correctly supported at the moment. Also make clear that other ODIM-compliant files may not be read correctly.

> we keep the old import_opera_hdf5 for backward compatibility

This is addressed by adding `import_opera_hdf5=import_odim_hdf5` in the importers module .

After this PR is merged, we can incrementally work on the more general ODIM reader by adding functionalities to `import_odim_hdf5`.

What do you think?

chiara-arpae:
Hi you all,

thank you for writing about the PR and sorry for late reply.
I agree with you.

I've just pushed new changes, following your advice about making a wrapper
and renaming the 'import_opera_hdf5' to 'import_odim_hdf5'.
For this, I'm very grateful to the computer scientist team of Arpae, who
made these changes.

I only add one more code line at
https://github.com/pySTEPS/pysteps/pull/172/commits/f9265ac6cc234572b5f09033f82d47d34fd2b53c
#f926ac in order to search for quality field in the most underlaying level
of the hierarchy only if it results None after the reading part that works
for PAAH hdf files but not for DPC files. Otherwise, if quality is not None
- that is the case of PAAH hdf files - the quality array stored will not be
overwritten.
It could be a fast and maybe nice way to avoid keywords.

What do you think?

Thank you very much for your attention,
Chiara

Il giorno sab 3 ott 2020 alle ore 21:17 Andres Perez Hortal <
notifications@github.com> ha scritto:

pysteps/io/importers.py
```diff
@@ -1260,9 +1261,10 @@ def _import_mch_geodata():
 
 
 @postprocess_import()
-def import_opera_hdf5(filename, qty="RATE", **kwargs):
-    """Import a precipitation field (and optionally the quality field) from an
-    OPERA HDF5 file conforming to the ODIM specification.
+def import_odim_hdf5(filename, qty="RATE", **kwargs):
+    """Import a precipitation field (and optionally the quality field) from a
+    HDF5 file conforming to the ODIM specification. OPERA and DPC files are
+    supported, but other ODIM-compliant files may not me read correctly.
```
aperezhortal:
```suggestion
    HDF5 file conforming to the ODIM specification. 
    **Important:** Currently, only the Pan-European (OPERA) and the 
    Dipartimento della Protezione Civile (DPC) radar composites are correctly supported.
    Other ODIM-compliant files may not be read correctly.
```

pysteps/io/importers.py
```diff
@@ -1433,6 +1455,9 @@ def import_opera_hdf5(filename, qty="RATE", **kwargs):
     return precip, quality, metadata
 
 
+import_opera_hdf5 = import_odim_hdf5
```
aperezhortal:
```suggestion
def import_opera_hdf5(filename, qty='RATE', **kwargs):
    """
    Wrapper to :py:func:`pysteps.io.importers.import_odim_hdf5`
    to maintain backward compatibility with previous pysteps versions.
    
    **Important:** Use :py:func:`~pysteps.io.importers.import_odim_hdf5` instead.
    """
    return import_odim_hdf5(filename, qty=qty, **kwargs)```

**Comment by aperezhortal** [comment]
Thanks @chiara-arpae! I left two additional suggestions. Besides that, it looks ready to merge for me.

**Review by dnerini — APPROVED** [review]

**Comment by dnerini** [comment]
Hi @chiara-arpae (and colleagues), thanks a lot for the nice contribution.
Thanks to @aperezhortal, too, for the support.

This looks good to be merged!

File: pysteps/io/importers.py [hunk 5]
```diff
@@ -1433,6 +1457,16 @@ def import_opera_hdf5(filename, qty="RATE", **kwargs):
...
 
 
+def import_opera_hdf5(filename, qty='RATE', **kwargs):
+    """
+    Wrapper to :py:func:`pysteps.io.importers.import_odim_hdf5`
+    to maintain backward compatibility with previous pysteps versions.
+    
+    **Important:** Use :py:func:`~pysteps.io.importers.import_odim_hdf5` instead.
+    """
+    return import_odim_hdf5(filename, qty=qty, **kwargs)```
```
dnerini:
```suggestion
    return import_odim_hdf5(filename, qty=qty, **kwargs)
```

dnerini:
Last detail, we forgot to update the information in `io/interface.py`. This will only take few more lines:

- [ ] add the new importer to the `_importer_methods` dictionary: `odim_hdf5=importers.import_odim_hdf5`
- [ ] update table in the `get_method` docstrings.

dnerini (changes requested):
~~Last detail, we forgot to update the information in `io/interface.py`. This will only take few more lines:~~

~~- [ ] add the new importer to the `_importer_methods` dictionary: `odim_hdf5=importers.import_odim_hdf5`~~
~~- [ ] update table in the `get_method` docstrings.~~

Edit: Well,  I suppose this is something I can myself after merging, too ;-)

dnerini (approved):

PR #173, opened 2020-09-14
Downscaling

dnerini:
* [x] add documentation
* [x] add gallery example
* [x] add test

RTD build: 
- gallery: https://pysteps.readthedocs.io/en/downscaling/auto_examples/rainfarm_downscale.html#sphx-glr-auto-examples-rainfarm-downscale-py
- reference: https://pysteps.readthedocs.io/en/downscaling/pysteps_reference/downscaling.html

### DIFF
--- a/examples/rainfarm_downscale.py
+++ b/examples/rainfarm_downscale.py
@@ -7,8 +7,8 @@ This example script shows how to use the stochastic downscaling method RainFARM
 available in pysteps.
 
 RainFARM is a downscaling algorithm for rainfall developed by Rebora et al. (2006).
-It uses Gaussian random fields which are generated using power-law spectral scaling
-in order to add realistic small-scale variability to the precipitation field.
+This method represents a realistic small-scale variability to the precipitation field 
+using Gaussian random fields generated using a power-law spectral scaling.
 
 """
 
--- a/examples/rainfarm_downscale.py
+++ b/examples/rainfarm_downscale.py
@@ -90,10 +90,11 @@ plt.tight_layout()
 # Remarks
 # -------
 #
-# The pysteps implementation of RainFARM currently covers spatial downscaling only,
-# that is, it can improve spatial resolution of a rainfall field, but unlikely
+# Currently, the pysteps implementation of RainFARM only covers spatial downscaling.
+# That is, it can improve the spatial resolution of a rainfall field. However, unlike
 # the original algorithm from Rebora et al. (2006), it cannot downscale the temporal
 # dimension.
+
 
 ###############################################################################
 # References
--- a/examples/rainfarm_downscale.py
+++ b/examples/rainfarm_downscale.py
@@ -6,6 +6,10 @@ Stochastic downscaling with RainFARM
 This example script shows how to use the stochastic downscaling method
 RainFARM available in pysteps.
 
+RainFARM is a downscaling method developed specifically for rainfall 
+using Gaussian random fields which are generated using power-law 
+spectral scaling.
+
 """
 
 import matplotlib.pyplot as plt
@@ -68,7 +72,7 @@ plot_precip_field(precip_lr, geodata=metadata_lr)
 # We can now use the RainFARM method to provide stochastic realizations
 # of the downscaled field.
 
-num_realizations = 3
+num_realizations = 2
 precip_hr = []
 for n in range(num_realizations):
     precip_hr = rainfarm.downscale(precip_lr, ds_factor=16)
--- a/examples/rainfarm_downscale.py
+++ b/examples/rainfarm_downscale.py
@@ -6,9 +6,9 @@ Precipitation downscaling with RainFARM
 This example script shows how to use the stochastic downscaling method RainFARM
 available in pysteps.
 
-RainFARM is a downscaling algorithm for rainfall developed by Rebora et al. (2006).
-This method represents a realistic small-scale variability to the precipitation field 
-using Gaussian random fields generated using a power-law spectral scaling.
+RainFARM is a downscaling algorithm for rainfall fields developed by Rebora et
+al. (2006). The method can represent the realistic small-scale variability of the
+downscaled precipitation field by means of Gaussian random fields.
 
 """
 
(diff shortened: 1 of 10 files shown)


dnerini:
@jleinonen , I realize now that your implementation of rainfarm includes spatial downscaling only, as opposed to the Rebora et al 2006 paper, which also includes temporal downscaling.

This is not a problem, I'm just wondering what's the best way to proceed. 

We could try to complete the code with the temporal part, but this risks to take some time which unfortunately I don't have at the moment.

We could simply specify in the docstring that this implementation is a simplified version which deals with spatial downscaling only. Hopefully the code will be completed and extended at some point.

What's your take on this?

dnerini:
This still misses some testing of the new module, but I'd say that we are anyway ready for a first round of review. Any feedback and suggestion for improvement is welcomed!

@jleinonen would also be great if you could have a quick look, thanks!

pysteps/downscaling/rainfarm.py
```diff
@@ -0,0 +1,124 @@
...
+from scipy.ndimage import convolve
+
+
+def _log_slope(log_k, log_power_spectrum):
+    lk_min = log_k.min()
+    lk_max = log_k.max()
+    lk_range = lk_max - lk_min
+    lk_min += (1 / 6) * lk_range
+    lk_max -= (1 / 6) * lk_range
+    dk = lk_range / 32
```
RubenImhoff: Looks like the dk-variable is not used anymore, is that correct?

pysteps/downscaling/rainfarm.py
```diff
@@ -0,0 +1,124 @@
...
+
+    selected = (lk_min <= log_k) & (log_k <= lk_max)
+    num_selected = np.count_nonzero(selected)
```
RubenImhoff: Also the num_selected-variable seems not to be used anywhere else, is that correct?

RubenImhoff (commented):
Hi @dnerini,

This is a great contribution! I also like the gallery, a very clear example of the method. 

I added two comment strings in the script for the gallery and I had two questions about the variables in rainfarm.py. So nothing major at all. :)
Regarding the rainfarm.py script, is it an idea to add some comment lines and perhaps refer to the equation numbers in the paper of Rebora et al? That may increase the readability of the script a bit. 

Other than that, good to go!

Thanks!

RubenImhoff (approved):

examples/rainfarm_downscale.py
```diff
@@ -0,0 +1,104 @@
...
+"""
+Precipitation downscaling with RainFARM
+=======================================
+
+This example script shows how to use the stochastic downscaling method RainFARM
+available in pysteps.
+
+RainFARM is a downscaling algorithm for rainfall developed by Rebora et al. (2006).
+It uses Gaussian random fields which are generated using power-law spectral scaling
+in order to add realistic small-scale variability to the precipitation field.
```
aperezhortal:
```suggestion
This method represents a realistic small-scale variability to the precipitation field 
using Gaussian random fields generated using a power-law spectral scaling.
```

examples/rainfarm_downscale.py
```diff
@@ -0,0 +1,104 @@
...
+plot_precip_field(precip, geodata=metadata)
+
+# Assign the fill value to all the Nans
+precip[~np.isfinite(precip)] = metadata["zerovalue"]
+
+###############################################################################
+# Upscale the field
+# -----------------
+#
+# In order to test our downscaling method, we first need to
```
aperezhortal:
```suggestion
# To test our downscaling method, we first need to

```

examples/rainfarm_downscale.py
```diff
@@ -0,0 +1,104 @@
...
+plt.tight_layout()
+
+###############################################################################
+# Remarks
+# -------
+#
+# The pysteps implementation of RainFARM currently covers spatial downscaling only,
+# that is, it can improve spatial resolution of a rainfall field, but unlikely
+# the original algorithm from Rebora et al. (2006), it cannot downscale the temporal
+# dimension.
```
aperezhortal:
```suggestion
# Currently, the pysteps implementation of RainFARM only covers spatial downscaling.
# That is, it can improve the spatial resolution of a rainfall field. However, unlike
# the original algorithm from Rebora et al. (2006), it cannot downscale the temporal
# dimension.

```

pysteps/downscaling/rainfarm.py
```diff
@@ -0,0 +1,124 @@
...
+    lk_sel = log_k[selected]
+    ps_sel = log_power_spectrum[selected]
+    (alpha, _) = np.polyfit(lk_sel, ps_sel, 1)
```
aperezhortal: `(alpha, _) = np.polyfit`.  suggests that `polyfit` returns a tuple of 2 different arguments while it actually returns a single array (of 2 elements).  
```suggestion
    alpha = np.polyfit(lk_sel, ps_sel, 1)[0]
```

pysteps/downscaling/rainfarm.py
```diff
@@ -0,0 +1,124 @@
...
+    r : array_like
+        Array of shape (m*ds_factor,n*ds_factor) containing
+        the downscaled field.
+
+    Notes
+    -----
+    The pysteps implementation of RainFARM currently covers spatial downscaling
+    only, that is, it can improve spatial resolution of a rainfall field, but
+    unlikely the original algorithm from Rebora et al. (2006), it cannot downscale
+    the temporal dimension.
```
aperezhortal:
```suggestion
    Currently, the pysteps implementation of RainFARM only covers spatial downscaling.
    That is, it can improve the spatial resolution of a rainfall field. However, unlike
    the original algorithm from Rebora et al. (2006), it cannot downscale the temporal
    dimension.
```

pysteps/downscaling/rainfarm.py
dnerini: Absolutely, well spotted!
see 1164fde

pysteps/downscaling/rainfarm.py
dnerini: see 1164fde

jleinonen (approved):

PR #175, opened 2020-09-25
Exporters

RubenImhoff:
The netCDF exporter still hard codes the cartesian units in meters, while the KNMI importer and projection works with kilometers now. This is a first simple implementation to start with. Let's see if we can further improve it. I set most of the other importers on meters, but I wasn't sure what the units of the MRMS and saf_crri data are. Hence, this still has to be changed, also in the testing scripts.

Furthermore, I saw that the ensemble members get the variable name 'ens_number'. The CF convention apparently recommends to use 'realization' or 'realizations'. I don't know if it is worth changing this - I personally like 'ens_number' -  but it may help with the standardization of the netCDF-exporter. 

- [x] Add 'cartesian_unit' to importers
- [x] Change hard coded unit in netCDF exporter
- [x] Change ensemble member standard name to realization to comply with the CF convention 
- [x] Correct unit in importer and test scripts of MRMS importer
- [x] Correct unit in importer and test scripts of saf_crri importer
- [x] Check for correct use of units

 @dnerini, @aperezhortal, @pulkkins, I requested you as reviewers for now. You probably have some (better) ideas on this too. And I'm sure you'll know the correct units of the other importers.


aperezhortal:
Hi @RubenImhoff, nice addition! I like the idea of including the cartesian units in the imported metadata.

The MRMS data has a regular lat/lon grid using a 0.01 degrees grid spacing (approximate equivalent to 1 km). Currently, the **pixelsize** values in the metadata are expressed in degrees (taken directly from the file), so __cartesian_units__ should be equal to "degrees". 

For the saf_crri files, it seems that the data resolution is ~3km (see abstract in https://www.tethys.cat/sites/default/files/pdf/articles/12tethys-01-eng.pdf) and they are expressed in meters by the importer ('xpixelsize'=3000.0).

RubenImhoff:
Hi @aperezhortal, 
Thanks for your response! Have you tested the netcdf exporter with the MRMS data? I.e., will it work to set the cartesian_units on 1 km, or is it wiser to stay away from that at all and just use lat-lon? 

I will implement the saf_crri units somewhere this week and I'll do some testing. :)

cvelascof:
Hi all,
I hope you all well.

> Furthermore, I saw that the ensemble members get the variable name 'ens_number'. The CF convention apparently recommends to use 'realization' or 'realizations'. I don't know if it is worth changing this - I personally like 'ens_number' - but it may help with the standardization of the netCDF-exporter.

My understanding is the 'long_name' attribute of a variable could be whatever text that best describes the data but the "standard_name" attribute must be the one recommended in the CF convention. Therefore we could assign our preferred label as "long_name" attribute to our ensembles member variable (for example "ens_member" or "member" or "realisation" or ...) but in all cases the "standard_name" attribute of the variable has to be "realization".

It would the same than what we are already doing for 'x', 'y', 'latitude' or 'longitude' coordinates.
For example here, 
https://github.com/pySTEPS/pysteps/blob/903567ea6badd48095658afcf737f96ac53f1999/pysteps/io/exporters.py#L500   


and then of course somehow the system should check for "standard_name" to identify the variable that stores the ensembles members and use that variable name instead of a hard-coded variable name

RubenImhoff:
@aperezhortal, nice work! Thanks for the contribution. I'll continue Friday with it, hopefully it is ready for review after that. :) Shouldn't be too much work.

pysteps/io/importers.py
```diff
@@ -400,14 +401,14 @@ def import_mrms_grib(filename, extent=None, window_size=4, **kwargs):
...
-        projection=proj_params,
+        projection=proj_def.strip(),
         yorigin="upper",
         threshold=_get_threshold_value(precip),
         x1=x1,
         x2=x2,
         y1=y1,
         y2=y2,
-        cartesian_unit="m"
+        cartesian_unit="m",
```
RubenImhoff: Should this be 'degrees' by the way? - In the mrms importer, I mean

aperezhortal: Yes,  since 
```
x1, y1 = pr(ul_lon, lr_lat)
x2, y2 = pr(lr_lon, ul_lat)
```
returns x and y in degrees (no projection).

RubenImhoff:
I think it is ready for review now. 

I have tested the various importers and everything seems to work. However, with a quick check of the output netCDFs in the program Panoply, I couldn't directly see if the projected cartesian coordinates of the MeteoSwiss and MRMS nowcasts are correct. @aperezhortal and @dnerini, would you be able check this with a quick test run?

dnerini (approved):
Hi @RubenImhoff
Very nice contribution, as it solves an annoying inconsistency in the metadata concerning the units of the projection, thanks! 
To me this looks ready to be merged.

aperezhortal (approved):

aperezhortal:
Thanks for fixing the units! I quickly check the cartesian coordinates for MRMS nowcasts and they expressed in degrees as expected :+1: 

It looks ready to merge for me.

RubenImhoff:
Thanks @dnerini, @aperezhortal and @cvelascof! I'll merge the PR. :)

PR #177, opened 2020-11-01
Refactor interface to plot_geography

dnerini:
Addressing FutureWarning from 1.3.2 which concerns the interface to the `plot_geography` method.

In short:
* rename `map` to `plot_map` 
* set `plot_map` as optional argument

Remaining open question: 
Should we remove option to use the deprecated basemap module and leave cartopy as unique option?

### DIFF
--- a/doc/source/pysteps_reference/visualization.rst
+++ b/doc/source/pysteps_reference/visualization.rst
@@ -5,6 +5,7 @@ pysteps.visualization
 Methods for plotting precipitation and motion fields.
 
 .. automodule:: pysteps.visualization.animations
+.. automodule:: pysteps.visualization.basemaps
 .. automodule:: pysteps.visualization.motionfields
 .. automodule:: pysteps.visualization.precipfields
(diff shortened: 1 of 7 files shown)


aperezhortal:
Nice work! For me, it looks ready to merge. 

> Remaining open question:
Should we remove option to use the deprecated basemap module and leave cartopy as unique option?

I'm in favor of removing basemap since last August they made their (likely) final release and it won't be supported anymore.

pulkkins:
> Remaining open question:
> Should we remove option to use the deprecated basemap module and leave cartopy as unique option?

For me it looks like there is no reason to use basemap anymore. In 0.18.0b1, they finally added a feature that was missing from cartopy. Namely, lon-lat labeling with different projections (https://github.com/SciTools/cartopy/pull/1117).

dnerini:
All right, thanks @aperezhortal  and @pulkkins. I'll merge this PR and open a new one for the deprecation of basemap.

PR #178, opened 2020-11-03
Implement cell tracking module

dnerini:
This PR implements the Thunderstorm Detection and Tracking (DATing) module , following the procedures used in the TRT Thunderstorms Radar Tracking algorithm (Hering et al 2004) used operationally at MeteoSwiss.

Modifications include advecting the identified thunderstorms with the optical flow obtained from pysteps, as well as additional options in the thresholding.

RTD
----
- feature doc: https://pysteps.readthedocs.io/en/cell_tracking/pysteps_reference/feature.html
- tracking https://pysteps.readthedocs.io/en/cell_tracking/pysteps_reference/tracking.html
- example: https://pysteps.readthedocs.io/en/cell_tracking/auto_examples/thunderstorm_detection_and_tracking.html

PR progress 
-------------
- [x] Create new `feature` and `tracking` modules, including the `__init__.py` and `interface.py` files
- [x] Move existing feature detection methods from `pysteps.utils.images` to `pysteps.feature`
- [x] Move LK tracking method from `pysteps.motion.lucaskanade.track_features` to `pysteps.tracking.lucaskanade.track_features`
- [x] Check requirements
- [x] Add example
- [x] Update documentation
- [x] Add/fix tests


dnerini:
Hello @feldmann-m

commits 2ffb21f, 1ddb3b3 and bb3a252  are necessary to render your example in readthedocs. 
(it's building now, let's see if it succeeds).

In particular, in 1ddb3b3 I removed from your plotting functions the part where it saved the figures. Both routine return the figure axes instead. 

In 2ffb21f I remove the downloading of the example data. This is not necessary, since the data are downloaded once for all the examples before readthedocs is build.

pysteps/feature/tstorm.py
```diff
@@ -0,0 +1,224 @@
...
+        data=None,
+        index=range(len(cell_labels)),
+        columns=["ID", "time", "x", "y", "max_x", "max_y", "max_ref", "cont"],
+    )
+    cells_id.time = time
+    for n in range(len(cell_labels)):
+        ID = n + 1
+        cells_id.ID[n] = ID
+        cells_id.x[n] = np.where(cells == cell_labels[n])[1]
+        cells_id.y[n] = np.where(cells == cell_labels[n])[0]
```
dnerini, line 207: Here `x` and `y` are in unit of number of pixels from the field origin, correct? 
This might create some confusion later on when, for example, one would try to plot the contours.

Should we instead work in geographical units? All pysteps data must come on a regular grid and the `metadata` dictionary will include the definition of the geographical projection.

pysteps/feature/tstorm.py
```diff
@@ -0,0 +1,224 @@
...
+
+    new_max = y_max, x_max
+
+    return new_max
+
+
+def get_profile(areas, binary, ref, loc_max, time, minref):
+    """
+    This function returns the identified cells in a dataframe including their x,y
+    locations, location of their maxima, maximum reflectivity and contours.
```
dnerini, line 220: Here `x` and `y` are in unit of number of pixels from the field origin, correct?
This might create some confusion later on when, for example, one would try to plot the contours.

Should we instead work in geographical units? All pysteps data must come on a regular grid and the metadata dictionary will include the definition of the geographical projection.

dnerini: At least partially addressed in f0ebdb4

examples/thunderstorm_detection_and_tracking.py
```diff
@@ -0,0 +1,199 @@
...
+for track in track_list:
+    if np.unique(track.ID) in IDs: track_filt.append(track)
+
+# Add their tracks
+plot_track(track_filt, geodata=metadata)
+plt.show()
+
+#%%
+################################################################################
+# Example with US MRMS data
```
dnerini: This is a great idea, @feldmann-m , but unfortunately loading the mrms data requires the pygrib package which is a bit problematic to install without conda (which we don't use on readthedocs). Also, the US data are somewhat heavy and might slow down the build of the documentation or event stop it due to time or memory limits.

I suggest that you only use Swiss data in this example, while the mrms dataset could become a more in-depth tutorial that we can publish as a Google Colab notebook. What do you think?

dnerini:
@feldmann-m if you were wondering why we need to avoid empty lists as default argument to a function (bcb898e), you can have a look here at the resulting documentation (see screenshot below).
If you interested to better understand the cause, you can have a look here:
https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument

image

feldmann-m:
> @feldmann-m if you were wondering why we need to avoid empty lists as default argument to a function ([bcb898e](https://github.com/pySTEPS/pysteps/commit/bcb898ed99ba4a8a2969ef85c975c24c67192621)), you can have a look here at the resulting documentation (see screenshot below).
> If you interested to better understand the cause, you can have a look here:
> https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument

Thank you for the advice! I will keep that in mind in future. Should I change this in the example as well?

dnerini:
> 
> 
> > @feldmann-m if you were wondering why we need to avoid empty lists as default argument to a function ([bcb898e](https://github.com/pySTEPS/pysteps/commit/bcb898ed99ba4a8a2969ef85c975c24c67192621)), you can have a look here at the resulting documentation (see screenshot below).
> > If you interested to better understand the cause, you can have a look here:
> > https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument
> 
> Thank you for the advice! I will keep that in mind in future. Should I change this in the example as well?

No, no need to change anything in the example script. This applies only when defining the default arguments in a function.

dnerini:
@feldmann-m can you please have a look at commit 97fb795? It seems to remove the SettingWithCopyWarnings, but would be better to double check that this did not change the behavior of your code.

RubenImhoff:
Thanks for the amazing work, @feldmann-m and @dnerini! I'll try to go through the code today. If you think the code is ready for testing, I might give it a try with the KNMI radar data.

pysteps/feature/tstorm.py
```diff
@@ -0,0 +1,320 @@
...
+...............
+:cite:`TRT2004`
+
+@author: mfeldman
+
+.. autosummary::
+    :toctree: ../generated/
+
+    detection
+    belonging
```
RubenImhoff: In the ReadTheDocs, there is no description of 'belonging'. The other three (detection, longdistance and get_profile) do have a description.

pysteps/tracking/lucaskanade.py
```diff
@@ -0,0 +1,201 @@
...
+.. _OpenCV: https://opencv.org/
+
+.. _`Lucas-Kanade`:\
+    https://docs.opencv.org/3.4/dc/d6b/group__video__track.html#ga473e4b886d0bcc6b65831eb88ed93323
+
+.. autosummary::
+    :toctree: ../generated/
+
+    track_features
+"""
```
RubenImhoff, line 20: Perhaps good to mention (for new users) here and/or in motion.lucaskanade.py that the same optical flow algorithm is used, but that the tracking functionality is defined in pysteps.tracking and vice versa for the standard motion derivation.

pysteps/feature/blob.py
```diff
@@ -0,0 +1,99 @@
...
+    .. _`feature.blob_*`:\
+    https://scikit-image.org/docs/dev/auto_examples/features_detection/plot_blob.html
+
+    Parameters
+    ----------
+    input_image : array_like
+        Array of shape (m, n) containing the input image. Nan values are ignored.
+    method : {'log', 'dog', 'doh'}, optional
+        The method to use: 'log' = Laplacian of Gaussian, 'dog' = Difference of
+        Gaussian, ''
```
RubenImhoff: What about 'doh'?

dnerini: good question, we should ask @pulkkins! But I assume these all come from the scikit-image documentation...

edit: doh = Determinant of Hessians (b68318f)

pysteps/feature/tstorm.py
```diff
@@ -0,0 +1,320 @@
...
+    for n in range(1, n_groups + 1):
+        ind = np.where(labels == n)
+        size = len(ind[0])
+        maxval = np.nanmax(input_image[ind])
+        if size < minsize:
+            binary[labels == n] = 0
+            labels[labels == n] = 0
+        if maxval < minmax:
+            binary[labels == n] = 0
+            labels[labels == n] = 0
```
RubenImhoff, line 142: Perhaps good to add some comments in these lines.

pysteps/feature/tstorm.py
```diff
@@ -0,0 +1,320 @@
...
+    minsize : float, optional
+        Minimal area for possible detected object. The default is 50 pixels.
+    minmax : float, optional
+        Minimum value of maximum in identified objects. Objects with a maximum lower
+        than this will be discarded. The default is 41 dBZ.
+    mindis : float, optional
+        Minimum distance between two maxima of identified objects. Objects with a
+        smaller distance will be merged. The default is 10 km.
+    dyn_thresh: bool, optional
+        Set to True to activate dynamic lower threshold. Restricts contours to more
```
loforest: When the dynamic threshold is activated, does it override the minmax or minref value? 
It is not clear how it operates.

pysteps/feature/tstorm.py
```diff
@@ -0,0 +1,320 @@
...
+    if not SKIMAGE_IMPORTED:
+        raise MissingOptionalDependency(
+            "skimage is required for thunderstorm DATing " "but it is not installed"
+        )
+    if not PANDAS_IMPORTED:
+        raise MissingOptionalDependency(
+            "pandas is required for thunderstorm DATing " "but it is not installed"
+        )
+    filt_image = np.zeros(input_image.shape)
+    filt_image[input_image > minref] = input_image[input_image > minref]
```
loforest: According to the definition of minref in the help, the comparison operator should be equal or greater than minref. Correct?

pysteps/tracking/tdating.py
```diff
@@ -0,0 +1,309 @@
...
+
+    Parameters
+    ----------
+    input_video : array-like
+        Array of shape (t,m,n) containing input image, with t being the temporal
+        dimension and m,n the spatial dimensions. Nan values are ignored.
+    timelist : list
+        List of length t containing string of time and date of each (m,n) field.
+    mintrack : int, optional
+        minimum length of cell-track to be counted. The default is 3.
```
RubenImhoff: What is the dimension, pixels?

pysteps/feature/tstorm.py
dnerini: 8ad3af4 should have fixed this

pysteps/tracking/tdating.py
```diff
@@ -0,0 +1,309 @@
...
+        The default is None.
+        If not None, requires that cell_list has the same length.
+    start : int, optional
+        If you wish to expand an existing list of cells, the input video must contain 2
+        timesteps prior to the merging. The start can then be set to 2, allowing the
+        motion vectors to be formed from the first three grids and continuing the cell
+        tracking from there. The default is 0, which initiates a new tracking sequence.
+    minref : float, optional
+        Lower threshold for object detection. Lower values will be set to NaN.
+        The default is 35 dBZ.
```
pulkkins, line 108: This might confuse some users. Default values are specified in dBZ. However, the expected unit of the input data is not specified anywhere. I guess it can take any input field regardless of the unit?

pulkkins: To me it looks like minref (and several other keyword arguments) are not used anywhere in the code.

pysteps/visualization/thunderstorms.py
```diff
@@ -0,0 +1,126 @@
...
+
+    PYPROJ_IMPORTED = True
+except ImportError:
+    PYPROJ_IMPORTED = False
+
+from pysteps.visualization import plot_precip_field
+
+
+def plot_track(track_list, geodata=None):
+    """
```
RubenImhoff, line 34: Shouldn't the function also contain xsize, ysize, poix and poiy, or is this part of 'geodata'?

RubenImhoff (commented):
And I finally got through all the code. :) Very nice work, @feldmann-m and @dnerini. I left a couple of small comments or questions, and I made some minor changes to the code (mostly docstring related). I have just tested the functionality with the KNMI radar data and it works, see also the figure below!

Tracks

pysteps/visualization/thunderstorms.py
dnerini: Well spotted, I need to update the docstring of this function!

dnerini: eaa2813

pysteps/feature/tstorm.py
```diff
@@ -0,0 +1,321 @@
...
+        meaningful area. The default is False.
+    output_feat: bool, optional
+        Set to True to return only the cell coordinates.
+    time : string, optional
+        Date and time as string. Used to label time in the resulting dataframe.
+        The default is '000000000'.
+
+    Returns
+    -------
+    cells_id : pandas dataframe
```
aperezhortal, line 101: Perhaps we can describe in more detail the contents (columns) of the dataframe. This will make clear what this function returns.

pysteps/tracking/lucaskanade.py
aperezhortal: Also, it can be briefly mentioned the differences between the two. To prevent any confusion.

RubenImhoff (approved):

dnerini:
This now looks to me pretty ready to be merged, what do you think, @feldmann-m ? Something's still missing?

aperezhortal:
Great contribution @dnerini and @feldmann-m!  

There are a few minor comments still unresolved. Other than that, it looks ready to go!

pysteps/feature/tstorm.py
feldmann-m: I added descriptions to lines 137 and 140 to describe the steps being done here.

pysteps/feature/tstorm.py
feldmann-m: In general it overrides the minref value, however this module has been removed for now, as I found additional issues.

pysteps/feature/tstorm.py
feldmann-m: I corrected it to >= in the code in line 120.

pysteps/tracking/tdating.py
feldmann-m: The dimension for mintrack is number of timesteps, I expanded the docstrings accordingly.

pysteps/tracking/tdating.py
feldmann-m: I expanded the description of the default input.
Thank you very much for pointing out the lack of use of the keywords. This was a mistake in the code, that has now been corrected. Unfortunately this revealed some ongoing issues in the dynthresh function, so I removed it for now.

pysteps/feature/tstorm.py
feldmann-m: I added a description of the dataframe columns to the docstrings.

feldmann-m:
> And I finally got through all the code. :) Very nice work, @feldmann-m and @dnerini. I left a couple of small comments or questions, and I made some minor changes to the code (mostly docstring related). I have just tested the functionality with the KNMI radar data and it works, see also the figure below!

Thank you very much for testing it on the KNMI data! It is great to see, that it works on other data as well. I also ran a test on MRMS data, which seems for work fine as well.

feldmann-m:
> Great contribution @dnerini and @feldmann-m!
> 
> There are a few minor comments still unresolved. Other than that, it looks ready to go!

I hope I addressed the comments in the updates from Thursday, if there are remaining unresolved issues, please let me know!

dnerini:
All right, I'm going to merge this PR, since all pending comments are resolved. 

@feldmann-m thank you so much for the hard work. I'm sure this new module will be highly appreciated and used!
Hopefully this will also inspire some new and exciting developments on thunderstorm detection, tracking, and nowcasting... we'll see! Again, well done!

PR #180, opened 2020-11-08
Deprecate Basemap dependency

### LINKED ISSUE #14: Axis tick labels with lon-lat lines
The current version of cartopy (0.16.0) can only show axis tick labels for latitudes and longitudes for two projections. Otherwise, it throws an exception:

TypeError: Cannot label gridlines on a <insert-projection-here> plot.  Only PlateCarree and Mercator plots are currently supported.

This works with basemap. Until this issue is fixed in cartopy, I suggest implementing a temporary workaround (e.g. implement a function that sets the axis labels manually).

### LINKED ISSUE #179: Deprecate basemap
> Remaining open question:
> Should we remove option to use the deprecated basemap module and leave cartopy as unique option?

For me it looks like there is no reason to use basemap anymore. In 0.18.0b1, they finally added a feature that was missing from cartopy. Namely, lon-lat labeling with different projections (https://github.com/SciTools/cartopy/pull/1117).

_Originally posted by @pulkkins in https://github.com/pySTEPS/pysteps/pull/177#issuecomment-722934843_

dnerini:
This PR removes the support for the Basemap module. As a consequence, all drawing of geographical features will be based on Cartopy only.

In detail:
- Add `drawlonlatlabels` option (closes #14)
- Include additional geographical features (reefs, minor islands) when `scale in ["10m", "50M"]`.
- Passing `geodata` without Cartopy installed silently fails, producing a georeferenced plot without any basemap (which is equivalent to pass `plot_map=None`).
- Add support for Geostationary and AzimuthalEquidistant projections.

See below for plots using pysteps example data:

test_plt_cartopy_2
test_plt_cartopy_3
test_plt_cartopy_4
test_plt_cartopy_5
test_plt_cartopy_6
test_plt_cartopy_7
test_plt_cartopy_8

### DIFF
--- a/examples/my_first_nowcast.ipynb
+++ b/examples/my_first_nowcast.ipynb
@@ -417,7 +417,7 @@
     "\n",
     "# Plot the last rainfall field in the \"training\" data.\n",
     "# train_precip[-1] -> Last available composite for nowcasting.\n",
-    "plot_precip_field(train_precip[-1], geodata=metadata, axis=\"off\", map=\"cartopy\")\n",
+    "plot_precip_field(train_precip[-1], geodata=metadata, axis=\"off\")\n",
     "plt.show()  # (This line is actually not needed if you are using jupyter notebooks)"
    ]
   },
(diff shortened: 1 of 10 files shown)


RubenImhoff (approved):
Nice work, @dnerini! Looks ready to go!

pulkkins:
Nice work! I'm wondering should we still keep the plot_map option in the interface. In addition to cartopy, there are quite many map plotting tools that could be supported in future releases of pysteps.

dnerini:
> I'm wondering should we still keep the plot_map option in the interface. In addition to cartopy, there are quite many map plotting tools that could be supported in future releases of pysteps.

It's a good point, it would also improve back-compatibility and provide a clear way to not plot a basemap (`plot_map=None`). I'll make the change asap.

edit: done in e5fed04

PR #181, opened 2020-11-15
Test with python 3.9

dnerini:
Test pysteps with python 3.9.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,12 @@ jobs:
         - PYTHON_VERSION=3.8
         - RUN_TESTS=true
 
+      - os: linux
+      python: 3.9
+      env:
+        - PYTHON_VERSION=3.9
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -32,6 +32,12 @@ jobs:
         - PYTHON_VERSION=3.8
         - RUN_TESTS=true
 
+      - os: linux
```
aperezhortal: Seems that the indentation here is two-spaces ahead.  This is confusing travis when that yml file is parsed. 

```suggestion
    - os: linux
```

aperezhortal: Committing this change triggered the Travis tests.

aperezhortal:
Sorry for the delay! I like the idea to test against python 3.9 :muscle: With the OK from the tests, it will be ready to go.

Issue #182, opened 2020-11-16
Improve the semi-Lagrangian extrapolation


pulkkins:
A closer investigation is needed to verify that the semi-Lagrangian extrapolation (pysteps.extrapolation.semilagrangian) is working correctly. It appears to produce visible artefacts in some cases. See the example below, which shows six-step extrapolation applied to a smooth precipitation field obtained by convolution. 

Similar artefacts appear with all optical flow methods, so I believe the problem is in the extrapolation. I'm suspecting that the culprit is the spline interpolation (currently implemented by using scipy.ndimage.map_coordinates). Another issue is that handling of nan values is not currently well-defined. This depends on the internal details of map_coordinates, which are not specified in the scipy documentation.
![nowcast_artefacts](https://user-images.githubusercontent.com/17096036/99272343-5dab9800-2830-11eb-9d9b-2a1615429b94.png)

To fix the above issues, I will try to implement simple bilinear interpolation with capability to handle nan values. In my opinion, this is how it should have been implemented in the first place. Spline interpolation, that is known to be less robust, should be an optional feature.

dnerini:
Interesting, thanks for pointing this out, @pulkkins !
Could it be related to the order of the spline interpolation? Currently we use `order=0` in `ndimage.map_coordinates`.

pulkkins:
The choice for order=0 was made for a good reason. If the inputs contain nan values (i.e. those outside the radar mask), choosing order=1 or order=2 yields large regions of nan values in the output. I guess this is because high-order interpolation needs a large number of neighbors.

pulkkins:
It turned out that `order=0` gives nearest interpolation. This is not what we want and probably explains those artefacts. Therefore I made a commit to use `order=1` (linear interpolation). It computes interpolated values from four nearest pixels by omitting nans. Cubic interpolation (`order=3`) with `prefilter=True` would probably be the best option but I haven't figured out how to make it handle nan values.

dnerini:
Very nice that you found a plausible cause. It would be interesting to see whether using linear interpolation instead of nearest interpolation affects the performance of the advection routine, especially in case of large domains. 

Would it be possible to compare the two settings in terms of memory and computing time?

pulkkins:
I'll do that and create a separate branch for testing. I also changed the default order back to 0 in the master. More testing is needed.

PR #184, opened 2020-11-17
Improved extrapolation

pulkkins:
This pull request aims to fix issue #182. The following changes were done to `pysteps.extrapolation.semilagrangian`:
- Allow the user to choose the interpolation order (keyword argument `order` of `scipy.ndimage.interpolation.map_coordinates`) and set its default value to 1
- Apply prefiltering if `order>1`, as suggested in the documentation of map_coordinates
- Use `order=1` for the advection field
- If `order>1`, replace nans with zeros and apply an extrapolated mask separately. this is needed because prefiltering cannot handle nan values
- Replace uppercase variable names with more descriptive ones

### DIFF
--- a/pysteps/extrapolation/semilagrangian.py
+++ b/pysteps/extrapolation/semilagrangian.py
@@ -2,8 +2,7 @@
 pysteps.extrapolation.semilagrangian
 ====================================
 
-Implementation of the semi-Lagrangian method of Germann et al (2002).
-:cite:`GZ2002`
+Implementation of the semi-Lagrangian method described in :cite:`GZ2002`.
 
 .. autosummary::
     :toctree: ../generated/
@@ -160,10 +159,18 @@ def extrapolate(
         coords_warped = [coords_warped[1, :, :], coords_warped[0, :, :]]
 
         velocity_inc_x = ip.map_coordinates(
-            velocity[0, :, :], coords_warped, mode="nearest", order=1, prefilter=False,
+            velocity[0, :, :],
+            coords_warped,
+            mode="nearest",
+            order=1,
+            prefilter=False,
         )
         velocity_inc_y = ip.map_coordinates(
-            velocity[1, :, :], coords_warped, mode="nearest", order=1, prefilter=False,
+            velocity[1, :, :],
+            coords_warped,
+            mode="nearest",
+            order=1,
+            prefilter=False,
         )
 
         velocity_inc[0, :, :] = velocity_inc_x
--- a/pysteps/extrapolation/semilagrangian.py
+++ b/pysteps/extrapolation/semilagrangian.py
@@ -120,15 +120,16 @@ def extrapolate(
     n_iter = kwargs.get("n_iter", 1)
     return_displacement = kwargs.get("return_displacement", False)
 
-    # replace nans with zeros and apply separate masking if interp_order > 1
+    # if interp_order > 1, apply separate masking to preserve nan and
+    # non-precipitation values
     if interp_order > 1:
-        mask_finite = np.isfinite(precip)
         minval = np.nanmin(precip)
-        mask_min = precip > minval
-        precip = precip.copy()
-        precip[~mask_finite] = 0.0
-        mask_finite = mask_finite.astype(float)
-        mask_min = mask_min.astype(float)
+        mask_min = (precip > minval).astype(float)
+        if allow_nonfinite_values:
+            mask_finite = np.isfinite(precip)
+            precip = precip.copy()
+            precip[~mask_finite] = 0.0
+            mask_finite = mask_finite.astype(float)
 
     prefilter = True if interp_order > 1 else False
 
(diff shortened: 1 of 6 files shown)


pulkkins:
Here are some verification results with different interpolation orders. The input data is from the NEXRAD/KFWS radar and the event used here consists of nine hours. Optical flow was computed with Lucas-Kanade and semi-Lagrangian extrapolation was applied to produce Lagrangian Persistence (LP) nowcasts to 30 min ahead. ETS was computed with threshold of 5 mm/h.
verification_ets_5
verification_mae
verification_RAPSD
Interestingly, the linear interpolation (order=1) gives slightly better forecast skill than the others. However, this is because of the loss of small-scale features (see the RAPSD figure), which is an undesired feature. On the other hand, the nearest interpolation (order=0) does not suffer from this but produces the artefacts mentioned in issue #182. So the question for choosing the default interpolation order is that do we want to maximize the forecast skill or preserve small-scale features?

pulkkins:
Here are the average computation times for a 30-minute extrapolation nowcast on my machine:
- order=0: 0.10 s
- order=1: 0.11 s
- order=2: 0.15 s
- order=3: 0.16 s

The domain size is 200x200 pixels. This test needs to be done in a larger domain (e.g. 1000x1000 pixels).

pulkkins:
A remaining problem is that the cubic interpolation produces a large number of small nonzero values to areas that should be set to zero. I noticed that this leads to incorrect autocorrelation estimates in S/PROG and STEPS and significantly lower forecast skill than with linear interpolation. ANVIL is also affected but to a lesser extent.

pysteps/extrapolation/semilagrangian.py
```diff
@@ -80,6 +81,8 @@ def extrapolate(
     vel_timestep : float
         The time step of the velocity field. It is assumed to have the same
         unit as the timesteps argument.
+    interp_order : int
+        Order of the interpolation to use. Default : 0 (nearest interpolation).
```
dnerini: the documentation doesn't match the actual default value (`interp_order=3`)

pysteps/extrapolation/semilagrangian.py
```diff
@@ -140,10 +152,10 @@ def interpolate_motion(D, V_inc, td):
         XYW = [XYW[1, :, :], XYW[0, :, :]]
 
         VWX = ip.map_coordinates(
-            velocity[0, :, :], XYW, mode="nearest", order=0, prefilter=False
+            velocity[0, :, :], XYW, mode="nearest", order=1, prefilter=False,
```
dnerini: can you briefly explain why the interpolation order here needs to be 1?

pysteps/extrapolation/semilagrangian.py
```diff
@@ -140,10 +152,10 @@ def interpolate_motion(D, V_inc, td):
         XYW = [XYW[1, :, :], XYW[0, :, :]]
 
         VWX = ip.map_coordinates(
```
dnerini: we could probably benefit from a better variable naming, `XYW`, `VWX` and so on are not very clear. Can we do something about it?

pysteps/extrapolation/semilagrangian.py
```diff
@@ -113,6 +116,15 @@ def extrapolate(
...
     return_displacement = kwargs.get("return_displacement", False)
 
+    # replace nans with zeros and apply separate masking if interp_order > 1
+    if interp_order > 1:
+        mask = np.isfinite(precip)
+        precip = precip.copy()
+        precip[~mask] = 0.0
+        mask = mask.astype(float)
+
+    prefilter = True if interp_order > 1 else False
```
dnerini, line 139: could you please briefly explain what the prefilter does? should we mention it in the docstrings?

dnerini (commented):
Very nicely done @pulkkins , and thanks a lot for the detailed analysis, too! 

I'm left with only one final doubt, if `order=3` produces some undesired artifacts, is it a good idea to set it as default value?

dnerini:
Also, the test are failing because of a ValueError in `test_nowcast_sseps`:

`ValueError: field must not contain NaNs`

is it possible that the behavior wrt nans changed in `pysteps.extrapolation.semilagrangian`?

pysteps/extrapolation/semilagrangian.py
pulkkins: Fixed.

pysteps/extrapolation/semilagrangian.py
pulkkins: Fixed.

pulkkins:
I changed the default interpolation order to 1. This choice has the best computational performance and does not produce any artefacts, though it leads to a minor loss of small-scale variability. I believe this is not an issue in most use cases.

pysteps/extrapolation/semilagrangian.py
pulkkins: Since we are integrating along the advection field, using the nearest interpolation might not be a good idea because the errors accumulate over lead time. I noticed that setting order=0 caused artefacts in some cases but they disappeared by setting order=1.

pysteps/extrapolation/semilagrangian.py
pulkkins: I'm not sure if there is any need to mention this in the docstrings. As stated in the documentation of `scipy.ndimage.interpolation.map_coordinates`, prefiltering is necessary when order>1. However, the details are not explained. I tested order=3 without prefiltering. Compared to order=1, this produced a very smooth output with even more loss of spectral power.

pysteps/extrapolation/semilagrangian.py
```diff
@@ -109,9 +115,23 @@ def extrapolate(
 
     # defaults
     verbose = kwargs.get("verbose", False)
-    D_prev = kwargs.get("D_prev", None)
+    displacement_prev = kwargs.get("displacement_prev", None)
```
aperezhortal, line 118: I like the new keyword name. However, it may introduce backward compatibility problems. If the old **D_prev** keyword is used, it is ignored without any warning and the function behaves as if displacement_prev=None. One solution could be showing a warning message if the D_prev keyword is found in `**kwargs`, explaining that the keyword name was changed, and D_prev is being ignored.

pulkkins: Good point. A warning message is now printed if D_prev is in `**kwargs`.

dnerini:
Thanks @pulkkins, this now seems ready to be merged into master.

dnerini (approved):

PR #186, opened 2020-11-21
Fix quiver streamplot bug

aperezhortal:
Fix bug in quiver and streamplot functions that modified the motion vector input for some situations.


dnerini (approved):
Thanks @aperezhortal for fixing this!  
Also, very well done by including a specific check in the tests.

This looks good to me to go.

PR #190, opened 2020-12-02
Run latest version of black and code cleaning

aperezhortal:
Run the latest version of black (20.8b1) on the entire code. 

After this PR is merged, running the same version of black for the other PR will help to minimize the file differences. 

[Update] 
Tuples, lists, or function arguments ending with a "," are reformatted by black to present one argument by line, even if the line length is less than 88. 
Removing the ending "," keeps everything in one line if it is less than 88 characters.

For example, before black:
```python
(b, 2.0, False, np.array([[0.0, 0.0, 1.0], [1.0, 1.0, 1.0], [1.0, 1.0, 1.0]]),)),    
```
After black
```python
(
     b,
     2.0,
     False,
     np.array([[0.0, 0.0, 1.0], [1.0, 1.0, 1.0], [1.0, 1.0, 1.0]]),
),
```
These, and other, cases will be addressed manually after #178 and #191 are merged to avoid any conflicts.

### Checklist

- [x] PR #178 merged
- [x] PR #191 merged
- [x] Run black
- [x] Add utf-8 encoding at the top of each source file
- [x] Remove unused variables and imports
- [x] Remove some empty lines in the docstrings (end of docstrings, after each parameter description)
- [x] Fix ` : ` in all docstrings


dnerini:
Additionally, we should use this PR to correct the formatting of the variable specification in all docstrings:
 it's `variable: type` instead of  `variabe : type`.

The blank space before the colon renders documentation in the wrong way.
For example, the docstrings in anvil.py

```
    Parameters
    ----------
    vil : array_like
        Array of shape (ar_order+2,m,n) containing the input fields ordered by
        timestamp from oldest to newest. The inputs are expected to contain VIL
        or rain rate. The time steps between the inputs are assumed to be regular.
```
renders the in the following way:

> Parameters
>     `vilarray_like`
>         Array of shape (ar_order+2,m,n) containing the input fields ordered by timestamp from oldest to newest. The inputs are expected to contain VIL or rain rate. The time steps between the inputs are assumed to be regular.

The same problem appears pretty much everywhere. I wonder why we didn't notice before, maybe it's linked to a new version in Sphinx or something.

aperezhortal:
While I was working on this PR, I found some inconsistencies between the latest black version that I have installed, and the one used by the pre-commit hooks. Previously, we have been using a mutable tag to define the version to installed by the pre-commit hook (`stable` revision). This is actually not recommended in the pre-commit hook documentation:

> pre-commit assumes that the value of rev is an immutable ref (such as a tag or SHA) and will cache based on that. Using a branch name (or HEAD) for the value of rev is not supported and will only represent the state of that mutable ref at the time of hook installation (and will NOT update automatically).

Also, it seems that the `stable` branch in the black repository is behind the latest release in PyPi. 

Hence, this PR also updates the pre-commit hook configuration to the latest black release (rev: `20.8b1`).  The black version (and the configuration file) used by the pre-commit hook can be updated periodically by running: `pre-commit autoupdate`
The latter step is also added to the developers' guidelines.

dnerini (approved):
Looks all good to me! Thanks a lot for going through all these somewhat tedious but important changes!!

Issue #197, opened 2020-12-30
Call of `basemaps.plot_geography` clears existing cartopy axis


dnerini:
Original report
----------------
(by Valentina Gregori)

> 
> Hi everybody, first of all I want to thank you all for the last release of pysteps and for the work you do to keep it update! However, while testing it in my code, I found that I'm not able anymore to overlay the quiver plot with a precipitation plot. In particular the following lines return only the quiver plot, but not the reflectivity field:
> ```
> ax = plot_precip_field(R, geodata=metadata, units='mm/h')
> quiver(V, ax=ax, geodata=metadata, step=25)
> ```
> and I noticed the same issue with the "plot_optical_flow.py" example.
> Thanks in advance for the help!

Explanation
------------

With version 1.4, by default all visualization routines try to plot a basemap with Cartopy. Therefore, calling `plot_precip_field` and `quiver` in succession (with Cartopy installed) produces a double call to `basemaps.plot_geography`.  The latter will clear any existing axis to initialize a new Cartopy axis, which explains the issue described above.

Workaround
--------------
(by @aperezhortal )

An easy workaround consists in forcing `motion.quiver` to not plot a basemap:

```
ax = plot_precip_field(R, geodata=metadata, units='mm/h')
quiver(V, ax=ax, geodata=metadata, step=25, map_kwargs=dict(plot_map=None))
```

Permanent solution
---------------------
There are a number of possible solutions that we can implement:

1. `motion.quiver` should not try to plot a basemap by default (assuming that in most cases this is called after `plot_precip_field`)
2.  implement a check for a cartopy axis in `basemaps.plot_geography`: if the current axis is a Cartopy axis, then the axis doesn't need to be cleared (but this would not prevent unnecessary calls to `plot_geography`)

dnerini:
#199 closes this issue.

PR #199, opened 2021-01-03
Refactor visualization module

aperezhortal:
# Refactor the visualization module

## 1) Refactor visualization.utils functions 

- parse_proj4_string: A more robust implementation of this function is implemented in pyproj. The refactored implementation uses a direct call to the pyproj function.
- proj4_to_cartopy: Refactor the proj4_to_cartopy function
  - Use the parse_proj4_string to construct a dictionary with the proj definitions
  - Simplify the function code by using additional dictionaries to relate the Pyproj and Cartopy parameters. 
  - Add support for additional cartopy keywords defined in the Proj4 definitions.
- Add two plotting utility functions to avoid code duplication and to improve the axis handling:
  - get_basemap_axis: This function safely gets a basemap axis. If the current axis is not a cartopy axis, it creates a basemap. Otherwise, the current axis is returned.
  - get_geogrid: Returns the x,y grid used in the plots along with additional metadata. 
-  Solves issue #197.

## 2) Refactor plotting functions

- Refactor the plotting functions to use the new plotting utils
- Merge `_plot_field_pcolormesh` and `_plot_field` in a single function.
- Avoid plotting the no-data mask separately. Instead, plot it in the `_plot_field` function. 
- Merge `quiver` and `streamplots` functions into a single one called `plot_motion` for avoiding code duplication.
The `quiver` and `streamplot` functions are now wrappers for `plot_motion`.
- **IMPORTANT**
  - Rename the `type` keyword in  `plot_precip_field` to `ptype` to avoid shadowing the built-in function with the same name. Add the corresponding deprecation warning.

## 3) Define a unique zorder for each type of plot

- Basemap features zorder
  - ocean: 0
  - land: 0
  - lakes: 0
  - rivers_lake_centerlines: 0
  - coastline: 15
  - cultural: 15
  - reefs: 15
  - minor_islands: 15
- precipitation: 10
- quiver: 20
- stream function: 30
- tracks and contours: 40

### DIFF
--- a/pysteps/visualization/basemaps.py
+++ b/pysteps/visualization/basemaps.py
@@ -11,6 +11,7 @@ Methods for plotting geographical maps using Cartopy.
     plot_geography
     plot_map_cartopy
 """
+from cartopy.mpl.geoaxes import GeoAxesSubplot
 from matplotlib import gridspec
 import matplotlib.pylab as plt
 import numpy as np
@@ -33,7 +34,18 @@ except ImportError:
 
 from . import utils
 
-VALID_BASEMAPS = ["cartopy"]
+VALID_BASEMAPS = ("cartopy",)
+
+#########################
+# Basemap features zorder
+# - ocean: 0
+# - land: 0
+# - lakes: 0
+# - rivers_lake_centerlines: 0
+# - coastline: 2
+# - cultural: 2
+# - reefs: 2
+# - minor_islands: 2
 
 
 def plot_geography(
@@ -60,7 +72,7 @@ def plot_geography(
         The PROJ.4-compatible projection string.
     extent: scalars (left, right, bottom, top)
         The bounding box in proj4str coordinates.
-    lw: float, optional
+    lw: float, optional`
         Linewidth of the map (administrative boundaries and coastlines).
     drawlonlatlines: bool, optional
         If set to True, draw longitude and latitude lines.
@@ -113,13 +125,6 @@ def plot_geography(
         )
 
     crs = utils.proj4_to_cartopy(proj4str)
-
-    # Replace current axis
-    if subplot is None:
-        cax = plt.gca()
-        subplot = cax.get_subplotspec()
-        cax.clear()
-        cax.set_axis_off()
 
     ax = plot_map_cartopy(
         crs,
@@ -182,17 +187,18 @@ def plot_map_cartopy(
             " but it is not installed"
         )
 
-    # Replace current axis
     if subplot is None:
-        cax = plt.gca()
-        subplot = cax.get_subplotspec()
-        cax.clear()
-        cax.set_axis_off()
-
-    if isinstance(subplot, gridspec.SubplotSpec):
-        ax = plt.subplot(subplot, projection=crs)
+        ax = plt.gca()
     else:
-        ax = plt.subplot(*subplot, projection=crs)
+        if isinstance(subplot, gridspec.SubplotSpec):
+            ax = plt.subplot(subplot, projection=crs)
+        else:
+            ax = plt.subplot(*subplot, projection=crs)
+
+    if not isinstance(ax, GeoAxesSubplot):
+        ax = plt.subplot(ax.get_subplotspec(), projection=crs)
+        # cax.clear()
+        ax.set_axis_off()
 
     ax.add_feature(
         cfeature.NaturalEarthFeature(
--- a/pysteps/visualization/basemaps.py
+++ b/pysteps/visualization/basemaps.py
@@ -281,12 +281,12 @@ def plot_map_cartopy(
         )
 
     if drawlonlatlines:
-        gl = ax.gridlines(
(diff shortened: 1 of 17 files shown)


pysteps/visualization/basemaps.py
```diff
@@ -94,35 +115,25 @@ def plot_geography(
...
-        cax = plt.gca()
-        cartopy_subplot = cax.get_subplotspec()
-        cax.clear()
-        cax.set_axis_off()
-
     ax = plot_map_cartopy(
         crs,
         extent,
-        cartopy_scale,
+        scale,
```
dnerini, line 134: The `scale` argument is very much specific to Cartopy and a different basemap plotting library might use a different convention, that's why the argument was previously named `cartopy_scale`.

pysteps/visualization/basemaps.py
```diff
@@ -11,13 +11,13 @@
     plot_geography
     plot_map_cartopy
 """
+from cartopy.mpl.geoaxes import GeoAxesSubplot
```
dnerini: As Cartopy is an optional dependency, this import should be included in a try ... except ImportError ... statement.

pysteps/visualization/utils.py
```diff
@@ -10,10 +10,19 @@
     parse_proj4_string
     proj4_to_cartopy
     reproject_geodata
+    get_geogrid
+    get_basemap_axis
 """
+import warnings
+
 import numpy as np
+from cartopy.mpl.geoaxes import GeoAxesSubplot
```
dnerini: Here too, Cartopy is an optional dependency

pysteps/visualization/utils.py
```diff
@@ -28,6 +37,41 @@
...
+PYPROJ_GLOB_KWRDS_TO_CARTOPY = {
+    "a": "semimajor_axis",
+    "b": "semiminor_axis",
+    "datum": "datum",
+    "ellps": "ellipse",
+    "f": "flattening",
+    "rf": "inverse_flattening",
+}
+
+PYPROJ_PROJECTION_TO_CARTOPY = dict(
```
dnerini: Similarly to above, this dictionary can only be defined if Cartopy is installed.

dnerini (commented):
Well, this is really excellent work Andres and a very much needed refactoring of the visualization module! It brings a lot of clarity in the code and solves both long standing and newer issues, thanks a lot for the contribution!

I also still need to test it directly myself with some code and data (including the pysteps examples), but for the moment I went through the PR and left few minor comments.

dnerini:
> 
    Basemap features zorder
        ocean: 0
        land: 0
        lakes: 0
        rivers_lake_centerlines: 0
        coastline: 2
        cultural: 2
        reefs: 2
        minor_islands: 2
    precipitation: 10
    quiver: 20
    stream function: 30

I like the way you formalized the zorder conventions used when plotting. However, personally I like to see the national boundaries on top of the precipitation field: this allows to keep a geographical reference even when the precipitation field is very large and covers most of the domain. Hence, I'd suggest to set the coastline, cultural, reefs and minor_islands features with a zorder > 10, what do you think?

aperezhortal:
> I'd suggest to set the coastline, cultural, reefs and minor_islands features with a zorder > 10, what do you think?

:+1:  I'll make the changes.

dnerini:
Hi @aperezhortal, thanks for including my previous comments in this PR.

I had a look at the examples rendered from this branch, and unfortunately, it seems that something's still off, for example in thunderstorm_detection_and_tracking.py:

Figure 1, v:latest:
image

Figure 1, v:refactor_visualization_module:
image

I haven't had the chance to investigate the cause of this yet, but hopefully it relates to something minor.

aperezhortal:
Hi @dnerini, thanks for spotting this! The `get_geogrid` function was returning the wrong coordinates units when cartopy was not installed. I also updated the `thunderstorms` visualization module to deal with the geogrid=None case and remove the catch for the `UnsupportedSomercProjection` exception. A workaround was previously implemented and that exception was never risen.

Now the documentation is almost equal to the main branch. You may note small differences in the motion vectors since now they are plotted using the grid coordinates with the same y-origin of the data. Thus, the "steps" starts on the y-origin passed in the geodata (upper) instead of starting at the bottom as before.

dnerini:
This looks really great, Andres. I noticed only one last inconsistency with the current master branch, Figure 4 in plot_steps_nowcast.py:

v:latest
image

v:refactor_visualization_module
image

Notice the reddish background color in the second image? I think the transparency for zero probabilities is not working properly.
Digging a bit into the code, I figured that `cmap.set_under("white", alpha=0)` was missing when plotting probabilities. It also seems that with `cmap.set_under`, it is not necessary to manually flag all values below the minimum threshold to obtain the desired transparency. 

I implemented my suggested changes in 9a0049c. Feel free to modify or revert them as needed.

dnerini (approved):

PR #202, opened 2021-01-07
Test steps netcdf exporter

### LINKED ISSUE #200: netCDF exporter does not contain any time steps
I may have a potential issue:
I have a test script for pysteps to make ensemble nowcasts, just to check whether everything works fine. It saves the nowcasts in a netcdf, but the netcdf does not contain any timesteps in the rainfall data when using pysteps 1.4. It seems that this is caused by the changes in ./nowcasts/steps.py. If I use steps.py from version 1.3, the script works fine and the output netcdf contains two timesteps with the ensemble nowcast in it. I can't figure out why. Perhaps my script is just outdated, but maybe it is an actual bug. I have put the script in the comment below, it uses the KNMI data from the pysteps data repository, so it should be relatively easy to run.

dnerini:
Closes #200 by including a specific test for the use of a callback function to export a STEPS forecast to netcdf.

Additionally, 42c44d5 fixes a bug introduced by cf34300 which caused an error when a callback was passed with `return_output=True`. 

- [x] @pulkkins, can you quickly have a look at 42c44d5 ? 
- [x] @RubenImhoff can you verify that the test would have caught the issue described in #200?

### DIFF
--- a/.github/workflows/test_pysteps.yml
+++ b/.github/workflows/test_pysteps.yml
@@ -62,8 +62,7 @@ jobs:
         run: |
           brew update-reset
           brew update
-          brew install gcc@9
-          gcc-9 --version
+          gcc-9 --version || brew install gcc@9
           pip install .
 
       - name: Install pygrib (not win)
--- a/pysteps/nowcasts/steps.py
+++ b/pysteps/nowcasts/steps.py
@@ -810,11 +810,12 @@ def forecast(
             R_f_stacked = np.stack(R_f_)
             if R_f_stacked.shape[1] > 0:
                 callback(R_f_stacked.squeeze())
-            R_f_ = None
 
         if return_output:
             for j in range(n_ens_members):
                 R_f[j].extend(R_f_[j])
+
+        R_f_ = None
 
     if measure_time:
         mainloop_time = time.time() - starttime_mainloop
--- a/pysteps/tests/test_nowcasts_steps.py
+++ b/pysteps/tests/test_nowcasts_steps.py
@@ -154,7 +154,9 @@ def test_steps_callback(tmp_path):
     assert precip_netcdf.shape[2:] == field_shape, "Wrong field shape"
 
     # assert that the saved output is the same as the original output
-    assert np.allclose(precip_netcdf, precip_output, equal_nan=True), "Wrong output values"
+    assert np.allclose(
+        precip_netcdf, precip_output, equal_nan=True
+    ), "Wrong output values"
 
     # assert that leadtimes and timestamps are as expected
     td = timedelta(minutes=timestep)
--- a/pysteps/tests/test_nowcasts_steps.py
+++ b/pysteps/tests/test_nowcasts_steps.py
@@ -111,7 +111,7 @@ def test_steps_callback(tmp_path):
     motion_field = np.zeros((2, *field_shape))
 
     exporter = io.initialize_forecast_exporter_netcdf(
-        outpath=tmp_path,
+        outpath=tmp_path.as_posix(),
         outfnprefix="test_steps",
         startdate=startdate,
         timestep=timestep,
@@ -125,7 +125,7 @@ def test_steps_callback(tmp_path):
     def callback(array):
         return io.export_forecast_dataset(array, exporter)
 
-    out = nowcasts.get_method("steps")(
+    precip_output = nowcasts.get_method("steps")(
         precip_input,
         motion_field,
         timesteps=n_timesteps,
@@ -134,6 +134,7 @@ def test_steps_callback(tmp_path):
         timestep=timestep,
         seed=42,
         n_ens_members=n_ens_members,
+        vel_pert_method=None,
         callback=callback,
         return_output=True,
     )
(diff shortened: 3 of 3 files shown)


dnerini:
Tests are currently failing for python 3.8 because of an AttributeError in `decorators.py`:72:

```
    @wraps(importer)
    def _import_with_postprocessing(*args, **kwargs):
    
        precip, *other_args = importer(*args, **kwargs)
    
        _dtype = kwargs.get("dtype", dtype)
    
        accepted_precisions = ["float32", "float64", "single", "double"]
        if _dtype not in accepted_precisions:
            raise ValueError(
                "The selected precision does not correspond to a valid value."
                "The accepted values are: " + str(accepted_precisions)
            )
    
        if isinstance(precip, np.ma.MaskedArray):
            invalid_mask = np.ma.getmaskarray(precip)
            precip.data[invalid_mask] = fillna
        else:
            # If plain numpy arrays are used, the importers should indicate
            # the invalid values with np.nan.
            _fillna = kwargs.get("fillna", fillna)
            if _fillna is not np.nan:
                mask = ~np.isfinite(precip)
                precip[mask] = _fillna
    
>       return (precip.astype(_dtype),) + tuple(other_args)
E       AttributeError: 'NoneType' object has no attribute 'astype'

/usr/share/miniconda/envs/test-environment/lib/python3.8/site-packages/pysteps/decorators.py:72: AttributeError
```

@aperezhortal do you have any hint on this?

aperezhortal:
I was able to reproduce the error in python 3.8. It seems to be originated in the `import_netcdf_pysteps` importer, on line 206:  `metadata["threshold"] = np.nanmin(precip[precip > np.nanmin(precip)])`. The `precip` variable has all NaN values, producing the following error:

```There was an error processing the file zero-size array to reduction operation fmin which has no identity"```

In case of failure, by default, the `import_netcdf_pysteps` catches the error, print a warning, and returns `None, None, None`. The first returned None value produced the error on the  `_import_with_postprocessing` decorator since it expects an array for the precipitation field.

Though, I still couldn't find why this error is only raised in python 3.8 and not in 3.6.

I'll dig a little bit more into it.

**Comment by RubenImhoff** [comment]
Nice work! I'll try to take a look at it today. :)

**Review by RubenImhoff — APPROVED** [review]
Great work, @dnerini! This tester would find the issue in the future.

**Comment by dnerini** [comment]
> I was able to reproduce the error in python 3.8. It seems to be originated in the `import_netcdf_pysteps` importer, on line 206: `metadata["threshold"] = np.nanmin(precip[precip > np.nanmin(precip)])`. The `precip` variable has all NaN values, producing the following error:
> 
> `There was an error processing the file zero-size array to reduction operation fmin which has no identity"`
> 
> In case of failure, by default, the `import_netcdf_pysteps` catches the error, print a warning, and returns `None, None, None`. The first returned None value produced the error on the `_import_with_postprocessing` decorator since it expects an array for the precipitation field.
> 
> Though, I still couldn't find why this error is only raised in python 3.8 and not in 3.6.
> 
> I'll dig a little bit more into it.

Thanks Andres for having a look. Unfortunatly I can't manage to reproduce the error in my 3.8 environment. Can you please share the steps you used?

**Comment by aperezhortal** [comment]
These are the steps I followed:
```
conda create -n test_py38 python=3.8

conda activate test_py38

conda install cython numpy jsmin jsonschema matplotlib netCDF4 opencv pillow pyproj pygrib scipy dask pyfftw cartopy h5py PyWavelets pandas scikit-image pytest pytest-cov

pip install git+https://github.com/pySTEPS/pysteps.git@test-netcdf-exporter
```
Then, from from the tests directory:
```
pytest test_nowcasts_steps.py
```

The error seems to related to the new version of scipy (1.6.0) that introduced some improvements to `scipy.ndimage` (see [here](https://github.com/scipy/scipy/releases/tag/v1.6.0)). If scipy is downgraded to v1.4 (used for the python 3.6 env in conda), the tests pass.

dnerini:
Thanks @aperezhortal for the help, I was able to reproduce the error by following your steps.

This looks a bit complicated and probably deserves a separate issue. For the moment I was able to avoid the error by setting `vel_pert_method=None` instead of the default `vel_pert_method="bps"` when calling the `nowcasts.steps` (see ef6288f). From what I understand, the AttributeError mentioned above is caused by 

1. output of `steps.forecast` being all NaNs (this is what @aperezhortal noticed in the first place) with scipy>=1.6.
2. which is caused by the call to `extrapolator_method(R_f_ip, V_pert, [t_diff_prev], **extrap_kwargs_)` (L#757 of steps.py) (Note that this is calling `semilagrangian.extrapolate` )
3. which caused by the variable `V_pert` being all NaNs
4. which is returned by `generate_vel_noise(vps[j], t_total[j] * timestep)` (L#754 in steps.py)
5. which is caused by lines `N = linalg.norm(V, axis=0)` and `V_n = V / np.stack([N, N])` (L#127-128 in noise/motion.py) when variable `V` is all zeros (as in my test case, since I set the motion field to zero).

Worth noticing that with scipy<1.6, step 2 seems to return a field of finite values instead, which doesn't trigger the final AttributeError. I need to look into this a bit more, but it seems that the behavior of `semilagrangian.extrapolate` with respect to input NaNs changed. Maybe related to [recent changes](https://github.com/scipy/scipy/releases/tag/v1.6.0) in `map_coordinates`? @pulkkins

pulkkins (approved):

PR #206, opened 2021-01-31
Update install docs

loforest:
I created a pysteps environment from the yml file and installed pysteps in development mode (pip -e). 
When importing pysteps I got this error:
ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 88 from C header, got 80 from PyObject
As suggested by @dnerini, the error was solved by starting from an empty environment instead of the one from yml.
I took the opportunity to update the installation/contribution docs and add some information here and there.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -66,17 +66,9 @@ environments on your local computer.
 Before continuing, Mac OSX users also need to install a more recent compiler.
 See instructions here: :ref:`OSX users <install_osx_users>`.
 
-The developer environment can be installed in two different ways. In case of installation
-or import errors, try both approaches. The first approach starts by creating an empty
-python environment::
-
-    conda create -n pysteps_dev python=3.8
-
-Pip will take care to install the necessary dependencies in this environment (see below).
-
-The second approach creates the environment from
+The developer environment can be created from the file
 `environment_dev.yml <https://github.com/pySTEPS/pysteps/blob/master/environment_dev.yml>`_
-file in the project's root directory running the command::
+in the project's root directory by running the command::
 
     conda env create -f environment_dev.yml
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.rst
```diff
@@ -63,7 +63,18 @@ Conda quickly installs, runs and updates packages and their dependencies.
...
+Before continuing, Mac OSX users also need to install a more recent compiler.
+See instructions here: :ref:`OSX users <install_osx_users>`.
+
+The developer environment can be installed in two different ways. In case of installation
+or import errors, try both approaches. The first approach starts by creating an empty
+python environment::
+
+    conda create -n pysteps_dev python=3.8
+
+Pip will take care to install the necessary dependencies in this environment (see below).
```
aperezhortal: Although this solves the reported issue, mixing pip and conda may cause issues in the future ([more info here](https://www.anaconda.com/blog/using-pip-in-a-conda-environment)).  For example, when a package is updated using `conda update <a packages>` and a package installed by pip is a dependency.

Perhaps this can be described as a second option in the case that using `environment_dev.yml` fails.

aperezhortal:
Hi Loris, I like the idea of including in the documentation a workaround just in case that using the enviroment.yml file does not work. I left a small comment above about possible issues mixing conda and pip.

I couldn't reproduce the error on my computer, but the windows-py3.8 tests are failing with the error that you mentioned. Maybe this is related to the issues in the [numpy-feedstock in conda-forge](https://github.com/conda-forge/numpy-feedstock/runs/1802141708) with the failed tests. Perhaps this error goes away with a newer NumPy minor or build release.

CONTRIBUTING.rst
loforest: I agree, it is a good idea to mention this only as the second option. In fact, starting from an empty conda env followed by pip will increase the chances of having to install other conda packages later. The article you sent says to install pip packages only after having installed all the conda ones.

loforest: I agree, it is a good idea to mention this only as the second option. In fact, starting from an empty conda env followed by pip will increase the chances of having to install other conda packages later. The article you sent says to install pip packages only after having installed all the conda ones.

aperezhortal:
Thanks Loris for updating the docs. Now all the tests are passing (maybe numpy was updated).

PR #207, opened 2021-02-07
Probability forecasts

### LINKED ISSUE #196: Local Lagrangian probability forecasts
I was wondering whether it could be possible to extend the `extrapolation.semilagrangian.extrapolate` method to include a stochastic component. This could be implemented as random deviation around the coordinates at which the precipitation field is evaluated. In practice, line 213

https://github.com/pySTEPS/pysteps/blob/e2c3484ef234e3ef1ac8f1f11b1e64bc6911f35c/pysteps/extrapolation/semilagrangian.py#L213

could become:

```
coords_warped = xy_coords + displacement + random_offset
```

where `random_offset` is a random displacement within an area which is proportional to the lead time (typically 1 km/min):

```
def sample_unit_circle(rows, cols):
    # source: https://stackoverflow.com/questions/46996866/sampling-uniformly-within-the-unit-circle
    length = np.sqrt(np.random.uniform(0, 1, size=(rows, cols)))
    angle = np.pi * np.random.uniform(0, 2, size=(rows, cols))
    dx = length * np.cos(angle)
    dy = length * np.sin(angle)
    return np.stack((dx, dy))

random_offset = sample_unit_circle(*xy_coords.shape) * td * 5
```

As a result, each extrapolation would sample the area around the original coordinate as in the Local Lagrangian approach of Germann and Zawadzki (2004). Worth noting that this approach would provide an efficient way to compute **point** probability forecasts (no spatial nor temporal correlation is considered).

There are of course a number of open issues, for example:

- How to efficiently account for the borders of the image?
- What's the best way to sample the offsets? Using a 2D Gaussian instead of uniform sampling could be a better idea.

I'll be happy to give it a try myself, but first it'd be good to hear your comments on this. Do you think this would be useful? Would the proposed implementation above work? Is is a good idea to try to implement it in `extrapolation.semilagrangian` or should we create a separate routine instead? 

In particular, I suspect that @pulkkins and @loforest might have something to say in this regard ;-)

Reference
----------
- Germann, U., & Zawadzki, I. (2004). Scale Dependence of the Predictability of Precipitation from Continental Radar Images. Part II: Probability Forecasts, Journal of Applied Meteorology, 43(1), 74-89. Retrieved Dec 19, 2020, from https://journals.ametsoc.org/view/journals/apme/43/1/1520-0450_2004_043_0074_sdotpo_2.0.co_2.xml

dnerini:
Resolve #196 

- [x] Complete documentation
- [x] Add unit tests
- [x] Add example

RDT: https://pysteps.readthedocs.io/en/probability-forecasts/
Example: https://pysteps.readthedocs.io/en/probability-forecasts/auto_examples/probability_forecast.html#sphx-glr-auto-examples-probability-forecast-py


dnerini:
The notebook below shows how the new probabilistic nowcasting module works.

```python
import matplotlib.pyplot as plt
import numpy as np

from pysteps.nowcasts.probability_forecast import forecast

```

    Pysteps configuration file found at: /home/ned/.pysteps/pystepsrc
    


# Numerical example



```python
# parameters
precip = np.zeros((100, 100))
precip[10:50, 10:50] = 1
velocity = np.ones((2, *precip.shape))
timesteps = [0, 2, 6, 12]
nsamples = 20
thr = 0.5
slope = 1 # pixels / timestep
method = "gaussian_mv"

# compute probability forecast
out = forecast(precip, velocity, timesteps, slope, nsamples, thr, method)

# plot
for n, frame in enumerate(out):
    plt.subplot(2, 2, n + 1)
    plt.imshow(frame, interpolation="nearest", vmin=0, vmax=1)
    plt.title(f"t={timesteps[n]}")
    plt.xticks([])
    plt.yticks([])
plt.show()

plt.imshow(out[2], interpolation="nearest", vmin=0, vmax=1)
plt.title(f"t={timesteps[2]}")
plt.xticks([])
plt.yticks([])
plt.colorbar()
plt.show()
```


image

    



    
image

    


# Real-data example


```python
from datetime import datetime

from pysteps import io, rcparams, utils, nowcasts, rcparams
from pysteps.motion.lucaskanade import dense_lucaskanade
```


```python
# data source
source = rcparams.data_sources["mch"]
root = rcparams.data_sources["mch"]["root_path"]
fmt = rcparams.data_sources["mch"]["path_fmt"]
pattern = rcparams.data_sources["mch"]["fn_pattern"]
ext = rcparams.data_sources["mch"]["fn_ext"]
timestep = rcparams.data_sources["mch"]["timestep"]
importer_name = rcparams.data_sources["mch"]["importer"]
importer_kwargs = rcparams.data_sources["mch"]["importer_kwargs"]

# read precip field
date = datetime.strptime("201607112100", "%Y%m%d%H%M")
fns = io.find_by_date(date, root, fmt, pattern, ext, timestep, num_prev_files=2)
importer = io.get_method(importer_name, "importer")
precip, __, metadata = io.read_timeseries(fns, importer, **importer_kwargs)
precip, metadata = utils.to_rainrate(precip, metadata)
precip[np.isnan(precip)] = 0

# motion
motion = dense_lucaskanade(precip)

# parameters
nleadtimes = 6
nsamples = 20
thr = 1 # mm / h
slope = 1.0 * timestep # km / min
method = "gaussian_mv"

# compute probability forecast
fct = forecast(precip[-1], motion, nleadtimes, slope, nsamples, thr, method)

# plot
for n, frame in enumerate(fct):
    plt.subplot(2, 3, n + 1)
    plt.imshow(frame, interpolation="nearest", vmin=0, vmax=1)
    plt.xticks([])
    plt.yticks([])
plt.show()

plt.imshow(fct[2], interpolation="nearest", vmin=0, vmax=1)
plt.xticks([])
plt.yticks([])
plt.colorbar()
plt.show()
```


    
image

    

image


    

    



```python
from pysteps.verification import reldiag_init, reldiag_accum,plot_reldiag
```


```python
# verifying observations
importer = io.get_method(importer_name, "importer")
fns = io.find_by_date(date, root, fmt, pattern, ext, timestep, num_next_files=nleadtimes)
obs, __, metadata = io.read_timeseries(fns, importer, **importer_kwargs)
obs, metadata = utils.to_rainrate(obs, metadata)
obs[np.isnan(obs)] = 0

# reliability diagram
reldiag = reldiag_init(thr)
reldiag_accum(reldiag, fct, obs[1:])
fig, ax = plt.subplots()
plot_reldiag(reldiag, ax)
ax.set_title("Reliability diagram")
plt.show()

```

    /home/ned/PycharmProjects/pysteps/pysteps/verification/plots.py:160: MatplotlibDeprecationWarning: The 'basey' parameter of __init__() has been renamed 'base' since Matplotlib 3.3; support for the old name will be dropped two minor releases later.
      iax.set_yscale("log", basey=10)



    
image

    



```python

```

dnerini:
@loforest @pulkkins now would be a good time to get some feedback on this PR :-)

Would be nice to close this in a relatively short time, if you think that this looks reasonably ready.

dnerini:
Updated example, after switching to the convolutional approach (commit a042a4e).

```python
import matplotlib.pyplot as plt
import numpy as np

from pysteps.nowcasts.lagrangian_probability import forecast

```

    Pysteps configuration file found at: /home/ned/.pysteps/pystepsrc
    


# Numerical example



```python
# parameters
precip = np.zeros((100, 100))
precip[10:50, 10:50] = 1
velocity = np.ones((2, *precip.shape))
timesteps = [0, 2, 6, 12]
thr = 0.5
slope = 1 # pixels / timestep

# compute probability forecast
out = forecast(precip, velocity, timesteps, thr, slope=slope)

# plot
for n, frame in enumerate(out):
    plt.subplot(2, 2, n + 1)
    plt.imshow(frame, interpolation="nearest", vmin=0, vmax=1)
    plt.title(f"t={timesteps[n]}")
    plt.xticks([])
    plt.yticks([])
plt.show()

plt.imshow(out[2], interpolation="nearest", vmin=0, vmax=1)
plt.title(f"t={timesteps[2]}")
plt.xticks([])
plt.yticks([])
plt.colorbar()
plt.show()
```


image

    



    
image

    


# Real-data example


```python
from datetime import datetime

from pysteps import io, rcparams, utils, nowcasts
from pysteps.motion.lucaskanade import dense_lucaskanade
from pysteps.verification import reldiag_init, reldiag_accum,plot_reldiag
```


```python
# data source
source = rcparams.data_sources["mch"]
root = rcparams.data_sources["mch"]["root_path"]
fmt = rcparams.data_sources["mch"]["path_fmt"]
pattern = rcparams.data_sources["mch"]["fn_pattern"]
ext = rcparams.data_sources["mch"]["fn_ext"]
timestep = rcparams.data_sources["mch"]["timestep"]
importer_name = rcparams.data_sources["mch"]["importer"]
importer_kwargs = rcparams.data_sources["mch"]["importer_kwargs"]

# read precip field
date = datetime.strptime("201607112100", "%Y%m%d%H%M")
fns = io.find_by_date(date, root, fmt, pattern, ext, timestep, num_prev_files=2)
importer = io.get_method(importer_name, "importer")
precip, __, metadata = io.read_timeseries(fns, importer, **importer_kwargs)
precip, metadata = utils.to_rainrate(precip, metadata)
# precip[np.isnan(precip)] = 0

# motion
motion = dense_lucaskanade(precip)

# parameters
nleadtimes = 6
timesteps = [2, 4, 8, 16]
thr = 1 # mm / h
slope = 1 * timestep # km / min

# compute probability forecast
extrap_kwargs = dict(allow_nonfinite_values=True)
fct = forecast(precip[-1], motion, timesteps, thr, slope=slope, extrap_kwargs=extrap_kwargs)

# plot
for n, frame in enumerate(fct):
    plt.subplot(2, 3, n + 1)
    plt.imshow(frame, interpolation="nearest", vmin=0, vmax=1)
    plt.xticks([])
    plt.yticks([])
plt.show()

plt.imshow(fct[2], interpolation="nearest", vmin=0, vmax=1)
plt.xticks([])
plt.yticks([])
plt.colorbar()
plt.show()
```


    
image

    

image


    


```python
# verifying observations
importer = io.get_method(importer_name, "importer")
fns = io.find_by_date(date, root, fmt, pattern, ext, timestep, num_next_files=len(timesteps))
obs, __, metadata = io.read_timeseries(fns, importer, **importer_kwargs)
obs, metadata = utils.to_rainrate(obs, metadata)
obs[np.isnan(obs)] = 0

# reliability diagram
reldiag = reldiag_init(thr)
reldiag_accum(reldiag, fct, obs[1:])
fig, ax = plt.subplots()
plot_reldiag(reldiag, ax)
ax.set_title("Reliability diagram")
plt.show()

```

    /home/ned/PycharmProjects/pysteps/pysteps/verification/plots.py:160: MatplotlibDeprecationWarning: The 'basey' parameter of __init__() has been renamed 'base' since Matplotlib 3.3; support for the old name will be dropped two minor releases later.
      iax.set_yscale("log", basey=10)



    
image

    



```python

```

pysteps/nowcasts/interface.py
```diff
@@ -33,15 +33,19 @@
...
 from pysteps.nowcasts import anvil, sprog, steps, sseps, extrapolation
+from pysteps.nowcasts import lagrangian_probability
 
 _nowcast_methods = dict()
 _nowcast_methods["anvil"] = anvil.forecast
 _nowcast_methods["eulerian"] = eulerian_persistence
 _nowcast_methods["extrapolation"] = extrapolation.forecast
 _nowcast_methods["lagrangian"] = extrapolation.forecast
+_nowcast_methods["probability"] = lagrangian_probability.forecast
+_nowcast_methods["lagrangian_probability"] = lagrangian_probability.forecast
```
RubenImhoff, line 44: Perhaps just go for one method name here? I think I favour "lagrangian_probability".

RubenImhoff: Update: I see that we also use two names for the Lagrangian (deterministic) extrapolation forecast. Would it be better to make one name out of it, or is it safer for the pysteps users (from a continuity perspective) to not change these names?

pysteps/nowcasts/interface.py
```diff
@@ -33,15 +33,19 @@
...
 _nowcast_methods["sprog"] = sprog.forecast
 _nowcast_methods["sseps"] = sseps.forecast
 _nowcast_methods["steps"] = steps.forecast
+_nowcast_methods["steps"] = steps.forecast
```
RubenImhoff: If the review window is correct, the two lines above are the same.

pysteps/nowcasts/interface.py
```diff
@@ -65,6 +69,9 @@ def get_method(name):
     |  lagrangian or  | this approach extrapolates the last observation       |
     |  extrapolation  | using the motion field (Lagrangian persistence)       |
     +-----------------+-------------------------------------------------------+
+    |  probability    | this approach computes a local lagrangian probability |
+    |                 | forecasts.                                            |
```
RubenImhoff: Lagrangian_probability as name here?

RubenImhoff: In addition, it may be helpful for the end users to explicitly state here that the forecasts are exceedance probabilities and that the output is not a forecast rainfall amount (which is the case for the other methods).

RubenImhoff (changes requested):
Hi @dnerini,

Great work, it looks really good and its simplicity makes it a great example. I added a couple of minor comments and questions, but nothing major. 
What we may have to add more explicitly (I added a comment about that too) is that this method returns a threshold exceedance probability rather than a forecast rainfall field, which is different from the other methods in pysteps. 

Finally, I really like your examples, they make the method directly clear. I did miss the code in the reviewing process, though. I assume you're planning to add it to the examples gallery?


Cheers,

Ruben

RubenImhoff:
Oh and final comment: maybe add a y-label to the colour bars in the example (label = "Exceedance probability of threshold _x_ mm/h")

pysteps/nowcasts/interface.py
```diff
@@ -65,6 +69,9 @@ def get_method(name):
     |  lagrangian or  | this approach extrapolates the last observation       |
     |  extrapolation  | using the motion field (Lagrangian persistence)       |
     +-----------------+-------------------------------------------------------+
+    |  probability    | this approach computes a local lagrangian probability |
```
loforest:
```suggestion
    |  probability    | this approach computes local lagrangian probability |
```

pysteps/nowcasts/lagrangian_probability.py
```diff
@@ -0,0 +1,119 @@
...
+    Parameters
+    ----------
+    precip: array-like
+       Two-dimensional array of shape (m,n) containing the input precipitation
+       field.
+    velocity: array-like
+       Array of shape (2,m,n) containing the x- and y-components of the
+       advection field. The velocities are assumed to represent one time step
+       between the inputs.
+    timesteps: int or list of floats
```
loforest, line 44: Does it start at 0 or 1?

pysteps/nowcasts/lagrangian_probability.py
```diff
@@ -0,0 +1,119 @@
...
+       Number of time steps to forecast or a list of time steps for which the
+       forecasts are computed (relative to the input time step). The elements of
+       the list are required to be in ascending order.
+    threshold: float
```
loforest, line 49: Would it be possible to pass more than one threshold?
If it makes it faster it may be worth, otherwise we just call the function multiple times with different thresholds.

pysteps/nowcasts/lagrangian_probability.py
```diff
@@ -0,0 +1,119 @@
...
+    precip_forecast = extrapolation.forecast(
+        precip,
+        velocity,
+        timesteps,
+        extrap_method,
+        extrap_kwargs,
+    )
+    nanmask = np.isnan(precip_forecast)
+    precip_forecast[nanmask] = np.nanmin(precip_forecast)
+    precip_forecast = (precip_forecast > threshold).astype(float)
```
loforest:
```suggestion
    precip_forecast = (precip_forecast >= threshold).astype(float)
```
Isn't it more common to put equal or greater than threshold?

pysteps/nowcasts/lagrangian_probability.py
```diff
@@ -0,0 +1,119 @@
...
+    Returns
+    -------
+    out: ndarray_
+        Three-dimensional array of shape (num_timesteps, m, n) containing a time
+        series of nowcast exceedence probabilities. The time series starts from
+        t0 + timestep, where timestep is taken from the advection field velocity.
+
+    References
+    ----------
+    Germann, U. and I. Zawadzki, 2004:
```
loforest, line 65: Which are the implementation differences vs the method of Germann?

pysteps/nowcasts/lagrangian_probability.py
```diff
@@ -0,0 +1,119 @@
...
+        timesteps = np.arange(1, timesteps + 1)
+    precip_forecast = extrapolation.forecast(
+        precip,
+        velocity,
+        timesteps,
+        extrap_method,
+        extrap_kwargs,
+    )
+    nanmask = np.isnan(precip_forecast)
+    precip_forecast[nanmask] = np.nanmin(precip_forecast)
```
loforest: It took me a while to understand this step: we only set to 0 (or the corresponding dbZ value). Maybe add some comments here and there?
I wonder what happens if we gave a small domain full with rain everywhere.

pysteps/nowcasts/lagrangian_probability.py
```diff
@@ -0,0 +1,119 @@
...
+            kernel,
+            mode="same",
+        )
+        precip_forecast[i, ...] = signal.convolve(
+            precip_forecast[i, ...],
+            kernel,
+            mode="same",
+        )
+        precip_forecast[i, ...] /= kernel_sum
+    precip_forecast = np.clip(precip_forecast, 0, 1)
```
loforest, line 106: When could it happen that values go beyond this range?

pysteps/nowcasts/lagrangian_probability.py
```diff
@@ -0,0 +1,119 @@
...
+    nanmask = np.isnan(precip_forecast)
+    precip_forecast[nanmask] = np.nanmin(precip_forecast)
+    precip_forecast = (precip_forecast > threshold).astype(float)
+    valid_pixels = (~nanmask).astype(float)
+    for i, timestep in enumerate(timesteps):
+        scale = timestep * slope
+        if scale == 0:
+            continue
+        kernel = _get_kernel(scale)
+        kernel_sum = signal.convolve(
```
loforest: Is this done because the kernel sum is not 1?
Would it be possible normalize the kernel at the end if the  _get_kernel function?

pysteps/tests/test_nowcasts_lagrangian_probability.py
```diff
@@ -0,0 +1,57 @@
...
+from pysteps.nowcasts.lagrangian_probability import forecast
+from pysteps.tests.helpers import get_precipitation_fields
+from pysteps.motion.lucaskanade import dense_lucaskanade
+
+
+def test_numerical_example():
+    """"""
+    precip = np.zeros((20, 20))
+    precip[5:10, 5:10] = 1
+    velocity = np.zeros((2, *precip.shape))
```
loforest, line 14: I learnt something new: * operator expands it to a tuple

pysteps/tests/test_nowcasts_lagrangian_probability.py
```diff
@@ -0,0 +1,57 @@
...
+    # motion
+    motion = dense_lucaskanade(precip)
+
+    # parameters
+    timesteps = [1, 2, 3]
+    thr = 1  # mm / h
+    slope = 1 * metadata["accutime"]  # min-1
+
+    # compute probability forecast
+    extrap_kwargs = dict(allow_nonfinite_values=True)
```
loforest, line 55: What is this option for?

loforest (commented):
This is very nice contribution. Thanks! 
The spatial patterns of the convolution method look better than the "Gaussian sampling" method. Though, there is quite a difference in the reliability diagram. How do you interpret it?

pysteps/nowcasts/interface.py
dnerini: It's a fair point, we have additional names in the interfaces to reflect the various ways a method is called in the community (e.g. extrapolation or lagrangian forecast are basically the same thing). Hopefully this is not a source of confusion for the user...

pysteps/nowcasts/lagrangian_probability.py
dnerini: Good question, we don't really know the technical details of his implementation ...

pysteps/nowcasts/interface.py
dnerini: well spotted! done in 4cd86b3

pysteps/nowcasts/interface.py
dnerini: good point! done in 4fdc677

pysteps/nowcasts/interface.py
dnerini: done in 4fdc677

pysteps/nowcasts/lagrangian_probability.py
dnerini: mmmh convolution would still need to be done independently for each threshold, but you would need to perform only one extrapolation forecast. I can try to add the option!

pysteps/nowcasts/lagrangian_probability.py
dnerini: they shouldn't, but I had tests failing because of very small values above 1 (issues with numerical precision?)

pysteps/nowcasts/lagrangian_probability.py
dnerini: yes, but then you would not be able to account for missing values. This way, you count the number of valid pixels within the kernel.

pysteps/nowcasts/lagrangian_probability.py
dnerini: Excellent point! It is actually much better to assign the threshold value itself! Thanks :-)

pysteps/nowcasts/lagrangian_probability.py
dnerini: edit: not that convinced that it would bring much benefit, with the risk of cluttering the code... I'd prefer leaving this feature out for the time being.

pysteps/nowcasts/lagrangian_probability.py
dnerini: as a number of timesteps (int), it has to be > 0.
as a list of temesteps (list of floats), it can be 0. (-> analysis)

pysteps/nowcasts/lagrangian_probability.py
dnerini: mmmh not sure if there is a clear definition anywhere, but to me the word "exceed" translates more to "greater than" rather than "equal or greater than", no? 

edit: true that usually one would use P(Q>=thr) rather than P(Q>thr), I'll change this in the code (and indicate it clearly in the docs).

pysteps/tests/test_nowcasts_lagrangian_probability.py
dnerini: it allows extrapolating a field with missing values

dnerini: it allows extrapolating a field with missing values

dnerini:
Thanks @RubenImhoff  and @loforest for the precious comments and inputs!

I tried to address them at my best, please let me know if you have more concerns of suggestions.

To answer the remaining points:

from @RubenImhoff :

> 
> 
> Finally, I really like your examples, they make the method directly clear. I did miss the code in the reviewing process, though. I assume you're planning to add it to the examples gallery?

Good idea, I will!

from @loforest :
> 
> 
> This is very nice contribution. Thanks!
> The spatial patterns of the convolution method look better than the "Gaussian sampling" method. Though, there is quite a difference in the reliability diagram. How do you interpret it?

Good question. The are substantial differences between the two approaches, one being the actual number of samples used to compute the probabilities (a fix number, 20, in the first case; while it is all the pixels within the neighborhood in the second case). You think this could partially explain the more pronounced underconfidence of the Gaussian sampling method?
Also, please note that I used different leadtimes in the second case. The figure below is more comparable with the first one.

image

pysteps/nowcasts/lagrangian_probability.py
dnerini: done in 3dd9ee9

pysteps/nowcasts/lagrangian_probability.py
dnerini: see e18f0be

pysteps/nowcasts/lagrangian_probability.py
dnerini: done in caf4f8d

loforest:
Form @dnerini:
> The are substantial differences between the two approaches, one being the actual number of samples used to compute the probabilities (a fix number, 20, in the first case; while it is all the pixels within the neighborhood in the second case). You think this could partially explain the more pronounced underconfidence of the Gaussian sampling method?

It is difficult to say from a single case. It could be related to how the precipitation in the East part of the domain where there is an irregular radar mask is handled. In the Gaussian sampling example  the mask is not present anymore in the nowcast. 

I also wonder about the impact of using spatially correlated pixels to build the p.d.f. vs sampling points there are more independent.

pysteps/nowcasts/lagrangian_probability.py
loforest: We could ask him. I am mostly curious about his opinion of using all the spatially correlated pixels vs a small subset to derive the p.d.f

pysteps/nowcasts/lagrangian_probability.py
loforest: Ok good. I was thinking in terms of warnings. 
Users activate when the "psychological" threshold is reached :-)

pysteps/nowcasts/lagrangian_probability.py
dnerini: Good idea, let's ask Urs at the first occasion!

dnerini:
Thanks a lot @RubenImhoff @loforest for your reviews!

I'm merging this PR.

PR #210, opened 2021-05-30
Refactor `utils.interpolate` module

### LINKED ISSUE #209: Switch to scipy's RBF method
The `rbfinterp2d` method in `utils.interpolate` is, despite its name, a simple inverse distance weighting. To bring back some clarity, we should use [scipy's Rbf method](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.Rbf.html) instead, while renaming the current method to something more appropriate.

To avoid changing the behavior of the LK optical flow, I also suggest that the default interpolation scheme should remain the same as it is now (just with a different name). This is especially important if scipy's implementation turns out to be significantly more computationally demanding.

dnerini:
Closes #209.

RTD built: https://pysteps.readthedocs.io/en/refactor-interpolate/

### DIFF
--- a/pysteps/decorators.py
+++ b/pysteps/decorators.py
@@ -14,6 +14,7 @@ the behavior of some functions in pysteps.
     preamble_interpolation
 """
 import inspect
+import uuid
 from collections import defaultdict
 from functools import wraps
 
@@ -204,6 +205,10 @@ def preamble_interpolation(nchunks=4):
                 subxgrids = [x for x in subxgrids if x.size > 0]
                 subygrids = np.array_split(ygrid, nchunks)
                 subygrids = [y for y in subygrids if y.size > 0]
+
+                # generate a unique identifier to be used for caching
+                # intermediate results
+                kwargs["hkey"] = uuid.uuid1().int
             else:
                 subxgrids = [xgrid]
                 subygrids = [ygrid]
@@ -246,3 +251,32 @@ def preamble_interpolation(nchunks=4):
         return _interpolator_with_preamble
 
     return _preamble_interpolation
+
+
+def memoize(maxsize=10):
+    """
+    LRU cache decorator implementation for any arbitrary input since
+    caching is purely based on the optional keyword argument 'hkey'.
+    """
+
+    def _memoize(func):
+        cache = dict()
+        hkeys = []
+
+        @wraps(func)
+        def _func_with_cache(*args, **kwargs):
+            hkey = kwargs.pop("hkey", None)
+            if hkey in cache:
+                return cache[hkey]
+            result = func(*args, **kwargs)
+            if hkey is not None:
+                cache[hkey] = result
+                hkeys.append(hkey)
+                if len(hkeys) > maxsize:
+                    cache.pop(hkeys.pop(0))
+
+            return result
+
+        return _func_with_cache
+
+    return _memoize
--- a/pysteps/decorators.py
+++ b/pysteps/decorators.py
@@ -11,6 +11,7 @@ the behavior of some functions in pysteps.
 
     postprocess_import
     check_input_frames
+    preamble_interpolation
 """
 import inspect
 from collections import defaultdict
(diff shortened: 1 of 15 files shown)


pysteps/decorators.py
```diff
@@ -140,3 +143,141 @@ def new_function(*args, **kwargs):
         return new_function
 
     return _check_input_frames
+
+
+def preamble_interpolation(nchunks=4):
```
aperezhortal: What about changing the name to `prepare_interpolator` ?

pysteps/decorators.py
```diff
@@ -140,3 +143,141 @@ def new_function(*args, **kwargs):
...
+        _interpolator_with_preamble.__doc__ = inspect.cleandoc(
+            _interpolator_with_preamble.__doc__
+        )
+
+        # Add extra kwargs docstrings
+        _interpolator_with_preamble.__doc__ = (
+            _interpolator_with_preamble.__doc__.format_map(
+                defaultdict(str, extra_kwargs_doc=extra_kwargs_doc)
+            )
+        )
```
aperezhortal: This code is similar to the `postprocess_import` decorator.  I moved this snippet to a separate function (b93131f).

pysteps/io/exporters.py
```diff
@@ -510,14 +510,14 @@ def initialize_forecast_exporter_netcdf(
...
-    var_lon = ncf.createVariable("lon", np.float, dimensions=("y", "x"))
+    var_lon = ncf.createVariable("lon", float, dimensions=("y", "x"))
     var_lon[:] = lon.reshape(shape)
     var_lon.standard_name = "longitude"
     var_lon.long_name = "longitude coordinate"
     # TODO(exporters): Don't hard-code the unit.
     var_lon.units = "degrees_east"
 
-    var_lat = ncf.createVariable("lat", np.float, dimensions=("y", "x"))
+    var_lat = ncf.createVariable("lat", float, dimensions=("y", "x"))
```
aperezhortal, line 520: Why using the python `float` type instead of the numpy's types? We can maybe use `np.float32` direclty to save space.

aperezhortal (approved):
Great work @dnerini. I only left some minor coments in code review.
Besides that, I think it is ready to merge :rocket:

pysteps/io/exporters.py
dnerini: Numpy introduced the following deprecation warning:
```
DeprecationWarning: `np.float` is a deprecated alias for the builtin `float`. To silence this warning, use `float` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.float64` here.
Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
```

So the change was simply meant to silence the warning without changing the behavior of the method. It is true that float32 might be a better solution for lat lon coordinates (@pulkkins ?), but I would suggest to not include such change in this PR.

pysteps/decorators.py
dnerini: Very good idea, thanks!

pysteps/decorators.py
dnerini: I like it! I'll make the change before merging

dnerini: done in d132a9f

pysteps/utils/interpolate.py
```diff
@@ -42,186 +41,134 @@ def rbfinterp2d(
...
-
-        More details provided in the wikipedia reference page linked below.
-    epsilon: float, optional
-        The shape parameter used to scale the input to the radial kernel.
-
-        A smaller value for ``epsilon`` produces a smoother interpolation. More
-        details provided in the wikipedia reference page linked below.
-    k: int or None, optional
+    power: positive float, optional
+        The power parameter used to comptute the distance weights as
```
loforest:
```suggestion
        The power parameter used to compute the distance weights as

```

pysteps/utils/interpolate.py
```diff
@@ -8,25 +8,24 @@
...
+from scipy.interpolate import Rbf
+from scipy.spatial import cKDTree
+from scipy.spatial.distance import cdist
+
+
+from pysteps.decorators import memoize, prepare_interpolator
+
+
+@prepare_interpolator()
+def idwinterp2d(coord, input_array, xgrid, ygrid, power=0.5, k=20, **kwargs):
```
loforest:
```suggestion
def idwinterp2d(xy, values, xgrid, ygrid, power=0.5, k=20, **kwargs):

```
Some suggestion to improve variable naming.

pysteps/utils/interpolate.py
```diff
@@ -42,186 +41,134 @@ def rbfinterp2d(
...
+        # use all points
+        dist = cdist(coord, gridv, "euclidean").transpose()
+        inds = np.arange(npoints)[None, :] * np.ones((gridv.shape[0], npoints)).astype(
+            int
+        )
 
-        # create cKDTree object to represent source grid
-        tree = scipy.spatial.cKDTree(coord)
+    # compute distance-based weights
+    dist = dist / mean_res + 0.5
```
loforest: I do not understand this line. Why do you divide the distance by the mean resolution?
Should lines 74-77 be moved jyst before this line?

pysteps/utils/interpolate.py
```diff
@@ -42,186 +41,134 @@ def rbfinterp2d(
...
-        idelta = subgrid.shape[0]
 
-        if k == 0:
-            # use all points
-            d = scipy.spatial.distance.cdist(coord, subgrid, "euclidean").transpose()
-            inds = np.arange(npoints)[None, :] * np.ones(
-                (subgrid.shape[0], npoints)
-            ).astype(int)
+@prepare_interpolator()
+def rbfinterp2d(coord, input_array, xgrid, ygrid, **kwargs):
```
loforest: Same suggestion as above for the variable names

pysteps/tests/test_decorators.py
```diff
@@ -0,0 +1,33 @@
+# -*- coding: utf-8 -*-
```
loforest, line 1: This file well beyond my level :-)

loforest (commented):
I do not have major comments on this PR. It helped me have an overview of some python advanced capabilities like memoization and how to perform input arg checks thanks to decorators without "polluting" the function. Cool stuff!

pysteps/utils/interpolate.py
dnerini: The idea was to express distances in terms of pixels rather than geographical units. This way, I could simply add 0.5 (half a pixel) to represent the fact that we are interpolating on a grid rather than to points (the distance can't never be 0). But I'm less and less conviced about this.... should I just remove this part (including lines 74-77)?

pysteps/utils/interpolate.py
dnerini: done in 5bd543d

pysteps/utils/interpolate.py
dnerini: done in 5bd543d

pysteps/utils/interpolate.py
loforest: Ok. I think we can leave it in and just add a more informative comment before line 94. 
If you move lines 74-77 just before we have a clear block if code dealing with the issue. 
I like the idea of not having zero distance. It somehow adds a mini nugget effect to IDW. In case of outliers the interpolated field would be slightly less wild. Btw, is there a way to add an-priori noise term to IDW?

dnerini: Great, thanks a lot for the feedback! Done in 29b52dd

pulkkins: Should we make the constant 0.5 a user-selectable parameter?

dnerini: uhm I guess it wouldn't hurt, although I don't think this represents a very sensitive parameter. But might be anyway more transparent to have it listed as tunable parameter.

pulkkins:
Great work @dnerini. To me this looks like it's almost ready for merging (see my comment above). I made some minor fixes to docstrings.

pysteps/utils/interpolate.py
dnerini: done in a5a1a44

dnerini:
Thanks a lot @aperezhortal @loforest @pulkkins for your reviews! 

I'm merging the PR.

PR #211, opened 2021-07-09
Rapsd fix

### LINKED ISSUE #208: rapsd returns a negative frequency for even-sized images
```
In [109]: _, freqs = rapsd(np.zeros((256, 256)), return_freq=True)
In [110]: freqs[-1]
Out[110]: -0.5
```

Was this intentional? Note this doesn't happen for odd sizes:

```
In [113]: _, freqs = rapsd(np.zeros((255, 255)), return_freq=True)
In [114]: all(f >= 0 for f in freqs)
Out[114]: True
```

It looks like this code is responsible -- was there a reason for the +1 here with even sizes?

```
    if L % 2 == 0:
        r_range = np.arange(0, int(L / 2) + 1)
    else:
        r_range = np.arange(0, int(L / 2))
```

Note that plot_spectrum1d is I think ignoring this negative frequency as it'll be mapped to NaN after taking logs.

pulkkins:
This pull request introduces a fix for Issue #208 and adds tests for the `utils.spectral.rapsd` function.

### DIFF
--- a/pysteps/utils/spectral.py
+++ b/pysteps/utils/spectral.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 """
 pysteps.utils.spectral
 ======================
--- a/pysteps/utils/spectral.py
+++ b/pysteps/utils/spectral.py
@@ -122,7 +122,7 @@ def rapsd(Z, fft_method=None, return_freq=False, d=1.0, normalize=False, **fft_k
     -------
     out: ndarray
       One-dimensional array containing the RAPSD. The length of the array is
-      int(L/2)+1 (if L is even) or int(L/2) (if L is odd), where L=max(M,N).
+      int(L/2) (if L is even) or int(L/2)+1 (if L is odd), where L=max(M,N).
     freq: ndarray
       One-dimensional array containing the Fourier frequencies.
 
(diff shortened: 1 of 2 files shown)


pysteps/utils/spectral.py
```diff
@@ -131,45 +132,45 @@ def rapsd(Z, fft_method=None, return_freq=False, d=1.0, normalize=False, **fft_k
...
+    if len(field.shape) != 2:
         raise ValueError(
-            f"{len(Z.shape)} dimensions are found, but the number "
+            f"{len(field.shape)} dimensions are found, but the number "
             "of dimensions should be 2"
         )
 
-    if np.sum(np.isnan(Z)) > 0:
+    if np.sum(np.isnan(field)) > 0:
         raise ValueError("input array Z should not contain nans")
```
dnerini:
```suggestion
        raise ValueError("input field should not contain nans")
```

dnerini (approved):
hello @pulkkins , very nice contribution! 
I have a super-minor suggestion below, but otherwise looks good to go!

loforest:
Hi @pulkkins,
Thanks a lot for this!
I did not have time to review it before being merged. 
I briefly looked at it and it all looks fine.

Issue #218, opened 2021-07-26
Refactor importers


dnerini:
The current importer module includes a fairly arbitrary list of importers that are specific to data formats used by various national weather services. This list cannot grow indefinitely each time a new use case appears. 

Instead, we should leverage the recent option to use [importer plugins](https://pysteps.readthedocs.io/en/latest/developer_guide/importer_plugins.html) and only keep importers that are general enough.

I therefore suggest to
- migrate importers that are specific to a national data format as separate plugins (hosted wherever it makes more sense, including the parent pySTEPS organization account).
- keep only importers that refer to internation standards, as for example the ODIM HDF5 format (see #214).

RubenImhoff:
For the `nwp_importers` work in progress in the separate pysteps-nwp-importers repo. See: https://github.com/pySTEPS/pysteps-nwp-importers/pull/1

PR #219, opened 2021-07-26
pysteps.io with xarray

### LINKED ISSUE #12: Implement a data model
Pysteps operates on 2d grids that contain metadata (such as the grid definition, units, applied transformations and temporal resolution). Currently this information is carried in a separate metadata dictionary or as separate arguments (see e.g. pysteps.nowcasts.steps.forecast) in a very inconsistent way. It would be a good idea to store all this information into the same object.

Possible solution: use xarray (http://xarray.pydata.org/en/stable) that allows storing metadata into attributes, as well as using named columns and integration with dask.

dnerini:
Include new xarray-based data model into the pysteps.io module (See #12).

The imported data are converted into an xarray Dataset by means of a decorator.

See below for an example using MeteoSwiss data.

```
> precip_ds = pysteps.io.import_mch_gif(filename, "AQC", 5.0)
> precip_ds.info()

xarray.Dataset {
dimensions:
        y = 640 ;
        x = 710 ;

variables:
        float64 precipitation(y, x) ;
                precipitation:standard_name = precipitation_rate ;
                precipitation:long_name = Precipitation product ;
                precipitation:product = AQC ;
                precipitation:unit = mm ;
                precipitation:accutime = 5.0 ;
                precipitation:transform = None ;
                precipitation:zerovalue = 0.0 ;
                precipitation:threshold = 0.0009628129986471908 ;
                precipitation:zr_a = 316.0 ;
                precipitation:zr_b = 1.5 ;
        float64 x(x) ;
                x:standard_name = projection_x_coordinate ;
                x:units = m ;
        float64 y(y) ;
                y:standard_name = projection_y_coordinate ;
                y:units = m ;

// global attributes:
        :institution = MeteoSwiss ;
        :projection = +proj=somerc  +lon_0=7.43958333333333 +lat_0=46.9524055555556 +k_0=1 +x_0=600000 +y_0=200000 +ellps=bessel +towgs84=674.374,15.056,405.346,0,0,0,0 +units=m +no_defs ;
}
```
the same for a BOM file:
```
xarray.Dataset {
dimensions:
        y = 512 ;
        x = 512 ;

variables:
        float64 precipitation(y, x) ;
                precipitation:standard_name = precipitation_rate ;
                precipitation:long_name = Precipitation product ;
                precipitation:product = None ;
                precipitation:unit = mm ;
                precipitation:accutime = 6 ;
                precipitation:transform = None ;
                precipitation:zerovalue = 0.0 ;
                precipitation:threshold = 0.05 ;
                precipitation:zr_a = None ;
                precipitation:zr_b = None ;
        float64 x(x) ;
                x:standard_name = projection_x_coordinate ;
                x:units = m ;
        float64 y(y) ;
                y:standard_name = projection_y_coordinate ;
                y:units = m ;

// global attributes:
        :institution = Commonwealth of Australia, Bureau of Meteorology ;
        :projection = +proj=aea  +lon_0=144.752 +lat_0=-37.852 +lat_1=-18.000 +lat_2=-36.000 ;
}
```

edit: limit the scope of this PR to the io module only.

### DIFF
--- a/pysteps/decorators.py
+++ b/pysteps/decorators.py
@@ -306,8 +306,6 @@ def _to_xarray(array, metadata):
 
     data_array.attrs.update(
         {
-            "standard_name": "precipitation_rate",
-            "long_name": "Precipitation product",
             "unit": metadata["unit"],
             "accutime": metadata["accutime"],
             "transform": metadata["transform"],
--- a/pysteps/decorators.py
+++ b/pysteps/decorators.py
@@ -68,62 +68,11 @@ def postprocess_import(fillna=np.nan, dtype="double"):
         @wraps(importer)
         def _import_with_postprocessing(*args, **kwargs):
 
-            precip, quality, metadata = importer(*args, **kwargs)
+            data_array = importer(*args, **kwargs)
 
-            x1 = metadata["x1"]
-            y1 = metadata["y1"]
-            xsize = metadata["xpixelsize"]
-            ysize = metadata["ypixelsize"]
-            x = np.arange(x1 + xsize // 2, x1 + xsize * precip.shape[1], xsize)
-            y = np.arange(y1 + ysize // 2, y1 + ysize * precip.shape[0], ysize)
-
-            ds = xr.Dataset(
-                    data_vars=dict(
-                        precipitation=(("y", "x"), precip),
-                    ),
-                    coords=dict(
-                        x=(("x"), x),
-                        y=(("y"), y),
-                    ),
-                )
-            if quality is not None:
-                ds["quality"] = (("y", "x"), quality)
-
-            ds.precipitation.attrs.update(
-                {
-                    "standard_name": "precipitation_rate",
-                    "long_name": "Precipitation product",
-                    "product": metadata.get("product", None),
-                    "unit": metadata["unit"],
-                    "accutime": metadata["accutime"],
-                    "transform": metadata["transform"],
-                    "zerovalue": metadata["zerovalue"],
-                    "threshold": metadata["threshold"],
-                    "zr_a": metadata.get("zr_a", None),
-                    "zr_b": metadata.get("zr_b", None),
-                }
-            )
-
-            ds.x.attrs.update(
-                {
-                    "standard_name": "projection_x_coordinate",
-                    "units": metadata["cartesian_unit"]
-                }
-            )
-
-            ds.y.attrs.update(
-                {
-                    "standard_name": "projection_y_coordinate",
-                    "units": metadata["cartesian_unit"]
-                }
-            )
-
-            ds.attrs.update(
-                {
-                    "institution": metadata["institution"],
-                    "projection": metadata["projection"],
-                }
-            )
+            if not isinstance(data_array, xr.DataArray):
+                array, _, metadata = data_array
+                data_array = _to_xarray(array, metadata)
 
             _dtype = kwargs.get("dtype", dtype)
 
@@ -133,13 +82,15 @@ def postprocess_import(fillna=np.nan, dtype="double"):
                     "The selected precision does not correspond to a valid value."
                     "The accepted values are: " + str(accepted_precisions)
                 )
+
             _fillna = kwargs.get("fillna", fillna)
             if _fillna is not np.nan:
-                ds.precipitation.fillna(_fillna)
+                data_array = data_array.fillna(_fillna)
 
-            ds["precipitation"] = ds.precipitation.astype(_dtype)
+            data_array = data_array.astype(_dtype)
 
-            return ds
+            return data_array
+
 
         extra_kwargs_doc = """
             Other Parameters
@@ -333,3 +284,55 @@ def memoize(maxsize=10):
         return _func_with_cache
 
     return _memoize
+
+
+def _to_xarray(array, metadata):
+    """Convert to xarray DataArray."""
+    x1, x2 = metadata["x1"], metadata["x2"]
+    y1, y2 = metadata["y1"], metadata["y2"]
+    xsize, ysize = metadata["xpixelsize"], metadata["ypixelsize"]
+    # x_coords = np.arange(x1, x2, xsize) + xsize / 2
+    # y_coords = np.arange(y1, y2, ysize) + ysize / 2
+    x_coords = np.arange(x1, x1 + xsize * array.shape[1], xsize) + xsize / 2
+    y_coords = np.arange(y1, y1 + ysize * array.shape[0], ysize) + ysize / 2
+
+    data_array = xr.DataArray(
+        data=array,
+        dims=("y", "x"),
+        coords=dict(
+            x=("x", x_coords),
+            y=("y", y_coords),
+        ),
+    )
+
+    data_array.attrs.update(
+        {
+            "standard_name": "precipitation_rate",
+            "long_name": "Precipitation product",
+            "unit": metadata["unit"],
+            "accutime": metadata["accutime"],
+            "transform": metadata["transform"],
+            "zerovalue": metadata["zerovalue"],
+            "threshold": metadata["threshold"],
+            "zr_a": metadata.get("zr_a", None),
+            "zr_b": metadata.get("zr_b", None),
+            "institution": metadata["institution"],
+            "projection": metadata["projection"],
+        }
+    )
+
+    data_array.x.attrs.update(
+        {
+            "standard_name": "projection_x_coordinate",
+            "units": metadata["cartesian_unit"]
+        }
+    )
+
+    data_array.y.attrs.update(
+        {
+            "standard_name": "projection_y_coordinate",
+            "units": metadata["cartesian_unit"]
+        }
+    )
+
+    return data_array
--- a/pysteps/decorators.py
+++ b/pysteps/decorators.py
@@ -68,7 +68,62 @@ def postprocess_import(fillna=np.nan, dtype="double"):
         @wraps(importer)
         def _import_with_postprocessing(*args, **kwargs):
 
-            precip_ds = importer(*args, **kwargs)
+            precip, quality, metadata = importer(*args, **kwargs)
+
+            x1 = metadata["x1"]
+            y1 = metadata["y1"]
+            xsize = metadata["xpixelsize"]
+            ysize = metadata["ypixelsize"]
+            x = np.arange(x1 + xsize // 2, x1 + xsize * precip.shape[1], xsize)
+            y = np.arange(y1 + ysize // 2, y1 + ysize * precip.shape[0], ysize)
+
+            ds = xr.Dataset(
+                    data_vars=dict(
+                        precipitation=(("y", "x"), precip),
+                    ),
+                    coords=dict(
+                        x=(("x"), x),
+                        y=(("y"), y),
+                    ),
+                )
+            if quality is not None:
+                ds["quality"] = (("y", "x"), quality)
+
+            ds.precipitation.attrs.update(
+                {
+                    "standard_name": "precipitation_rate",
+                    "long_name": "Precipitation product",
+                    "product": metadata.get("product", None),
+                    "unit": metadata["unit"],
+                    "accutime": metadata["accutime"],
+                    "transform": metadata["transform"],
+                    "zerovalue": metadata["zerovalue"],
+                    "threshold": metadata["threshold"],
+                    "zr_a": metadata.get("zr_a", None),
+                    "zr_b": metadata.get("zr_b", None),
+                }
+            )
+
+            ds.x.attrs.update(
+                {
+                    "standard_name": "projection_x_coordinate",
+                    "units": metadata["cartesian_unit"]
+                }
+            )
+
+            ds.y.attrs.update(
+                {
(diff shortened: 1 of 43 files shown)


dnerini:
Spontaneous questions:

- should we use "precipitation" to name the main variable? Can we make it more general so that to make it clear that other variables can be used (e.g. cloud cover)?
- we should revise the metadata to better align them to CF standards and other libraries used for radar data processing (e.g. wradlib). 
  - "units" instead of "unit"
  - remove any metadata that can be easily derived from the data (e.g. threshold, zerovalue)
  - ...

cvelascof:
Spontaneous responses:

>     * should we use "precipitation" to name the main variable? Can we make it more general so that to make it clear that other variables can be used (e.g. cloud cover)?

at this point I believe that "precipitation" is a good choice, but I agree that we may have other cases such "rain_rate" 


>     * we should revise the metadata to better align them to CF standards and other libraries used for radar data processing (e.g. wradlib).

Yes ... it is better to follow CF standards as xarray use them as well  

but I would prefer to keep key metadata in the structure as parameters of the variable ... because those metadata that initially would easily derived from the data may be lost after a number of data transformations, or they could not be that easy to derive, for example if radar field is full of non zero_values then how the zero_value can be derived?

dnerini:
OK, for the moment I decided to return a DataArray instead of a Dataset (hence ignoring the quality field, which we never used anyway). This way, we remain agnostic in what variable we are reading (could be anything).

```
 <xarray.DataArray (y: 640, x: 710)>
array([[nan, nan, nan, ..., nan, nan, nan],
       [nan, nan, nan, ..., nan, nan, nan],
       [nan, nan, nan, ..., nan, nan, nan],
       ...,
       [nan, nan, nan, ..., nan, nan, nan],
       [nan, nan, nan, ..., nan, nan, nan],
       [nan, nan, nan, ..., nan, nan, nan]])
Coordinates:
  * x        (x) float64 2.555e+05 2.565e+05 2.575e+05 ... 9.635e+05 9.645e+05
  * y        (y) float64 -1.595e+05 -1.585e+05 ... 4.785e+05 4.795e+05
Attributes:
    unit:         mm
    accutime:     5.0
    transform:    None
    zerovalue:    0.0
    threshold:    0.0009628129986471908
    zr_a:         316.0
    zr_b:         1.5
    institution:  MeteoSwiss
    projection:   +proj=somerc  +lon_0=7.43958333333333 +lat_0=46.95240555555...
```

I was thinking that it would make more sense if the DataArray would include a singleton "t" dimension with the timestamp of the radar image, what you think?

cvelascof:
my personal preference is to use Dataset from source input as stores together multiple variables, like 'precipitation' and 'projection' information. The 'main/interset' variable always can be selected from the Dataset before to pass into pySTEPS routines.   

Also, "time" should be a coordinate as that is what really differentiate one radar field from other one (assuming that all of them share common metadata for all time steps)  ... also using 'time' as coordinate helps to open and concatenate multiple radar files using xarray.open_mfdataset

dnerini:
I see your point concerning the dataset, @cvelascof, and this could be easily implemented for importers. I just wonder the need for datasets when we mostly work with single variables (radar precip). One can always build a new dataset when needed (adding NWP data for example).  

And do we need an extra variable for the projection? I've seen it done before for netcdfs, but we could try to simplify this by adopting EPSG codes, at which point it would be enough to add a crs=EPSG:xxxx attribute. What do you think?

also most of our methods work on single arrays, and thus the migration on xarrays would be easier on DataArrays. Also, we shouldn't make any assumption on the name of the variables.

One possible solution could be to return a dataset when importing but then the user would have to pass the variable of interest to a given method, for example:

```
ds = pysteps.io.read_timeseries() # or xr.open_mfdataset()
precip = ds.precip.pysteps.to_rainrate()
motion = pysteps.motion.dense_lucaskanade(precip)
```

RubenImhoff:
That could work, @dnerini (thanks for all the work, by the way)! We only should not forget to change that in the examples in that case (to avoid confusion). 
@cvelascof, does projection need to be a variable? I.e., can't it just be an attribute as in @dnerini's example?

I find it hard to judge whether we need to move to DataArrays or -Sets. As @dnerini pointed out, we generally work with single arrays. What I can imagine with our move towards blending and perhaps later machine learning techniques, is that we want to involve (at some point..) more variables than just precipitation. If the information originates from one dataset, e.g. the NWP model, it may be a bit double to import that in multiple DataArrays that contain similar grid, time and metadata information. However, either way will work and I think we just have to decide on what we see as both a clearer and cleaner method. I personally have no preference.

pysteps/decorators.py
```diff
@@ -283,3 +283,53 @@ def _func_with_cache(*args, **kwargs):
...
+
+
+def _to_xarray(array, metadata):
+    """Convert to xarray DataArray."""
+    x1, x2 = metadata["x1"], metadata["x2"]
+    y1, y2 = metadata["y1"], metadata["y2"]
+    xsize, ysize = metadata["xpixelsize"], metadata["ypixelsize"]
+    # x_coords = np.arange(x1, x2, xsize) + xsize / 2
+    # y_coords = np.arange(y1, y2, ysize) + ysize / 2
+    x_coords = np.arange(x1, x1 + xsize * array.shape[1], xsize) + xsize / 2
```
RubenImhoff, line 297: This gives the same result as the commented line in line 293, right? Well, we could remove one of the options in that case. :)

In addition, just to check, this line is meant to point to the cell centers? (in that case everything is fine!)

pysteps/io/readers.py
```diff
@@ -31,50 +32,41 @@ def read_timeseries(inputfns, importer, **kwargs):
 
     Returns
     -------
-    out: tuple
-        A three-element tuple containing the read data and quality rasters and
+    out: xr.Dataset
+        A xarray Dataset containing the read data and quality rasters and
```
RubenImhoff: Seeing the current implementation, this should be DataArray, but that of course depends on our decision to move toward DataSets or -Arrays.

RubenImhoff: And small typo: "An xarray DataSet"

pysteps/tests/helpers.py
```diff
@@ -153,42 +147,31 @@ def get_precipitation_fields(
...
-        fns, importer, **_importer_kwargs
-    )
+    data_array = io.read_timeseries(fns, importer, **_importer_kwargs)
 
     if not return_raw:
 
         if (num_prev_files == 0) and (num_next_files == 0):
             # Remove time dimension
-            reference_field = np.squeeze(reference_field)
+            reference_field = np.squeeze(data_array)
```
RubenImhoff: Should this variable also be called data_array, otherwise it is not used anymore in the following lines?

pysteps/tests/helpers.py
```diff
@@ -153,42 +147,31 @@ def get_precipitation_fields(
...
 
         # Log-transform the data [dBR]
-        reference_field, ref_metadata = stp.utils.dB_transform(
-            reference_field, ref_metadata, threshold=0.1, zerovalue=-15.0
-        )
+        data_array = stp.utils.dB_transform(data_array, threshold=0.1, zerovalue=-15.0)
 
         # Set missing values with the fill value
         np.ma.set_fill_value(reference_field, -15.0)
         reference_field.data[reference_field.mask] = -15.0
```
RubenImhoff, line 183: Should reference_field be called data_array here?

pysteps/tests/helpers.py
```diff
@@ -153,42 +147,31 @@ def get_precipitation_fields(
...
 
-    if metadata:
-        return reference_field, ref_metadata
-
-    return reference_field
+    return data_array
```
RubenImhoff: Now that we don't return the metadata, do you we check somewhere whether the data_array contains metadata?

pysteps/tests/test_io_knmi_hdf5.py
```diff
@@ -10,54 +11,40 @@
...
 def test_io_import_knmi_hdf5_metadata(variable, expected, tolerance):
     """Test the importer KNMI HDF5."""
-    root_path = pysteps.rcparams.data_sources["knmi"]["root_path"]
-
-    filename = os.path.join(root_path, "2010/08", "RAD_NL25_RAP_5min_201008260000.h5")
-
-    __, __, metadata = pysteps.io.import_knmi_hdf5(filename)
-
-    smart_assert(metadata[variable], expected, tolerance)
+    smart_assert(data_array.attrs[variable], expected, tolerance)
```
RubenImhoff, line 50: Now that x, y, pixelsize, etc. is part of the DataArray and not an attribute or metadata anymore (and thus not tested as such), should the KNMI importer also have a 'geodata' that is tested here, similar to the other importers?

pysteps/tests/test_io_mrms_grib.py
```diff
@@ -14,46 +16,38 @@ def test_io_import_mrms_grib():
     """Test the importer for NSSL data."""
 
     root_path = pysteps.rcparams.data_sources["mrms"]["root_path"]
-
     filename = os.path.join(
         root_path, "2019/06/10/", "PrecipRate_00.00_20190610-000000.grib2"
     )
+    precip_full = pysteps.io.import_mrms_grib(filename, fillna=0, window_size=1)
```
RubenImhoff: Should we for consistency call this data_array too?

RubenImhoff (commented):
Great work, @dnerini! I left some comments and questions, but other than that (and if we have made up our minds about the use of DataArrays or -Sets) good to go.

pysteps/decorators.py
cvelascof: perhaps we should stay with DataArray for the moment for the core rutines. Said that however, xarray opens netCDF as Datasets so eventually user may need to select the variable from a dataset to run our core routines.

cvelascof: About 'projection', CF conventtions stat that "A grid mapping variable may be referenced by a data variable in order to explicitly declare the coordinate reference system (CRS) used for the horizontal spatial coordinate values"
https://cfconventions.org/Data/cf-conventions/cf-conventions-1.8/cf-conventions.html#grid-mappings-and-projections
so if outputs are to be netCDF, it would be useful to have Datasets with both the variable and the 'grid mapping variable' (aka projection)

dnerini: > 
> 
> This gives the same result as the commented line in line 293, right? Well, we could remove one of the options in that case. :)
> 
It should, but at the moment the commented lines do not work because some importers wrongly report the x1, x2, y1, y2 coordinates as the coordinates of the corner _pixels_ instead of the corners themselves... but we should use the commented code (it's safer), which is why I would prefer to leave them there for the time being.

> In addition, just to check, this line is meant to point to the cell centers? (in that case everything is fine!)
Exactly!

dnerini: > 
> 
> perhaps we should stay with DataArray for the moment for the core rutines. Said that however, xarray opens netCDF as Datasets so eventually user may need to select the variable from a dataset to run our core routines.

OK, let's stick to DataArrays for now, but we can easily switch to dataset at any moment before realising V2.

dnerini: > 
> 
> About 'projection', CF conventtions stat that "A grid mapping variable may be referenced by a data variable in order to explicitly declare the coordinate reference system (CRS) used for the horizontal spatial coordinate values"
> https://cfconventions.org/Data/cf-conventions/cf-conventions-1.8/cf-conventions.html#grid-mappings-and-projections
> so if outputs are to be netCDF, it would be useful to have Datasets with both the variable and the 'grid mapping variable' (aka projection)

Certainly very important for the netcdf exporters, perhaps less internally to the code itself?

pysteps/tests/helpers.py
dnerini: absolutely, which is done when testing the importers

pysteps/tests/test_io_knmi_hdf5.py
dnerini: See #218: the idea for V2 is to remove the national importers from the core library (pysteps) and refactor them as optional plugins. In this sense, the KNMI will become an external package (e.g. "pysteps-knmi-importer") that will be maintained (and tested) externally (by you I suppose ;-) )

dnerini:
This PR is by no means definitive, but should be seen instead as a first step to integrate the xarray data model into pysteps. I suggest we merge it pretty soon and advance in other modules, knowing that we will anyway have to come back to the io to make more adjustments before V2 will be ready.

RubenImhoff (approved):

aperezhortal:
Great work @dnerini! I agree that we should merge it soon and advance in other modules.

I pushed two commits that add a temporary "legacy" option to allow the importers and readers to behave as in version 1. Also, the tests that are still written for v1, were updated to use this new legacy option. The tests that couldn't be easily fix were manually skipped and tagged with a TODO message. This legacy option will allow maintaining the testing functionality through the migration to xarray.

dnerini:
Ah very nice @aperezhortal , thanks! Looks like tests are now failing because of missing dependencies. I'll cherry-pick commit b4c165ade1f31fb7567457a6bd95686702d8f70a to fix it as soon as possible (should have some time this afternoon).

Issue #216, opened 2021-07-23
pysteps 2.0


ladc:
During a meeting with @dnerini, @loforest and @wdewettin, the suggestion was made to start working towards a new version of pysteps. 

1. The initial motivation is the implementation of the **blending with NWP** (#212) which is likely to require some API-breaking changes.
2. Secondly, a move towards a new data model based on **xarray** would simplify the existing code as well as the future implementation of the blending. This was suggested before, see #12 and [this gist](https://gist.github.com/kmuehlbauer/645e42a53b30752230c08c20a9c964f9) by @kmuehlbauer. This move might also simplify #215 and #5 since the self-describing structure of xarrays is close to the netCDF format.
3. Implementation of @dnerini's Kalman Filter would be nice to have.
4. The interface should be generic enough to allow for machine learning nowcasting and post-processing methods, which require multiple predictors. Maybe OO or duck typing can be useful here. The interface to perform the nowcasts could have a generic fit/predict/update structure.
5. Logging can also be improved: #140
6. Achieve 90% test coverage (#32)

dnerini:
At this point, it would be interesting to hear the opinion of the other core contributors @pulkkins @aperezhortal @cvelascof @RubenImhoff and in general the whole pysteps community.

Also, the list above is by no means exhaustive nor fixed. Version 2 will be the opportunity to introduce substantial changes based on our experience so far: we need your inputs! For this purpose, I created a new [project](https://github.com/orgs/pySTEPS/projects/1) where to list and track issues and PRs related to this effort.

In terms of roadmap, I'd suggest the following:

## Phase 1: up to end of October 2021(*):
- master contains pysteps-v1
- version 1.5 is released, after which no more PRs into v1 except for bug fixes (only patches are released)
- setup new development branch for pysteps-v2 (alpha), features branches are forked and merged into this v2 branch

**Goal:** pysteps-v2 is published as a pre-release (beta)

## Phase 2: up to September 2022 (ERAD):

- previous master with v1 is renamed to pysteps-v1
- v1 will release patches only
- development branch pysteps-v2 is renamed master and becomes default branch
- submit abstract to ERAD2022

**Goal**: consolidate v2, which becomes default version.

## Phase 3: 
- pysteps v2 is stable
- pysteps v1 is discontinued (no more support, no more releases of any kind)
- paper on v2 
- next focus: machine/deep learning methods

**Goal**: close project, publish results.

[...]

PR #221, opened 2021-07-27
Ide nowcasting

pulkkins:
This branch implements the Lagrangian INtegro-Difference equation model with Autoregression (LINDA) developed by Pulkkinen et al. (2021).

Remaining issues that will be addressed in the next version:
- The timesteps argument is required to be an integer. Supplying the nowcast timesteps in a list is not yet supported.
- There are several functions that overlap with existing methods implemented in pysteps. For instance, _estimate_ar*_params, _iterate_ar_model and _window_tukey can be replaced with existing methods or moved outside linda.py.


pysteps/nowcasts/linda.py
```diff
@@ -0,0 +1,1359 @@
...
+
+warnings.filterwarnings("ignore")
+
+
+def forecast(
+    precip_fields,
+    advection_field,
+    timesteps,
+    feature_method="blob",
+    feature_kwargs={},
```
dnerini: Using empty dictionaries as default argument is [dangerous](https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python) and should be avoided. Using None is the recommended practice for such cases (same as with `vel_pert_kwargs`).

pysteps/nowcasts/linda.py
```diff
@@ -0,0 +1,1359 @@
...
+            vel_pert_kwargs = dict()
+
+    print("Computing LINDA nowcast")
+    print("-----------------------")
+    print("")
+
+    print("Inputs")
+    print("------")
+    print(
+        "dimensions:           {}x{}".format(
```
dnerini: can we use f-strings instead? e.g.

`"dimensions:    {}x{}".format(precip_fields.shape[1], precip_fields.shape[2])`

becomes

`f"dimensions:    {precip_fields.shape[1]}x{precip_fields.shape[2]}"`

pysteps/nowcasts/linda.py
```diff
@@ -0,0 +1,1359 @@
...
+            print(
+                "velocity perturbations, perpendicular: {:.2f}, {:.2f}, {:.2f}".format(
+                    vp_perp[0], vp_perp[1], vp_perp[2]
+                )
+            )
+            vel_pert_kwargs = vel_pert_kwargs.copy()
+            vel_pert_kwargs["vp_par"] = vp_par
+            vel_pert_kwargs["vp_perp"] = vp_perp
+
+    fct_gen = _linda_init(
```
dnerini: nice! this goes already into the OOP direction we are discussing in #216 
@ladc

pysteps/nowcasts/linda.py
```diff
@@ -0,0 +1,1359 @@
...
+    else:
+        print("done.")
+
+    if measure_time:
+        return fct_gen, precip_fields_lagr_diff, time.time() - starttime_init
+    else:
+        return fct_gen, precip_fields_lagr_diff
+
+
+# compute "masked" convolution where non-finite values are ignored
```
dnerini: can you make this comment the docstrings of the method? (same below)

pysteps/nowcasts/linda.py
```diff
@@ -0,0 +1,1359 @@
...
+    bb_y1, bb_x1, bb_y2, bb_x2 = _compute_ellipse_bbox(phi, sigma1, sigma2, cutoff)
+
+    x = np.arange(int(bb_x1), int(bb_x2) + 1).astype(float)
+    if len(x) % 2 == 0:
+        x = np.arange(int(bb_x1) - 1, int(bb_x2) + 1).astype(float)
+    y = np.arange(int(bb_y1), int(bb_y2) + 1).astype(float)
+    if len(y) % 2 == 0:
+        y = np.arange(int(bb_y1) - 1, int(bb_y2) + 1).astype(float)
+
+    X, Y = np.meshgrid(x, y)
```
dnerini: `X` and `Y` do not conform to the snake_case convention for variables, could use `x_grid` and `y_grid` instead.
(same applies to `XY`).

pysteps/nowcasts/linda.py
```diff
@@ -0,0 +1,1359 @@
...
+        add_perturbations,
+        vel_pert_method,
+        vel_pert_kwargs,
+        kmperpixel,
+        timestep,
+        num_workers,
+        measure_time,
+    )
+    if measure_time:
+        fct_gen, precip_fields_lagr_diff, init_time = fct_gen
```
dnerini, line 294: `init_time` is not used...

pysteps/nowcasts/linda.py
```diff
@@ -0,0 +1,1359 @@
...
+                False,
+            )
+
+        precip_fct_ensemble = []
+
+        rs = np.random.RandomState(seed)
+
+        if DASK_IMPORTED and num_workers > 1:
+            res = []
+            for i in range(num_ens_members):
```
dnerini:
```suggestion
            for _ in range(num_ens_members):
```

pysteps/nowcasts/linda.py
```diff
@@ -0,0 +1,1359 @@
...
+        for i in range(len(estim_weights)):
+            res.append(dask.delayed(worker)(i))
+
+        psi = dask.compute(*res, num_workers=num_workers, scheduler="threads")
+    else:
+        psi = []
+        for i in range(len(estim_weights)):
+            psi.append(worker(i))
+
+    return [np.sum([psi[i] * interp_weights[i] for i in range(len(psi))], axis=0)]
```
dnerini:
```suggestion
    return [np.sum([psi_ * interp_weights[i] for i, psi_ in enumerate(psi)], axis=0)]
```

pysteps/nowcasts/linda.py
```diff
@@ -0,0 +1,1359 @@
...
+    field_src,
+    field_dst,
+    weights,
+    mask,
+    kernel_type="anisotropic",
+    kernel_params={},
+    num_workers=1,
+):
+    masks = []
+    for i in range(len(weights)):
```
dnerini: simplify syntax
```
for weight in weights:
    masks.append(np.logical_and(mask, weight > 1e-3))
```

pysteps/nowcasts/linda.py
```diff
@@ -0,0 +1,1359 @@
...
+
+    return pert_gen
+
+
+# iterate autoregressive process
+# TODO: use the method implemented in pysteps.timeseries.autoregression
+def _iterate_ar_model(input_fields, psi):
+    input_field_new = 0.0
+
+    for i in range(len(psi)):
```
dnerini: use enumerate ?

pysteps/nowcasts/linda.py
```diff
@@ -0,0 +1,1359 @@
...
+
+    precip_fct = precip_fields[-1].copy()
+    precip_fields_lagr_diff = precip_fields_lagr_diff.copy()
+
+    displacement = None
+    extrap_kwargs["return_displacement"] = True
+    precip_fct_out = []
+
+    for i in range(precip_fields_lagr_diff.shape[0]):
+        for j in range(ari_order - i):
```
dnerini: is it right here that index `j` is not used?

pysteps/nowcasts/linda.py
```diff
@@ -0,0 +1,1359 @@
...
+    acf_window_radius=None,
+    extrap_method="semilagrangian",
+    extrap_kwargs={},
+    add_perturbations=True,
+    pert_thrs=(0.5, 1.0),
+    num_ens_members=40,
+    vel_pert_method="bps",
+    vel_pert_kwargs=None,
+    kmperpixel=None,
+    timestep=None,
```
dnerini, line 80: The default parameters will raise an ValueError since both timestep and kmperpixel are None while vel_pert_method="bps"

dnerini (changes requested):
Seppo congratulations on this impressive piece of work! 
I look forward to see this new method in action, for now I left few comments and suggestions, together with a handful of minor commits (for tests). 

Would it be also possible to add an example script as already done with ANVIL?

pysteps/nowcasts/linda.py
```diff
@@ -0,0 +1,1353 @@
...
+    timesteps,
+    feature_method="blob",
+    feature_kwargs=None,
+    ari_order=1,
+    kernel_type="anisotropic",
+    localization_window_radius=None,
+    errdist_window_radius=None,
+    acf_window_radius=None,
+    extrap_method="semilagrangian",
+    extrap_kwargs={},
```
dnerini: same issue with default empty dictionary.

pysteps/nowcasts/linda.py
pulkkins: I thought that it's not needed because these are private methods that are not visible outside the module. Is that OK?

pysteps/nowcasts/linda.py
pulkkins: True, but the convolution needs to be applied multiple times depending on i and the ARI order.

pysteps/nowcasts/linda.py
pulkkins: The default setting is now to disable velocity perturbations. Do we also have the same problem with steps.forecast?

pysteps/nowcasts/linda.py
dnerini: It's just that they look as docstrings to me and would make the code more readable as such imho

dnerini:
I noticed some duplication in the code, for example you reimplemented the Tukey tapering window even though we already have it available in the utils.tapering module. Probably not something urgent to be done now, but I would strongly recommend to try to reduce code duplication by using methods already available in pysteps. If for some reasons the existing methods do not cover your needs, we should first try to extend or modify them instead of creating new ones (especially now that V2 #216 will give us more freedom for breaking changes). Otherwise, the risk is to get to a point of having a code base that is impossible to maintain...

What do you think, @pulkkins ? Is there something we can do now or is it better to close this PR and keep this in mind for a small refactoring in V2?

dnerini:
Concerning performance now, I run some basic profiling using Swiss radar data (640x710 domain) and default arguments to produce one hour nowcast (12 time steps). 

### Default parameters

```
Computing LINDA nowcast
-----------------------

Inputs
------
dimensions:           640x710
number of time steps: 3

Methods
-------
nowcast type:         ensemble
feature detector:     blob
extrapolator:         semilagrangian
kernel type:          anisotropic

Parameters
----------
number of time steps:       12
ARI model order:            1
localization window radius: 128.0
error dist. window radius:  96.0
error ACF window radius:    160.0
ensemble size:              40
parallel workers:           1
seed:                       24

Detecting features... found 138 blobs in 1.10 seconds.
Transforming to Lagrangian coordinates... 0.41 seconds.
Estimating the first convolution kernel... 406.42 seconds.
Estimating the ARI(p,1) parameters... 3.27 seconds.
Estimating the second convolution kernel... 304.47 seconds.
Estimating forecast errors... 15.12 seconds.
Estimating perturbation parameters... 561.82 seconds.
```
image
Note that the computation of the 40 ensemble members for 12 time steps took more than 1 hour! Peak memory at about 2.5 GB during the model fit.

Hence
- the default feature detection parameters produce a too high number of features, which blows up the computation time. 
- the default number of members (40) is too high for a sequential approach.

### Higher detection threshold, smaller ensemble size

Increasing the detection threshold to 10 mm/h reduces the number of blobs to 11, speeding up considerably the forecast.
Also, the default number of members can be easily reduced to 10. For larger ensemble sizes, we should advise on using Dask by specifying a number of workers > 1.
The detection parameter is not necessarily the most robust way to control the number of features. Is there any better way of doing it?

```
[...]

Parameters
----------
number of time steps:       12
ARI model order:            1
localization window radius: 128.0
error dist. window radius:  96.0
error ACF window radius:    160.0
ensemble size:              10
parallel workers:           1
seed:                       24

Detecting features... found 11 blobs in 1.10 seconds.
Transforming to Lagrangian coordinates... 0.43 seconds.
Estimating the first convolution kernel... 39.61 seconds.
Estimating the ARI(p,1) parameters... 0.34 seconds.
Estimating the second convolution kernel... 33.70 seconds.
Estimating forecast errors... 1.74 seconds.
Estimating perturbation parameters... 54.22 seconds.
```
image

Total time is now down to about 6 minutes, with ca 1.6 seconds / time step / member, and a 1.2 GB peak memory.

### Parallelization with Dask

Using 10 workers (for parallel threads) get total time further down to 80 seconds, although at the cost of a larger memory footpring (4 GB):
```
[...]
Parameters
----------
number of time steps:       12
ARI model order:            1
localization window radius: 128.0
error dist. window radius:  96.0
error ACF window radius:    160.0
ensemble size:              10
parallel workers:           10
seed:                       24

Detecting features... found 11 blobs in 0.98 seconds.
Transforming to Lagrangian coordinates... 0.27 seconds.
Estimating the first convolution kernel... 7.49 seconds.
Estimating the ARI(p,1) parameters... 0.13 seconds.
Estimating the second convolution kernel... 6.70 seconds.
Estimating forecast errors... 1.57 seconds.
Estimating perturbation parameters... 36.29 seconds.
```
image

pysteps/nowcasts/linda.py
```diff
@@ -0,0 +1,1407 @@
...
+
+    dist_params = []
+    stds = []
+    acf_fft_ampl = []
+
+    if DASK_IMPORTED and num_workers > 1:
+        res = []
+        for i in range(feature_coords.shape[0]):
+            res.append(dask.delayed(worker)(i))
+        res = dask.compute(*res, num_workers=num_workers, scheduler="multiprocessing")
```
dnerini: can you explain why you selected "multiprocessing" here instead of "threads" as elsewhere? 
For some reasons, this is giving me a runtime error:
```
RuntimeError:
        An attempt has been made to start a new process before the
        current process has finished its bootstrapping phase.
```

pulkkins: The reason is that "threads" gives absolutely no parallelization here. This is probably because of the GIL. I agree that "multiprocessing" is not the best solution, but so far I haven't figured out a better way.

pulkkins: Thanks @dnerini for your effort so far. You have done quite extensive work to go through this pull request.

pulkkins:
> I noticed some duplication in the code, for example you reimplemented the Tukey tapering window even though we already have it available in the utils.tapering module. Probably not something urgent to be done now, but I would strongly recommend to try to reduce code duplication by using methods already available in pysteps. If for some reasons the existing methods do not cover your needs, we should first try to extend or modify them instead of creating new ones (especially now that V2 #216 will give us more freedom for breaking changes). Otherwise, the risk is to get to a point of having a code base that is impossible to maintain...
> 
> What do you think, @pulkkins ? Is there something we can do now or is it better to close this PR and keep this in mind for a small refactoring in V2?

Let's do that in V2. The purpose of this pull request is to implement the first usable version of LINDA. It can be refined later.

pysteps/nowcasts/linda.py
dnerini: You're right about the GIL, although I did experience a significant speedup with "threads" (not sure why and how).

But my main question here is why the runtime error? is it only me or there is a bug somewhere we need to fix before releasing it? would be safer to set it to "threads" instead?

pulkkins: I think I found the problem. For some reason, all programs that call a module containing `dask.compute(...,scheduler="multiprocessing")` need to do this inside the `if __name__ == "__main__":` block. This is not very convenient for the user. So let's use "threads" instead.

pysteps/nowcasts/linda.py
```diff
@@ -0,0 +1,1410 @@
...
+    ari_order=1,
+    kernel_type="anisotropic",
+    localization_window_radius=None,
+    errdist_window_radius=None,
+    acf_window_radius=None,
+    extrap_method="semilagrangian",
+    extrap_kwargs=None,
+    add_perturbations=True,
+    pert_thrs=(0.5, 1.0),
+    num_ens_members=40,
```
dnerini: Ok, one last comment ;-)
Can we reduce the default number of ensemble members to something much lower such as 10? 40 members is too many for default settings (see the profiling results above).

dnerini:
With the changes introduced in #225, were you thinking of adding a default `max_num_features` ? Something around 20?

pulkkins:
> With the changes introduced in #225, were you thinking of adding a default `max_num_features` ? Something around 20?

Done.

dnerini:
Very nice, well done @pulkkins ! If you agree, id like to do one last profiling with the default parameters and, if everything looks good, I think we can merge.

Edit: here it is for the same case as above (640x710 domain, 12 time steps).

| | Sequential | Parallel |
| :--- | ------------- | ----- |
| Parameters | 25 blobs, 10 members, 1 worker | 25 blobs, 10 members, 10 workers |
| | image | image |
| Total time (forecast loop) |  11 min (ca. 4 seconds / member / time step) | 4 min (ca. 0.5 seconds / member / time step) |
| Peak memory | 1.2 GB  | 4.2 GB |

dnerini (approved):

dnerini:
The RTD built for this branch can be found here: 
https://pysteps.readthedocs.io/en/ide_nowcasting/

PR #223, opened 2021-07-28
pysteps.utils with xarray

### LINKED ISSUE #198: `dB_transform` ignores existing metadata
Original report
----------------
(by Valentina Gregori)

> I also noticed another small issue (already in the previous version of pysteps). When I do a conversion to reflectivity using the function conversion.to_reflectivity(R, metadata) the output value of metadata['zerovalue'] is not as expected. I think this happens because in the transformation.dB_transform function, which is called by conversion.to_reflectivity, the code that checks if "zerovalue" is a key of metadata is missing:
> ```
> if zerovalue is None:
>     zerovalue = metadata.get("zerovalue", threshold-5)
> ```
> for now, it seems that the "zerovalue" is only replaced by "threshold-5". Do you think it makes sense?
> Thanks in advance!

dnerini:
Include new xarray-based data model into the `pysteps.utils` module (See #12).

## `utils.conversion`
- remove `metadata` arguments (since all attributes are now part of the input xr.DataArray object)
- make all methods available as [xarray accessors](http://xarray.pydata.org/en/stable/internals/extending-xarray.html) for DataArrays (e.g. can use both the `data_array.pysteps.to_rainrate()` or `pysteps.utils.to_rainrate(data_array)` syntax)
- `to_reflectivity()` has the option to convert to dB [dBZ] or keep linear scale [Z]
- add type hints for inputs and outputs
- reduce code duplication
- adapt tests

## `utils.dimension` (DEPRECATED)
- deprecate `aggregate_fields_time` in favor of xarray's `coarsen` method.
- deprecate `aggregate_fields_space` in favor of xarray's `coarsen` method.
- deprecate `aggregate_fields` in favor of xarray's `coarsen` method.
- deprecate `clip_domain` in favor of xarray's `sel` method.
- deprecate `square_domain` since it is obsolete.

## `utils.images`
- `morph_opening` takes xr.DataArray as input, and it also handles cases with time dimension "t"
- argument `n` of `morph_opening` is renamed `kernel` and can either be the size of the kernel or the kernel itself.
- update tests

## `utils.interpolate`
- all interpolation methods take and return xarray's object (both Dataset and DataArray)
- migrate deprecated scipy's `Rbf` to new `RBFInterpolator` (scipy>=1.7)
- update tests

## `utils.transformation`
- remove `metadata` arguments (since all attributes are now part of the input xr.DataArray object)
- make all methods available as [xarray accessors](http://xarray.pydata.org/en/stable/internals/extending-xarray.html) for DataArrays (e.g. can use both the `data_array.pysteps.sqrt_transform()` or `pysteps.utils.sqrt_transform(data_array)` syntax)
- add general `back_transform` method to remove any existing data transformation.
- `boxcox_transform` and `db_transform` use a simpler offset to avoid computing log(0), instead of having to define the `threshold` and `zerovalue` parameters as in V1 (closes #198)
- refactor `nq_transform`: in particular, the inverse transformation needs an empirical target distribution (the `template` argument).
- add type hints for inputs and outputs
- reduce code duplication
- rewrite tests


dnerini:
OK let's say this is ready for review!

- still failing tests (need to adapt all methods that use the utils module)
- didn't touch at the `utils.spectral` module for now, do we want to adopt xarray for complex data too? this would be similar to what is done in [xrft](https://xrft.readthedocs.io/en/latest/). Proposition: should we adopt xrft ? @pulkkins 

I know it is a lot of changes, I don't expect you to go through all of them. Rather, please try to see whether you agree with the general changes in the interfaces listed in the summary at the top.

Thanks!

RubenImhoff:
This is an amazing amount of work, @dnerini! I'll try to have a look at it soon.

examples/anvil_nowcast.py
```diff
@@ -1,4 +1,5 @@
 # coding: utf-8
+# coding: utf-8
```
RubenImhoff: Very minor: I believe we have: 
`# -*- coding: utf-8 -*-`
in most of the scripts

examples/anvil_nowcast.py
```diff
@@ -91,25 +90,18 @@
...
 
-# log-transform the data and the threshold value to dBR units for S-PROG
-rainrate_field_db, _ = transformation.dB_transform(
-    rainrate_field, metadata, threshold=0.1, zerovalue=-15.0
-)
-rainrate_thr, _ = transformation.dB_transform(
-    np.array([0.5]), metadata, threshold=0.1, zerovalue=-15.0
-)
+# log-transform the data to dBR units for S-PROG
+rainrate_field_db = rainrate_field.pysteps.db_transform()
```
RubenImhoff, line 90: No need to define the threshold and zerovalue, or are the same values present in the "metadata" of the DataArray?

RubenImhoff: (it will work anyway, just checking that the numbers are the same)

pysteps/io/exporters.py
```diff
@@ -497,14 +500,14 @@ def initialize_forecast_exporter_netcdf(
...
-    var_xc.units = metadata["cartesian_unit"]
+    var_xc.units = metadata.get("cartesian_unit", "m")
 
     var_yc = ncf.createVariable("y", np.float32, dimensions=("y",))
     var_yc[:] = yr
     var_yc.axis = "Y"
     var_yc.standard_name = "projection_y_coordinate"
     var_yc.long_name = "y-coordinate in Cartesian system"
-    var_yc.units = metadata["cartesian_unit"]
+    var_yc.units = metadata.get("cartesian_unit")
```
RubenImhoff: Similar to var_xc?
```suggestion
    var_yc.units = metadata.get("cartesian_unit", "m")
```

pysteps/io/nowcast_importers.py
```diff
@@ -106,6 +106,8 @@ def import_netcdf_pysteps(filename, onerror="warn", **kwargs):
         Precipitation field in mm/h. The dimensions are [latitude, longitude].
         The first grid point (0,0) corresponds to the upper left corner of the
         domain, while (last i, last j) denote the lower right corner.
+    quality: None
+        Not implement.
```
RubenImhoff, line 110: What will we do with the quality field? It is useful, but so far, it is not used to my knowledge.

pysteps/tests/test_io_mrms_grib.py
```diff
@@ -33,10 +33,7 @@ def test_io_import_mrms_grib():
         "transform": None,
         "zerovalue": 0,
         "threshold": 0.1,
-        "x1": -129.99999999999997,
-        "x2": -60.00000199999991,
-        "y1": 20.000001,
-        "y2": 55.00000000000001,
+        "bounding_box": (-129.99999999999997, -60.00000199999991, 20.000001, 55.00000000000001),
```
RubenImhoff, line 36: I like the implementation of the bounding_box!

pysteps/utils/conversion.py
```diff
@@ -12,308 +12,153 @@
...
+        DataArray to be converted.
+
+    Other parameters
+    ----------------
+    accutime: float
+        The accumulation time in minutes, used if converting from rain depth.
+        If not passed as an argument, the accutime specified in the
+        attributes of the input xr.DataArray is used.
+
+    zr_a, zr_b: float
```
RubenImhoff, line 38: Is this something we still want to set in the attributes of the DataArray?

pysteps/utils/interface.py
```diff
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+# -*- coding: utf-8 -*-
```
RubenImhoff: Seems like:
`# -*- coding: utf-8 -*-`
is twice in there.

RubenImhoff (approved):
Amazing work, @dnerini! Because you've done so much work, I can't promise that I've spotted all possible mistakes in the code, so let's also wait for another review by one of the others. ;)

examples/anvil_nowcast.py
dnerini: The idea is to replace them with a simpler offset, see for example [here](https://github.com/pySTEPS/pysteps/blob/c7a9e26f829334c2eb657be2591fd4100ef09044/pysteps/utils/transformation.py#L142).

This is not yet enforced in all methods, especially in the examples, which will need to be updated along as we refactor the pysteps methods.

PR #228, opened 2021-08-09
Bom import xarray

### LINKED ISSUE #215: Implement importer for NWP (netcdf) data
Implement function to read in CF-compliant NetCDF data, not (yet) including interpolation to the radar / nowcasting grid. This might be simplified by moving to xarray first. Required for #212 .

cvelascof:
This PR reads directly BoM netCDF files using xarray and returns a DataArray.
Also it implements some tests for the new importer based on xarray.

### DIFF
--- a/pysteps/io/__init__.py
+++ b/pysteps/io/__init__.py
@@ -8,3 +8,4 @@ from .exporters import *
 from .importers import *
 from .nowcast_importers import *
 from .readers import *
+from .nwp_importers import *
--- a/pysteps/io/importers.py
+++ b/pysteps/io/importers.py
@@ -559,9 +559,9 @@ def _import_bom_rf3_geodata_xr(ds_in,
 
     if factor_scale != 1.0:
         ds_in['x'] = ds_in.x*1000.
-        ds_in.x.attrs.update({'units': 'm'})        
+        ds_in.x.attrs.update({'units': 'm'})
         ds_in['y'] = ds_in.y*1000.
-        ds_in.y.attrs.update({'units': 'm'})        
+        ds_in.y.attrs.update({'units': 'm'})
 
     cartesian_unit = ds_in.x.units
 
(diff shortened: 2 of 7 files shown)


cvelascof:
This PR uses xarray to load the BoM netcdf files and adds the pySTEPS metadata as attributes to the DataArrays (Data and coordinates).
New routines were created and named as the old ones plus the postfix "_xr". Old routines are still in the repository for legacy purposes. All tests related with the new routines are passing.

Test failing are related with io_import_mrms_grib routine.
@dnerini . @RubenImhoff  can you give it a look and try to fix them?

RubenImhoff:
The failing MRMS tests seems like something @aperezhortal knows.

cvelascof:
hi @ladc ... good point ... _import_bom_rf3_geodata_xr already have varname as parameter but it would be nice to connect it with kwargs passed to the parent function import_bom_rf3_xr. I will do that soon.

On the other hand, importers.py is in theory only for radar inputs ... so probably we may need a new nwp_importers.py to collect the specific routines to read and preprocess nwp datasets. For example, deal with lat/lon rather than x and y, or transform accumulated precipitation from runtime that it is typical for NWP variable to accumulated precipitation for each time step that it is what actually will be blended with radar data.
 What do you think?

cvelascof:
it seems chunks are not working in python3.6 for some reason but they work great using python3.8... so at this point I will not pass chunks information to xarray.
Something to think about later on.

cvelascof:
@ladc  this adds an importer for some of the sample nwp data in pysteps-data. We may need to add this nwp reader to other branches. This is just the foundation for additional NWP data readers.

ladc:
@cvelascof Great! I don't know how realistic it is to make a generic NWP importer function that can read CF-compliant netcdf files but this looks like a good start.

PR #229, opened 2021-08-09
Linear blending

### LINKED ISSUE #226: Implement linear blending
Implement a linear blending algorithm which can be applied on the different nowcast algorithms.

wdewettin:
This addresses issue #226. I added a nowcasting method called linear blending. This works for all nowcast methods (deterministic and ensemble) and is also compatible with NWP ensembles. Furthermore, I added an example which uses some dummy NWP data to blend with and which plots the results. Lastly, I also added a test for the linear blending function.


RubenImhoff:
Great work, @wdewettin! I was (actually, we were - that is including @wdewettin and @ladc) wondering whether we should place this method in a/the blending folder or whether it fits in the nowcasts folder as currently done by @wdewettin. Any thoughts?

pysteps/nowcasts/linear_blending.py
```diff
@@ -0,0 +1,172 @@
...
+      The time difference (in minutes) between consecutive forecast fields.
+    nowcast_method: {'anvil', 'eulerian', 'lagrangian', 'extrapolation', 'lagrangian_probability', 'sprog', 'steps', 'sseps'}
+      Name of the nowcast method used to forecast the precipitation.
+    R_nwp: array_like or NoneType, optional
+      Array of shape (timesteps, m, n) in the case of no ensemble or
+      of shape (n_ens_members, timesteps, m, n) in the case of an ensemble
+      containing the NWP precipitation fields ordered by timestamp from oldest
+      to newest. The time steps between the inputs are assumed to be regular
+      (and identical to the time step between the nowcasts). If no NWP
+      data is given the value of R_nwp is None and no blending will be performed.
```
RubenImhoff: In case of R_nwp = None, the script is similar to steps.py, right? Do we want to use the script as a postprocessing tool, i.e. the user makes a nowcast with steps.py and passes the output on to linear_blending.py, or do we want to keep it this way and have some duplicate code with steps.py? As steps.py may also change to allow for other blending methods, I wonder what everyone thinks about the best way forward.

pysteps/nowcasts/linear_blending.py
```diff
@@ -0,0 +1,172 @@
...
+    nowcast_method_func = nowcasts.get_method(nowcast_method)
+    R_nowcast = nowcast_method_func(
+        precip,
+        velocity,
+        timesteps,
+        **nowcast_kwargs,
+    )
+
+    # Transform the precipitation back to mm/h
+    R_nowcast = transformation.dB_transform(R_nowcast, threshold=-10.0, inverse=True)[0]
```
RubenImhoff: I believe the threshold is present in the metadata (the transformer at least returns the metadata). We could add this to the function, then we don't have to hard code it here.

pysteps/nowcasts/linear_blending.py
```diff
@@ -0,0 +1,172 @@
...
+
+        if n_ens_members_min != n_ens_members_max:
+            if n_ens_members_nwp == 1:
+                R_nwp = np.repeat(R_nwp[np.newaxis, :, :], n_ens_members_max, axis=0)
+            elif n_ens_members_nowcast == 1:
+                R_nowcast = np.repeat(
+                    R_nowcast[np.newaxis, :, :], n_ens_members_max, axis=0
+                )
+            else:
+                repeats = [
```
RubenImhoff: After consulting @wdewettin and @ladc, I figured out what this piece of code does. So if you e.g. have 10 ensemble nowcasts members and 3 NWP members, the output will be an ensemble of 10 members. Hence, the three NWP members are blended with the first three members of the nowcast (member one with member one, two with two, etc.), subsequently, the same NWP members are blended with the next three members (NWP member one with member 4, NWP member 2 with member 5, etc.), until 10 is reached.

Is this the way to go (works for me) or do we want an output of 30 (10 x 3 members here) in this case?

pysteps/nowcasts/linear_blending.py
```diff
@@ -0,0 +1,172 @@
...
+      start_blending and end_blending the nowcasts and NWP data are linearly
+      merged with each other. After end_blending only the NWP data is used.
+    nowcast_kwargs: dict, optional
+      Dictionary containing keyword arguments for the nowcast method.
+
+    Returns
+    -------
+    R_blended: ndarray
+      Array of shape (timesteps, m, n) in the case of no ensemble or
+      of shape (n_ens_members, timesteps, m, n) in the case of an ensemble
```
RubenImhoff: With the way it is implemented at this moment: n_ens_members equals the maximum number of ensemble members in the input, so either the number of the nowcast or that of the NWP. Perhaps good to add that here.

pysteps/nowcasts/linear_blending.py
RubenImhoff: Oh and @wdewettin, let's add some comments to the code here so that this procedure is clear. :) (B.t.w., I can add that too, but I'll wait for the responses - and thus possible changes - of the others)

pysteps/nowcasts/linear_blending.py
```diff
@@ -0,0 +1,172 @@
...
+                )
+            else:
+                R_blended[:, i, :, :] = (
+                    weight_nwp * R_nwp[:, i, :, :]
+                    + weight_nowcast * R_nowcast[:, i, :, :]
+                )
+
+            # Find where the NaN values are and replace them with NWP data
+            nan_indices = np.isnan(R_blended)
+            R_blended[nan_indices] = R_nwp[nan_indices]
```
RubenImhoff: Is it possible that the NWP data also has NaNs? If so, we can also fill those by the nowcast data.

pysteps/nowcasts/linear_blending.py
RubenImhoff: (B.t.w., I can add that too, but I'll wait for the responses - and thus possible changes - of the others)

examples/plot_linear_blending.py
```diff
@@ -0,0 +1,162 @@
...
+
+# Import the dummy NWP data (vertical line moving to the left)
+R_nwp = dummy_nwp(R[-1, :, :], n_leadtimes + 1, max=7, mean=4, speed=0.2 * 350)
+
+# Log-transform the data to unit of dBR, set the threshold to 0.1 mm/h,
+# set the fill value to -15 dBR
+R, metadata = transformation.dB_transform(R, metadata, threshold=0.1, zerovalue=-15.0)
+
+# Set missing values with the fill value
+# R[~np.isfinite(R)] = -15.0
```
RubenImhoff: If not used, we could remove these lines.

examples/plot_linear_blending.py
```diff
@@ -0,0 +1,162 @@
...
+R, _, metadata = io.read_timeseries(fns, importer, legacy=True, **importer_kwargs)
+
+# Convert to rain rate
+R, metadata = conversion.to_rainrate(R, metadata)
+
+# Upscale data to 2 km to limit memory usage
+R, metadata = dimension.aggregate_fields_space(R, metadata, 2000)
+
+# Import the dummy NWP data (vertical line moving to the left)
+R_nwp = dummy_nwp(R[-1, :, :], n_leadtimes + 1, max=7, mean=4, speed=0.2 * 350)
```
RubenImhoff: Perhaps plot the radar rainfall and NWP rainfall forecast field somewhere, to give the reader (as this example may end up in the pysteps gallery) and idea of what the example fields look like.

RubenImhoff (commented):
Great, @wdewettin! I left some comments and questions. Some of them I could implement myself as well. Let's just wait for the comments of the others (and feel free to comment yourself on them), then we'll continue with the final changes. 
Exciting to have this option in pysteps soon!

pysteps/nowcasts/linear_blending.py
wdewettin: Yes! I will await the responses from the others, and then add some comments

examples/plot_linear_blending.py
wdewettin: True, in the plot_steps_nowcast.py example, this is how the missing data was handled, but here we use the (dummy) NWP data

examples/plot_linear_blending.py
wdewettin: Do you mean plotting them for every timestep? Or only the first one(s)?

RubenImhoff: First ones should be fine. :)

dnerini:
> 
> 
> Great work, @wdewettin! I was (actually, we were - that is including @wdewettin and @ladc) wondering whether we should place this method in a/the blending folder or whether it fits in the nowcasts folder as currently done by @wdewettin. Any thoughts?

Because of the way linear blending works, I personally think it could be provided as a simple postprocessing method (therefore within the postprocessing module) that takes as inputs two separate nowcasts and the relatives weights. 

This said, I like the way @wdewettin implemented it, where one has to specify the the nowcasting method and pass the corresponding NWP forecast. But because of this dependency on the actual nowcasting methods, it could make more sense to move into a new blending module. We could say that as a general (and not necessarily very strict) rule, the methods included in a module should be independent one to the other. In this sense, the nowcating module should only include alternative nowcasting methods.

dnerini:
Very nice contribution @wdewettin ! I highly appreciate the fact that you already included all the documentation (including a gallery example!) and tests, well done!

I plan to go through your PR in detail as soon as possible. For now, I have a more general request. Would it be possible to make your method work with an xarray DataArray object instead of numpy arrays?

wdewettin:
Yes, @dnerini, I will take a look at making my method work with xarrays!

wdewettin:
@dnerini Sorry for my late response! I looked into making it work with xarrays instead of numpy arrays and I found that it would be better to wait with that for the time being. All the functions on which the linear blending function is based still output numpy arrays, and it will be a lot easier to make the linear blending function work with xarrays if those functions are changed first.

pysteps/nowcasts/linear_blending.py
ladc: I think this will not occur. Either the NWP data is present or it's not, NaNs are not likely to occur.

pysteps/nowcasts/linear_blending.py
```diff
@@ -0,0 +1,177 @@
+# -*- coding: utf-8 -*-
+"""
+pysteps.nowcasts.linear_blending
+================================
+
+Implementation of the linear blending between nowcast and NWP data.
```
cvelascof: @RubenImhoff @wdewettin thanks for implementing this module.

pysteps/nowcasts/linear_blending.py
cvelascof: Here the object returned should be equal to the output of the nowcasting method used if R_nwp=None. 
I understand that if we are going to create a function the ingest both sources of precipitation, then few lines of code must be duplicated. Perhaps in the future a higher level blending script should be created in the way that accept multiple sources of precipitation, and weights for each source (or a method to determine those weigths) and return blended fields. This approach may be easier to maintain, but perhaps harder to understand by users.

pysteps/nowcasts/linear_blending.py
cvelascof: I understand why this is done here, but this is assuming that `R_nwp` is in mm/hr and that the `precip` is in dB. If this is the case, then we need to indicate those limitations in the descrition of the function.  Alternatively, both sources should be in `mm/hr` and the conversion of `precip` to `dB` be done within this function as well.

pysteps/nowcasts/linear_blending.py
cvelascof: Great job @wdewettin and @RubenImhoff

pysteps/nowcasts/linear_blending.py
cvelascof: I agree. It wuld be nice to add additional comments in this part of the script as @RubenImhoff suggested.

cvelascof: I agree. It would be nice to add additional comments in this part of the script as @RubenImhoff suggested.

pysteps/nowcasts/linear_blending.py
cvelascof: I understand why this is done here, but this is assuming that `R_nwp` is in mm/hr and that the `precip` is in dB. If this is the case, then we need to indicate those limitations in the description of the function.  Alternatively, both sources could be in `mm/hr` and the conversion of `precip` to `dB` be done within this function as well. 

Said that the returned output `R_blended` will be always in `mm/hr`

cvelascof:
@RubenImhoff @wdewettin @dnerini apologies for the mess in my comments ... not sure what did it go wrong?

RubenImhoff:
All requested changes implemented. Main task left is to make the code xarray ready for pysteps-v2. As discussed before, we may also do that in a separate PR.

Issue #212, opened 2021-07-13
Implement STEPS blending scheme


cvelascof:
A great addition to pySTEPS should be to include a module and support routines to blend multiple sources of forecast rainfall.
For example, blending a radar rainfall nowcast with a Numerical Weather Prediction (NWP) rainfall forecast or multiple sources of rainfall estimates.

The branch steps_blending (https://github.com/pySTEPS/pysteps/tree/steps_blending) is created to develop a version of STEPS blending scheme as proposed by Seed et al, 2013 (https://agupubs.onlinelibrary.wiley.com/doi/full/10.1002/wrcr.20536)

Proposed modules and routines:

Modules:
- [ ] blend_multiple_power_spectrum: weighted combination of multiple power spectrum
- [x] blend_multiple_optical_flow_fields: weighted combination of multiple optical flow fields
- [x] calculate_weights: Calculate blending weights for STEPS blending
- [x] blend_cascades: weighted combination of cascade levels based on STEPS weights
- [x] blend_means_sigmas: weighted combination of input means and standard deviations using STEPS weights
- [x] recompose_cascade: recompose rainfall field from blended cascade fields
- [x] main module: to drive and integrate multiple sources of rainfall

Support:
- [x] Docs
- [x] Tests for all modules
- [x] Examples using notebooks

Data - case studies:
- [x] radar rainfall datasets
- [x] NWP rainfall datasets

This is just a first trial to describe what would be needed to complete this feature.
Early versions of some of those modules are already in the branch but they need more work and development.

dnerini:
Hello @cvelascof, very happy to see this important branch back on track, I look forward to the new blending module!

dnerini:
Do I get it right that all methods will assume that NWP and radar data come at the same spatio-temporal resolution as well as extent? 

Also, do you aim at blending with an ensemble or a deterministic NWP?

cvelascof:
It is easier to assume to start that NWP and radar data come at the same spatial projection and at the same spatial-temporal resolution. First version should combine one deterministic NWP with the radar nowcast. The critical aspect now is to define an structure in the module that allows us to add more complications to this first version later on, for example more than one NWP, or an ensemble NWP.

RubenImhoff:
The idea is indeed to start simple, but to later allow for different spatio-temporal resolutions, extents and ensemble members.

dnerini:
Very good, I like the approach!

[...]

PR #231, opened 2021-08-17
Functions to store and compute climatological weights

ladc:
This implements the functions save_weights and calc_clim_weights which save the current and compute the climatological weights (a running mean of the weights of the past n days, where typically n=30. These climatological weights are used in the scale- and skill-dependent blending of STEPS with NWP forecasts, see #212,  which will be part of pysteps v2.0, see #216.
The current weights, based on the correlations per cascade level, are regressed towards these climatological weights in the course of the forecast.


pysteps/blending/clim.py
```diff
@@ -0,0 +1,176 @@
...
+    Add the current NWP weights to update today's daily average weight. If the
+    day is over, update the list of daily average weights covering a rolling
+    window.
+
+    Parameters
+    ----------
+    current_weights: array-like
+      Array of shape [model, scale_level, ...]
+      containing the current weights of the different NWP models per cascade level.
+    outdir_path: string
```
RubenImhoff: validtime still misses here. What kind of date string should it be?

RubenImhoff (commented):
Nice work, @ladc!! 
I have one question left and I made some changes to make all the tests working. :)

pysteps/blending/clim.py
ladc: Thanks for the feedback! I looked into using numpy's built-in datetime64 type but got discouraged by the apparent lack of documentation. So I'll stick to datetime instead.

ladc (commented):
I guess this works now!

PR #234, opened 2021-08-24
Add importer for RMI NWP data

ladc:
Add importer for the RMI's NWP data. 
Add tests (for ALARO/AROME at native 1.3km grid for now)
Tests will not work without [PR4 in pysteps-data](https://github.com/pySTEPS/pysteps-data/pull/4)

### DIFF
--- a/pysteps/io/importers.py
+++ b/pysteps/io/importers.py
@@ -483,41 +483,51 @@ def import_bom_rf3_xr(filename, **kwargs):
             "but it is not installed"
         )
 
-    ds = _import_bom_rf3_data_xr(filename, **kwargs,)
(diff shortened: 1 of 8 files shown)


ladc:
The importers seem quite similar so probably creating a generic CF-compliant netCDF importer instead is feasible.

RubenImhoff (approved):
Nice work, @ladc! I'll add a KNMI nwp importer later too. 
In a while, I think we can move all three nwp importers to one general CF-compliant importer, as the importers are rather similar so far.

ladc:
Thanks Ruben! Some of the tests are still failing though, and I'm not sure why. It's only for the 3.8 version and I suppose it's unrelated to my changes.

PR #235, opened 2021-08-25
Add KNMI Harmonie NWP importer and tests

RubenImhoff:
Following earlier commits by @cvelascof  and @ladc, I'm also adding a KNMI nwp importer. At a later stage, we should try to make a more universal importer for all nwp products.


ladc (commented):
Looks good! And indeed all very similar, so maybe to be merged into a single CF-compliant importer  :-)

PR #236, opened 2021-09-16
Reprojection functionality

### LINKED ISSUE #230: Allow for reprojection and up/downscaling of DataArrays
With our move to pysteps v2 and our plan to implement blending with NWP, we will run into problems with different grids and spatial projections of the models. An example of this is the KNMI NWP model data, which have a lat-lon grid and a coarser resolution than the radar data. Hence, this asks for a reprojection and downscaling of the NWP rainfall fields. 

Does anyone know about good Python-tools for reprojection and down/upscaling? I'd be happy to try to implement this next week or so.

RubenImhoff:
Thanks to @wdewettin for most of the work! This PR resolves issue #230. 

Reprojection functionality for changing the projection and grid (up/downscaling possible) of input fields - generally NWP data for blending - with a destination field (the original radar field for instance). This functionality was not yet present in pysteps, but becomes necessary for blending two different data sources. Note that this implementation uses a new module, rasterio, which we may have to make an optional dependency in pysteps. It is worth reconsidering this, but at least the script is working and stable now with both the Belgian and Dutch data. 

- [x] Reprojection script
- [x] Changes to KNMI radar and NWP importer to allow for correct reprojection
- [x] Changes to NWP importer and/or decorators to make sure the grid is specified for the cell centers, see issue #215.
- [x] Testing script
- [x] Documentation


pysteps/utils/reprojection.py
```diff
@@ -0,0 +1,87 @@
+# -*- coding: utf-8 -*-
+
+from rasterio import Affine as A
+from rasterio.warp import reproject, Resampling
+import xarray as xr
+import numpy as np
+
+
+def reprojection(R_src, R_dst):
```
dnerini: please use variable names following the standard snake_case convention, e.g.
```suggestion
def reprojection(src_array, dst_array):
```

pysteps/utils/reprojection.py
```diff
@@ -0,0 +1,87 @@
...
+    """Reprojects precipitation fields to the domain of another precipiation
+    field.
+
+    Parameters
+    ----------
+    R_src: xarray
```
dnerini: here and elsewhere, specify whether you expect an xarray.Dataset or xarray.DataArray

```suggestion
    R_src: xr.DataArray
```

pysteps/utils/reprojection.py
```diff
@@ -0,0 +1,87 @@
+# -*- coding: utf-8 -*-
+
+from rasterio import Affine as A
+from rasterio.warp import reproject, Resampling
+import xarray as xr
+import numpy as np
+
+
+def reprojection(R_src, R_dst):
+    """Reprojects precipitation fields to the domain of another precipiation
```
dnerini: there are few typos in the docstrings
```suggestion
    """Reprojects precipitation fields to the domain of another precipitation
```

pysteps/utils/reprojection.py
```diff
@@ -0,0 +1,87 @@
...
+
+    # Initialise the reprojected (x)array
+    R_rprj = np.zeros((R_src.shape[0], R_dst.shape[-2], R_dst.shape[-1]))
+
+    # For every timestep, reproject the precipitation field of R_src to
+    # the domain of R_dst
+    if R_src.attrs["yorigin"] != R_dst.attrs["yorigin"]:
+        R_src = R_src[:, ::-1, :]
+
+    for i in range(R_src.shape[0]):
```
dnerini: with xarray, we can directly refer to the name of the dimension, without having to assume that the time is the first one:
```suggestion
    for array_slice in R_src.transpose("t", ...):
```
we transpose the dimension "t" to the first position so to iterate along it.

dnerini (commented):
Very nice work to all: this is an important utility functions when blending data from different sources!

I left few minor suggestions directly into the code. 

rasterio: I don't mind the additional optional dependency (please remember to update the list of optional dependencies in the documentation), but since in V2 we anyway work with xarray, I would suggest to use [rioxarray](https://corteva.github.io/rioxarray/stable/) instead. This should simplify our code, most notably the need to loop through time, and the definition of the output DataArray at the very end. In fact, rioxarray implements a [reproject_match](https://corteva.github.io/rioxarray/stable/rioxarray.html?highlight=reproject#rioxarray.raster_array.RasterArray.reproject_match) method which looks very similar to what we are doing here.

RubenImhoff:
Thanks for your useful feedback, @dnerini! I didn't have time to work on it this week, but the changes will follow asap. :)

pysteps/utils/reprojection.py
RubenImhoff: Changed it, for now, to 
`for i in range(src_array["t"].shape[0]):`,
because we also need an index for R_rprj. If there is a more elegant way to do this, let me know (and possibly later with rioxarray, the for loop is not needed anymore)!

RubenImhoff:
@cvelascof and @ladc, see also the description in issue #215, I have made some changes to the nwp importers to set the cell coordinates to the center of the cells. Could you check if this is also the center for the rmi and bom importers?

RubenImhoff:
So far, I've sticked to rasterio instead of rioxarray. When trying rio, I kept running into projection issues. Maybe it is just me, or rio is not as stable yet as rasterio. Feel free to try it out and adjust the scripts, if you know how to do it properly! 

I do agree with @dnerini that it would be nice to eventually move to rioxarray.

RubenImhoff:
Tests still fail due to rasterio, I haven't figured out yet why..

cvelascof:
the sample NWP datasets used for BoM are already regrided to the same projection of the BoM radar, so coordinates are at the center of the pixel

RubenImhoff:
Except for the use of rioxarray instead of rasterio, I think this PR is getting ready for a final review. We may implement rioxarray later, unless someone has enough experience with it to make it compatible for a variety of spatial projections in our radar and nwp products. Only the tests still fail, due to the import of the rasterio module. Any idea what causes this?

dnerini:
> Only the tests still fail, due to the import of the rasterio module. Any idea what causes this?

mmmh you sure it is linked to the import? I see a RuntimeError when you load the values from the netcdf at line 87 in reprojection.py. My guess is that this creates too many new threads (because of the loop?). Things to try: load the data before the loop, try to use xarray's [apply_ufunc](http://xarray.pydata.org/en/stable/generated/xarray.apply_ufunc.html) instead.

environment.yml
```diff
@@ -14,3 +14,4 @@ dependencies:
   - pyproj
   - scipy
   - xarray
+  - rasterio
```
dnerini: we tend to list only main dependencies here

examples/test_steps_blending_rmi.py
```diff
@@ -0,0 +1,287 @@
...
+import xarray as xr
+import os
+from datetime import datetime
+from pprint import pprint
+from pysteps import io, rcparams
+from pysteps.utils import conversion, transformation, reprojection
+from pysteps.visualization import plot_precip_field
+
+# Set workdir and number of output time steps
+work_dir = "c:\\Users\\imhof_rn\\pysteps"
```
dnerini: remove absolute local path

examples/test_steps_blending_rmi.py
```diff
@@ -0,0 +1,287 @@
...
+importer_name = data_source["importer"]
+importer_kwargs = data_source["importer_kwargs"]
+
+# Find the radar files in the archive
+fns = io.find_by_date(
+    date, root_path, path_fmt, fn_pattern, fn_ext, timestep, num_prev_files=2
+)
+
+importer = io.get_method(importer_name, "importer")
+R_input = io.read_timeseries(fns, importer, legacy=False, **importer_kwargs)
```
dnerini: to import test data, you should use the helper function [get_precipitation_fields](https://github.com/pySTEPS/pysteps/blob/229dd9394b22423ec60741edc6f2ca10fcac274b/pysteps/tests/helpers.py#L25)
example [here](https://github.com/pySTEPS/pysteps/blob/229dd9394b22423ec60741edc6f2ca10fcac274b/pysteps/tests/test_noise_fftgenerators.py#L7).

examples/test_steps_blending_rmi.py
```diff
@@ -0,0 +1,287 @@
...
+# PLOTTING
+
+# R, _, metadata = io.import_odim_hdf5(filename, "RATE", legacy=True, **importer_kwargs)
+# Convert to rain rate
+# R, metadata = conversion.to_rainrate(R, metadata)
+# Nicely print the metadata
+# pprint(metadata)
+# Plot the rainfall field
+# plot_precip_field(R, geodata=metadata)
+# plt.show()
```
dnerini: if not strictly necessary, I'd rather remove all commented code blocks

examples/test_steps_blending_rmi.py
```diff
@@ -0,0 +1,287 @@
...
+nwp_rep_metadata.update(**R_NWP.y.attrs)
+nwp_rep_metadata.update(**R_NWP.attrs)
+
+# plot_precip_field(R_NWP[0,:,:], geodata=nwp_rep_metadata)
+# plt.show()
+
+########## Now start the blending! ################
+
+import pysteps
+from pysteps import cascade, blending
```
dnerini: please all imports at the top

examples/test_steps_blending_rmi.py
```diff
@@ -0,0 +1,287 @@
...
+
+R_input.data[R_input.data < radar_metadata["threshold"]] = 0.0
```
dnerini: usual annoying comment: let's try to follow python's naming convention more strictly than in the past: var names use snake_case

pysteps/utils/reprojection.py
```diff
@@ -0,0 +1,113 @@
...
+import numpy as np
+
+try:
+    import rasterio
+
+    RASTERIO_IMPORTED = True
+except ImportError:
+    RASTERIO_IMPORTED = False
+from rasterio import Affine as A
+from rasterio.warp import reproject, Resampling
```
dnerini: these two import will throw an error if rasterio is not installed, you can include them in the try ... except block, too

pysteps/utils/reprojection.py
```diff
@@ -0,0 +1,113 @@
...
+    x1_dst = dst_array.x.attrs["x1"]
+    y2_dst = dst_array.y.attrs["y2"]
+    xpixelsize_dst = dst_array.attrs["xpixelsize"]
+    ypixelsize_dst = dst_array.attrs["ypixelsize"]
+    dst_transform = A.translation(float(x1_dst), float(y2_dst)) * A.scale(
+        float(xpixelsize_dst), float(-ypixelsize_dst)
+    )
+
+    # Initialise the reprojected (x)array
+    R_rprj = np.zeros((src_array.shape[0], dst_array.shape[-2], dst_array.shape[-1]))
```
dnerini: snake_case ;-)

pysteps/utils/reprojection.py
```diff
@@ -0,0 +1,113 @@
...
+
+    # For every timestep, reproject the precipitation field of src_array to
+    # the domain of dst_array
+    if src_array.attrs["yorigin"] != dst_array.attrs["yorigin"]:
+        src_array = src_array[:, ::-1, :]
+
+    for i in range(src_array["t"].shape[0]):
+        reproject(
+            src_array.values[i, :, :],
```
dnerini:
```suggestion
            src_array.isel(t=i),
```

pysteps/utils/reprojection.py
```diff
@@ -0,0 +1,113 @@
...
+
+    # Initialise the reprojected (x)array
+    R_rprj = np.zeros((src_array.shape[0], dst_array.shape[-2], dst_array.shape[-1]))
+
+    # For every timestep, reproject the precipitation field of src_array to
+    # the domain of dst_array
+    if src_array.attrs["yorigin"] != dst_array.attrs["yorigin"]:
+        src_array = src_array[:, ::-1, :]
+
+    for i in range(src_array["t"].shape[0]):
```
dnerini, line 84: as already mentioned above, we could use xarray's apply_ufunc as an alternative

requirements.txt
```diff
@@ -9,3 +9,4 @@ jsonschema
 netCDF4
 xarray
 bottleneck
+rasterio
```
dnerini: we tend to list only main dependencies here

setup.py
```diff
@@ -68,6 +68,7 @@
     "jsonschema",
     "xarray",
     "bottleneck",
+    "rasterio",
```
dnerini: not a mandatory dependency, I suggest to remove it from here

dnerini (changes requested):

dnerini:
> So far, I've sticked to rasterio instead of rioxarray. When trying rio, I kept running into projection issues. Maybe it is just me, or rio is not as stable yet as rasterio. 

Thanks for trying it out! I'm really fine with sticking to rasterio if it simplifies things to us. As you say, we can alwas decide to migrate to rioxarray later on.

pysteps/utils/reprojection.py
dnerini: yep I see. You can use the .isel method to get the corresponding fields from src_array (I left a comment below).

examples/test_steps_blending_rmi.py
RubenImhoff: Oops, I see that this example script (that will later on, in a different version be used in the main_blending_module PR) ended up in this PR. I guess I was working on two PRs at the same time. I'll remove it, so that will directly solve the listed issues here. ;)

RubenImhoff:
Thanks a lot for looking into this, @dnerini! I'll make the changes coming Monday. :)

RubenImhoff:
Hmm, I keep getting "pysteps.exceptions.MissingOptionalDependency: rasterio package is required for the reprojection tool, but it isnot installed". What am I doing wrong in the development dependencies? It is listed as an optional dependency.

dnerini:
maybe there is a problem with the installation of rasterio in the test environment? Try to remove the try... except block to expose directly the original error message

RubenImhoff:
Good call, that changes it in "ModuleNotFoundError: No module named 'rasterio'" unfortunately. I think I've missed an optional dependency or so for the tester, but can't find it :(

dnerini:
Sorry I didn't notice that rasterio is missing from the test dependencies , look at the top of .github/workflows/test_pysteps.yaml

RubenImhoff:
Timing, just found it myself too. It had to be somewhere. ;) Thanks for the help!

RubenImhoff:
Tests succeed now! Only the tests for pysteps 3.6 fail due to rasterio looking for the gdal_version, which gives an error in only pysteps 3.6 for windows. Any ideas on how to fix this? 
Other than that, the PR is probably good to go. A recommended change for later is to move from rasterio to rioxarray.

pysteps/utils/reprojection.py
dnerini: as reference, you can look at the new version of the utils.images.morph_opening() method:
https://github.com/pySTEPS/pysteps/blob/b67f8c0d0659f8b81f954c6fb01ccdd8b59657b7/pysteps/utils/images.py#L71

dnerini:
> 
> 
> Tests succeed now! Only the tests for pysteps 3.6 fail due to rasterio looking for the gdal_version, which gives an error in only pysteps 3.6 for windows. Any ideas on how to fix this? Other than that, the PR is probably good to go. A recommended change for later is to move from rasterio to rioxarray.

GDAL can be a real pain to install, but I thought conda would help in that sense. Maybe you can check that the right version of GDAL is installed? 

Besides the error, your tests in `test_utils_reprojection.py` cannot pass if rasterio is not installed, which can happen as it is an optional dependency.
please add `pytest.importorskip("rasterio")` at the top of the test, and do the same for any other optional dependency in there.

PR #248, opened 2021-12-29
Verification sal

EsmailGhaemi:
The Structure-Amplitude-Location (SAL) method has been added. Note that the original approach suggested by (Wernli et al., 2009) is based on a simple minimum threshold (f), but here we used tstorm function to detect objects (Feldmann et al., 2021). Using this function allows to have a more realistic way to detect precipitation as objects.


pysteps/verification/spatialscores.py
```diff
@@ -667,6 +667,377 @@ def fss_compute(fss):
     return 1.0 - numer / denom
 
 
+import numpy as np
+import pandas as pd
+from math import sin, cos, sqrt, atan2, radians, hypot
+from skimage.measure import regionprops_table
+from pysteps.feature import tstorm as tstorm_detect
+import warnings
```
aperezhortal: I suggest moving these imports to the top of the file.  

Also, `pandas` and `scikit-image` are optional dependencies in pysteps.  Our current approach to deal with optional dependencies is using `try-catch` blocks during the imports, setting a variable (i.e. `XXX_IMPORTED`) to True if the import was successful, or False otherwise. Then, the function using these dependencies should check the value XXX_IMPORTED variable and raise a `MissingOptionalDependency` exception if the dependency is missing.
See for example the [tstorm](https://github.com/pySTEPS/pysteps/blob/master/pysteps/feature/tstorm.py).

pysteps/verification/spatialscores.py
```diff
@@ -667,6 +667,377 @@ def fss_compute(fss):
...
+
+warnings.filterwarnings(action="ignore")
```
aperezhortal: What kind of warnings need to be ignored? Is it necessary suppressing them at the module level? 

If ignoring warnings is necessary, consider [Temporarily Suppressing Warnings](https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings) using context managers.

pysteps/verification/spatialscores.py
```diff
@@ -667,6 +667,377 @@ def fss_compute(fss):
...
+
+
+def SAL(X_f, X_o,
```
aperezhortal: [The recommended code style](https://pysteps.readthedocs.io/en/latest/developer_guide/contributors_guidelines.html#code-style) in pysteps uses `lower_with_under` style for the function name, and `CapWords` for classes. 

I suggest renaming the function to `sal()`.

pysteps/verification/spatialscores.py
```diff
@@ -667,6 +667,377 @@ def fss_compute(fss):
...
+    df_obs: 2-d ndarray for the observation data.
+
+    df_forc: 2-d ndarray for the prediction data.
+
+    maximum_distance: maximum distance of the study area.
+    If the projection is rectangular (e.g., UTM), this value is the diagonal of the study area.
+    If the projection is not rectangular (e.g., lon/lat), 'max_dist' function calculates this value.
+
+    minref: minimum precipitation value for detecting object(s), If r star is lower than this threshold.
+    The default is 0.1 mm.
```
aperezhortal: Small adjustments to follow the numpydoc style. Also, consider adding the units of the `maximum_distance` and the `minref` values.

```suggestion
    Parameters
    ----------
    df_obs: 2-d ndarray
        Observation data.
    df_forc: 2-d ndarray
        Prediction data.
    maximum_distance: float
        Maximum distance of the study area in XXX units.
        If the projection is rectangular (e.g., UTM), this value is the diagonal of the study area.
        If the projection is not rectangular (e.g., lon/lat), 'max_dist' function calculates this value.
    minref: float
        Minimum precipitation value for detecting object(s), If r star is lower than this threshold.
        The default is 0.1 mm.
```

pysteps/verification/spatialscores.py
```diff
@@ -667,6 +667,377 @@ def fss_compute(fss):
...
+        l = np.nan
+    dic = {"S": s, "A": a, "L": l}
+    sal = pd.DataFrame(dic, index=[1])
+    sal.index.name = "step"
+    return sal
+
+
+def detect_objects_tstorm(df, minref, maxref, mindiff, minsize, minmax, mindis):
+    """This function detects thunderstorms using a multi-threshold approach (Feldmann et al., 2021).
+    Parameters
```
aperezhortal:
```suggestion
    Parameters
    ----------
```

pysteps/verification/spatialscores.py
```diff
@@ -667,6 +667,377 @@ def fss_compute(fss):
...
+    df: array-like
+    Array of shape (m,n) containing input data. Nan values are ignored.
+
+    minref: float, optional
+    Lower threshold for object detection. Lower values will be set to NaN. The default is 0.1.
```
aperezhortal: To compress the docstrings I suggest removing the spaces between parameters docstings.

```suggestion
    df: array-like
        Array of shape (m,n) containing input data. Nan values are ignored.
    minref: float, optional
        Lower threshold for object detection. Lower values will be set to NaN. The default is 0.1.
```
Also, note that four spaces are prepended to the parameter description.

pysteps/verification/spatialscores.py
```diff
@@ -667,6 +667,377 @@ def fss_compute(fss):
...
+    Minimal area for possible detected object. The default is 4 pixels.
+
+    minmax: float, optional
+    Minimum value of maximum in identified objects. Objects with a maximum lower than this will be discarded. The default is 0.1.
+
+    mindis: float, optional
+    Minimum distance between two maxima of identified objects. Objects with a smaller distance will be merged. The default is 5 km.
+
+
+    Returns
```
aperezhortal:
```suggestion
    Returns
    -------
```

pysteps/verification/spatialscores.py
```diff
@@ -667,6 +667,377 @@ def fss_compute(fss):
...
+        "image_intensity",
+    ]
+    table = pd.DataFrame(
+        regionprops_table(labels, intensity_image=df, properties=properties)
+    )
+    return table
+
+
+def vol(ds, minref, maxref, mindiff, minsize, minmax, mindis):
+    """This function calculates the scaled volume parameter based on Wernli et al (2008).
```
aperezhortal: We tend to use imperative mood in the function description.
```suggestion
    """Calculate the scaled volume parameter based on Wernli et al (2008).
```

pysteps/verification/spatialscores.py
```diff
@@ -667,6 +667,377 @@ def fss_compute(fss):
...
+        "intensity_mean",
+        "image_intensity",
+    ]
+    table = pd.DataFrame(
+        regionprops_table(labels, intensity_image=df, properties=properties)
+    )
+    return table
+
+
+def vol(ds, minref, maxref, mindiff, minsize, minmax, mindis):
```
aperezhortal: The `vol` name is not very descriptive of the function's task. Perhaps something like `scaled_volume_parameter`?

pysteps/verification/spatialscores.py
```diff
@@ -667,6 +667,377 @@ def fss_compute(fss):
...
+                "sum": tot.sum(),
+                "scaled_v": v_ob,
+            }
+        )
+    vol_value = pd.DataFrame(ch)
+    return vol_value
+
+
+def c_m(df):
+    """This function calculates the center of total (precipitation) mass in one time step.
```
aperezhortal:
```suggestion
    """Calculate the center of total (precipitation) mass in one timestep.
```

pysteps/verification/spatialscores.py
```diff
@@ -667,6 +667,377 @@ def fss_compute(fss):
...
+                "number_of_cells": n,
+                "sum": tot.sum(),
+                "scaled_v": v_ob,
+            }
+        )
+    vol_value = pd.DataFrame(ch)
+    return vol_value
+
+
+def c_m(df):
```
aperezhortal: Consider using a more descriptive name for this function. E.g. `precipitation_mass_center`.

pysteps/verification/spatialscores.py
```diff
@@ -667,6 +667,377 @@ def fss_compute(fss):
...
+    cent:
+    The coordinates of center of mass.
+    """
+    from scipy import ndimage
+
+    cent = ndimage.measurements.center_of_mass(np.nan_to_num(df))
+    return cent
+
+
+def Amplitude(ob, pre):
```
aperezhortal: Function names use the `lower_with_under` naming convention. `WordWithCaps` are used to name classes.
To maintain consistency across the code, I suggest renaming the function to `sal_amplitude`.

pysteps/verification/spatialscores.py
```diff
@@ -667,6 +667,377 @@ def fss_compute(fss):
...
+    """This function calculates the amplitude component for SAL based on Wernli et al (2008).
```
aperezhortal:
```suggestion
    """Calculate the amplitude component for SAL based on Wernli et al (2008).
```

pysteps/verification/spatialscores.py
```diff
@@ -667,6 +667,377 @@ def fss_compute(fss):
...
+    a:
+    Amplitude parameter which has a value between -2 to 2.
+    """
+    R_obs = np.nanmean(ob)
+    R_model = np.nanmean(pre)
+    a = (R_model - R_obs) / (0.5 * (R_model + R_obs))
+    return a
+
+
+def l1_param(ob, pre):
```
aperezhortal: Since this function is associated to the SAL method, what about renaming it to something like `sal_l1_param`?

pysteps/verification/spatialscores.py
```diff
@@ -667,6 +667,377 @@ def fss_compute(fss):
...
+    """This function calculates the first parameter of location component for SAL based on Wernli et al (2008).
```
aperezhortal:
```suggestion
    """Calculate the first parameter of location component for SAL based on Wernli et al (2008).
```

pysteps/verification/spatialscores.py
```diff
@@ -667,6 +667,377 @@ def fss_compute(fss):
...
+    return l1
+
+
+def weighted_r(df, minref, maxref, mindiff, minsize, minmax, mindis):
+    """This function is to calculated The weighted averaged distance between the centers of mass of the
+    individual objects and the center of mass of the total precipitation field (Wernli et al, 2008).
+
+    Parameters
+    ----------
+    df: 2-d xarray for a time step containing input precipitation data.
```
aperezhortal: >      df: 2-d xarray for a time step containing input precipitation data.
Xarray or pandas dataframes?

pysteps/verification/spatialscores.py
```diff
@@ -667,6 +667,377 @@ def fss_compute(fss):
...
+        sump = df_obj.image_intensity[i].sum()
+
+        r.append({"sum_dist": sumr, "sum_p": sump})
+    rr = pd.DataFrame(r)
+    w_r = rr.sum_dist.sum() / (rr.sum_p.sum())
+    return w_r
+
+
+def l2_param(df_obs, df_forc, minref, maxref, mindiff, minsize, minmax, mindis):
+    """This function calculates the second parameter of location component for SAL based on Wernli et al (2008).
```
aperezhortal:
```suggestion
    """Calculate the second parameter of location component for SAL based on Wernli et al (2008).
```

aperezhortal (commented):
Hi @EsmailGhaemi, thanks for submitting the PR! 
Verification metrics are very welcome :+1: .

I did a first pass through the PR and I left minor comments regarding the code style.

To maintain consistency across verification metric interfaces, perhaps the auxiliary functions used for the SAL computation can be moved to a separate module (like `sal_utils.py` or `sal.py`).  And implement  `sal`, `sal_init`, `sal_accum`, `sal_merge`, and `sal_compute`  in the `spatialscores.py`  module. By doing so the SAL internals using pandas dataframes and the pysteps.verification function using arrays and dictionaries are kept separated. 

Also, did you consider adding unit tests for the SAL computation? In that way, we can test that everything works as expected before and after the merge.

pysteps/verification/spatialscores.py
dnerini: done in 7d9ce66 and 97cfed9

pysteps/verification/spatialscores.py
dnerini: done in 1816f2d

pysteps/verification/spatialscores.py
dnerini: done in 1816f2d

pysteps/verification/spatialscores.py
dnerini: done in 1816f2d

pysteps/verification/spatialscores.py
dnerini: done in 1816f2d

pysteps/verification/spatialscores.py
dnerini: done in 1816f2d

dnerini:
Thanks for the PR @EsmailGhaemi !

I took the liberty to advance with the more obvious changes, including the suggestions from @aperezhortal.
If possible, I'd like to ask @feldmann-m for a review, since this implementation now heavily relies on her feature detection method.

Next, I'll work on some more refactoring, including the use of the _init, _accum, and _compute methods.

dnerini:
> To maintain consistency across verification metric interfaces, perhaps the auxiliary functions used for the SAL computation can be moved to a separate module (like `sal_utils.py` or `sal.py`). And implement `sal`, `sal_init`, `sal_accum`, `sal_merge`, and `sal_compute` in the `spatialscores.py` module. By doing so the SAL internals using pandas dataframes and the pysteps.verification function using arrays and dictionaries are kept separated.

I also thought about it, and at the moment I'm more of the idea that SAL can be considered a separate case as init, accum, merge, and compute do not really apply here... thus commit 71033e4 extract all the SAL methods into a separate module. What do you think?

pysteps/verification/sal.py
```diff
@@ -0,0 +1,357 @@
...
+    ----------
+    :cite: Wernli, H., Hofmann, C., & Zimmer, M. (2009).
+    :cite: Feldmann, M., Germann, U., Gabella, M., & Berne, A. (2021).
+
+    Notes
+    -----
+    This implementation uses the thunderstorm detection algorithm by Feldmann et al (2021)
+    for the identification of precipitation objects within the considered domain.
+
+    This approach uses multi-threshold algorithm to detect objects, instead of having a
```
dnerini: @EsmailGhaemi do you think it would be difficult to implement the original approach too? Would be nice to actually give the option to use both, as tstorm requires a bit more attention in specifying the parameters ...

aperezhortal:
> I also thought about it, and at the moment I'm more of the idea that SAL can be considered a separate case as init, accum, merge, and compute do not really apply here... thus commit 71033e4 extract all the SAL methods into a separate module. What do you think?

:+1: 

Also, maybe we can import the `sal.sal()` function the verification metrics to make it accessible from the `spatialscores`. 
```
from pysteps.verification import spatialscores

spatialscores.sal(array1, array2)
```
In this way, all the spatial metrics can be found in the same place.

pysteps/verification/salscores.py
```diff
@@ -0,0 +1,345 @@
...
+    This approach uses multi-threshold algorithm to detect objects, instead of having a
+    single threshold (f).
+
+    See also
+    --------
+    pysteps.feature.tstorm
+    """
+    if np.nanmax(observation >= 0.1) & np.nanmax(
+        prediction >= 0.1
+    ):  # to avoid errors of nan values or very low precipitation
```
dnerini: I'm not convinced by this ... can we do it better? 

I guess the absence of precipitation is problematic for Structure and Location, but not really for Amplitude? Can we make the code robust enough? Or at list make the threshold an optional argument ...

pysteps/verification/sal.py
dnerini: See c6022c2.

@feldmann-m you think it makes sense to set the parameters of tstorm.detection as

```
        tstorm_kwargs = {
            "minsize":  5,
            "minmax": 0,
            "mindis": 5,
            "maxref": np.Inf,
            "minref": 1/15 * np.nanmax(precip),
        }
```

to obtain a simpler object detection based solely on a detection threshold (1 / 15 * precip_max)?

EsmailGhaemi: > @EsmailGhaemi do you think it would be difficult to implement the original approach too? Would be nice to actually give the option to use both, as tstorm requires a bit more attention in specifying the parameters ...

Hi and thanks a lot for making it way better :) It is a good idea if we have the original approach using f value as well. I think it would be better to have a separate function for that. For instance, sal_single_threshold. If you agree, I'll add a function for object detection using f value here.

pysteps/verification/salscores.py
EsmailGhaemi: I agree it is not the best solution. Maybe If there is no precipitation in observation or forecast, just show a message "No object is detected in observation/ forecast"?

Also, I think we need to mention that scikit-image package should be 0.18 or higher since the name of some functions are different in older versions.

pysteps/verification/sal.py
dnerini: Hi @EsmailGhaemi, I tried to implement the single threshold approach in commit c6022c2. Please have a look at the code and see if it makes sense to you.

pysteps/verification/salscores.py
dnerini: > I agree it is not the best solution. Maybe If there is no precipitation in observation or forecast, just show a message "No object is detected in observation/ forecast"?

mmmh I wonder if it should not up to the end user to filter out cases without enough precipitation. We should always return a value when possible, and maybe NaN if no cells are found for those components that depend on object detection (structure and location).

> Also, I think we need to mention that scikit-image package should be 0.18 or higher since the name of some functions are different in older versions.

Yep I realized that: the properties names of `regionprops` changed with v0.19 (see https://github.com/scikit-image/scikit-image/releases). I tried to use the old names (max_intensity, intensity_image), since these are supposed to be still supported in higher versions, but the last test failed and I still need to have a look into it...

pysteps/verification/sal.py
EsmailGhaemi: Yes, fantastic! 
However, Wernli et al., (2009) suggested to use the 95th quantile (larger than 0.1 mm), instead of maximum value. If you agree, I'll change it to 95th quantile ((1/15) * precip_95quantile).

pysteps/verification/salscores.py
EsmailGhaemi: I agree.

I also tried using the old names but it didn't work. Using higher versions (>=0.18) and the new names solved my problem.

pysteps/verification/sal.py
dnerini: Of course, please do

pysteps/verification/salscores.py
dnerini: Good to know eheh, then I might revert back to the new names and add a check for the sklearn version

dnerini (approved):
I think this is now ready to be merged. There are few remaining minor issues, but these can also be addressed with separate PRs, e.g.:
- add init, accum, merge, compute methods for aggregating SAL scores 
- add visualization methods as the ones used in the original paper

before merging, I'd like just to hear from @feldmann-m about the integration of her detection method.

also, @EsmailGhaemi please give a final round of review to the current version: after the recent changes I want to make sure that I didn't introduce any bug in your original code ...

pysteps/verification/salscores.py
```diff
@@ -0,0 +1,386 @@
...
+            "The scikit-image package is required for the SAL "
+            "verification method but it is not installed"
+        )
+    if tstorm_kwargs is None:
+        tstorm_kwargs = dict()
+    if thr_factor is not None:
+        zero_value = np.nanmin(precip)
+        tstorm_kwargs = {
+            "minsize": tstorm_kwargs.get("minsize", 5),
+            "minmax": tstorm_kwargs.get("minmax", 0),
```
feldmann-m: as minmax is a threshold that applies to the field exceeding minref, I would suggest to put minmax=minref to disable it
"0" can be a problem, if minref happens to be negative

pysteps/verification/salscores.py
```diff
@@ -0,0 +1,386 @@
...
+    if not SKIMAGE_IMPORTED:
+        raise MissingOptionalDependency(
+            "The scikit-image package is required for the SAL "
+            "verification method but it is not installed"
+        )
+    if tstorm_kwargs is None:
+        tstorm_kwargs = dict()
+    if thr_factor is not None:
+        zero_value = np.nanmin(precip)
+        tstorm_kwargs = {
```
feldmann-m, line 352: mindiff is not defined here and thus will use the default "valley depth" of 6 to split contiguous areas - to get results consistent with the threshold approach, this should also be defined as a function of the value distribution (e.g. a quantile)

feldmann-m (commented):
From my side the use of the feature detection in tstorm looks good and works fine. I added a few comments on the default values selected for the threshold method. I ran a small test example based on the tstorm example code. Are there plans to add an example for SAL as well?

pysteps/verification/salscores.py
dnerini: good point! What value could we assign to avoid any splitting of contiguous areas?

feldmann-m: To subpress all splits, you can put anything bigger than max-min of the entire field. As some people use this with dBZ fields that have negative values, I would compute the valid range explicitly.

dnerini: done in 8789c26 , could you quickly check please?

pysteps/verification/salscores.py
dnerini: done in 721b23a, thanks!

pysteps/verification/salscores.py
feldmann-m: everything looks good to me!

EsmailGhaemi: Thanks a lot, it looks great. Just a minor suggestion: I suggest we use the 95th quantile instead of maximum for the max_min value as well. 

For plotting: I think using a scatter plot with the L component as the color arg would work. The input can be sal or accum_sal. I'll add it ASAP.

feldmann-m: To completely disable the splitting, max_min needs to cover the complete range from the minimal to the maximal value, not just the 5th-95th percentile. Perhaps another solution could be to put np.Inf here, similar to the max_ref threshold?

EsmailGhaemi: Hmm, I think i'm missing something here. The max_min is the same value as mindiff in tstorm (using skim.h_maxima), which is the minimum difference between two maximums in the area to separate it into two objects. 
So if this value is more than maximum (here is max_min +1), no object is detected.

feldmann-m: Ah, I see, my mistake. Can you try min_diff=np.nanmax(precip)-minref ?

dnerini: Yep we got wrong the mindiff parameter, well spotted!
I ended up with a new set of parameters, see 6b88dcd
The key is to set a very low maxref to practically exclude mindiff and only rely on the minimum threshold.

Example:
image

Perhaps not perfect, but close enough to the algorithm in the original papers? what do you think?

EsmailGhaemi: Thank you, yes. With these thresholds, we are closer to the original papers.
And it is good that users have the option to change the definition of precipitation 'object', so they can change it based on their study areas, precipitation regimes, etc.

dnerini:
Tests for python=3.6 are still failing...

I have troubles reproducing the error locally (still working on it). 

Also, #253 might solve the problem by setting the minimum python version to 3.7 ...

dnerini:
Setting the minimum python version to 3.7 (d92904a) fixed the remaining failing tests.

I'm merging this PR, thanks again to @EsmailGhaemi for the original contribution and to @aperezhortal and @feldmann-m for the valuable inputs during the review!

> For plotting: I think using a scatter plot with the L component as the color arg would work. The input can be sal or accum_sal. I'll add it ASAP.

@EsmailGhaemi it's an excellent idea to provide such a plotting functionality! I suggest to introduce it with a separate PR, perhaps together with an example script for the gallery?

PR #251, opened 2022-01-04
[WIP] Decouple examples gallery from the main repository

### LINKED ISSUE #242: Decouple gallery from main documentation
Currently the pysteps example gallery is part of the main pysteps repo, meaning that the static example pages are compiled by readthedocs for each change. 

This slows down considerably the rtd builds, and puts us very much at the limit of their free resources. 

One solution could be to fully decouple the example gallery (as a new repo, eg "pysteps-gallery") that uses GitHub actions to build and publish the example pages (although I'm not sure exactly if and how this could be possible yet).

The main advantages being that now the examples can be build only when is really necessary (eg a new release of the main package), the rtd builds wil take a fraction of the time it currently takes, and the examples will possibly have more visibility thanks to a dedicated repo.

Another interesting feature could be to use jupyter notebooks for all examples. We could then use [nbsphinx](https://nbsphinx.readthedocs.io/) to build the html pages, while also giving the possibility to open them directly with Google Colab.

aperezhortal:
## Problem

Currently, the examples gallery is a part of the pysteps project and the examples are executed during the documentation build in Read The Docs (RTD). Running these examples is a time-consuming operation that puts us close to the RTD's build limits. 
 
## Proposed solution

The discussions in #242 converge on decoupling the example gallery from the main repo, keeping the examples in a separate repository that uses Github Actions to execute the examples and build the examples gallery. The rendered gallery is `pulled` from the separate repo when the pysteps documentation is built in RTD.

## Changes introduced in this PR

- Remove the examples files (`examples/*.py` files) from the pysteps repo. The examples were moved to the [pysteps_tutorials](https://github.com/pySTEPS/pysteps_tutorials) repository. The `pysteps_tutorials` repository contains only the Sphinx documentation for the example gallery. Here, the GitHub actions render the example gallery (using sphinx-gallery) and save the rendered gallery (rst and png files) in separate branches. **IMPORTANT**, the examples have the form of python scripts for now. Although this can be updated to jupyter notebooks if needed. 
- Update the documentation build process to pull the rendered example gallery from the `pysteps_tutorials` repository.
- See the generated documentation in https://pysteps.readthedocs.io/en/external_gallery/
#### Should we use Jupyter notebooks or python scripts for the example gallery? 

I personally prefer python scripts. Although jupyter notebooks are good for exploratory analysis and can be nicely visualized in Github, keeping track of the changes is very difficult since the diffs are not very human-readable. On the other hand, keeping track of the differences in python scripts is straightforward. Another advantage of using python scripts is that we can benefit from the static code analysis of the IDEs.

Also, `sphinx-gallery` automatically converts the python scripts to jupyter notebooks and makes them available in the documentation. There is an option (experimental) to add the `open in binder` button that we may explore. 

I check how the example galleries are handled in other projects and some use jupyter notebooks while others use scripts:
- [wradlib](https://github.com/wradlib/wradlib-notebooks): Jupyter notebooks in a separate repository.
- [Astropy](https://github.com/astropy/astropy-notebooks): Jupyter notebooks in a separate repository.
- [fastai](https://github.com/fastai/fastai/tree/master/nbs/examples): Jupyter notebooks in the main repository.
- [Scikit-learn](https://github.com/scikit-learn/scikit-learn/tree/main/examples): Python scripts in the project repository.
- [Pytorch](https://github.com/pytorch/examples): Python scripts in the project repository.
- [Tensorflow](https://github.com/tensorflow/examples): Python scripts in the project repository.
- [Matplotlib](https://github.com/matplotlib/matplotlib/tree/main/examples): Python scripts in the project repository.

It seems that it boils down to a matter of personal preference. Both approaches are good, and if we choose to use notebooks,  updating the current PR is straightforward.

What are your thoughts?


doc/source/user_guide/index.rst
```diff
@@ -16,6 +16,6 @@ the package, see the :ref:`pysteps-reference`.
     install_pysteps
     example_data
     set_pystepsrc
-    ../auto_examples/index
+    ../examples_gallery/index
```
dnerini, line 19: so the url to the example gallery will change, correct? Then I think there are few more places (e.g. README, github page) where we need to update links.

.github/workflows/execute_tutorial_notebooks_in_external_repo.yml
```diff
@@ -0,0 +1,88 @@
...
+#
+# "main repository": Project repository with the python package and the documentation.
+# "linked repository": Repository where the example gallery of jupyter notebooks is
+#                      stored.
+########################################################################################
+name: Trigger notebook execution in the external repository
+
+env:
+  # URL for the external repository linked with the notebooks in this project.
+  LINKED_REPO: pysteps/pysteps_tutorials
```
dnerini: I'm just being annoying here, but for other pysteps-related repos we used the hyphen, so what about changing the name to `pysteps-tutorials`?

.github/workflows/execute_tutorial_notebooks_in_external_repo.yml
```diff
@@ -0,0 +1,88 @@
+########################################################################################
+# Trigger the notebooks' execution in the external repository.
+#
+# If the "main" branch triggered this event, then the nb execution is triggered in the
+# "main" branch, otherwise in the dev branch.
```
dnerini: I really like this idea! So, did I understand correctly that all branches but master will automatically render into "dev"?

.github/workflows/execute_tutorial_notebooks_in_external_repo.yml
```diff
@@ -0,0 +1,88 @@
...
+env:
+  # URL for the external repository linked with the notebooks in this project.
+  LINKED_REPO: pysteps/pysteps_tutorials
+  THIS_REPO: pysteps/pysteps
+
+on:
+  push:
+    branches: [ main, dev ]
+  pull_request:
+    branches: [ main, dev ]
```
dnerini: any push in any branch will trigger or only pushes within pull requests?

dnerini (commented):
Fantastic job @aperezhortal!

I only have few understanding questions and there are a couple of super minor changes that we need to include (basically updating the references to the gallery page).

dnerini:
I also noticed that `pysteps_tutorials` uses `dev` as default branch. Is this intentional?

.github/workflows/execute_tutorial_notebooks_in_external_repo.yml
aperezhortal: With that configuration only pushes, or PR, to the `main` or the `dev`. Other branches are disabled by default. In that way, we can control if other development branches should trigger the examples rendering or not.

aperezhortal:
> I also noticed that pysteps_tutorials uses dev as default branch. Is this intentional?

Just a consequence of the branch that I was using :smile: 
I pushed the dev commits to the main branch, and set the latter as default.

.github/workflows/execute_tutorial_notebooks_in_external_repo.yml
aperezhortal: Yes. However, right now only the main branch is pull in the examples repo. I will add something to pull the same branch that sent the triggering commit to the external repo. In that way, we can test if the examples execute correctly with development branches.

dnerini:
I have another annoying comment (sorry): it should be "Example gallery" all over instead of "Examples gallery", shouldn't it?

aperezhortal:
> ... it should be "Example gallery" all over instead of "Examples gallery"

Yes, you are right. I will fix it shortly. 

I will also change the PR to Work In progress since I'm experimenting with a slightly different workflow where the rendering is triggered by empty commit messages to a dedicated branch. This allows finer control on which pysteps branch to pull, which branch in the gallery to render, and what RTD branch to trigger. :nerd_face:

aperezhortal:
PR #262 implements a different workflow that I think will be easier to maintain since the example gallery is not decoupled. 

Do we keep this PR open? Or do we close it in favor of  #262?

dnerini:
Yes, I agree, let's close this PR in favor of #262.

PR #257, opened 2022-01-15
Replace miniconda by micromamba in CI

aperezhortal:
This PR takes #256 one step further using the `mamba-org/provision-with-micromamba@main` action to create the environment, instead of `conda-incubator/setup-miniconda@v2`

Although the computational time saved is minimal (a minute at the most), it simplifies a little bit the workflow specification, condensing the previous "miniconda installation" and "environment creation" task into a single one. 

More information on micromamba: https://github.com/mamba-org/mamba#micromamba

### DIFF
--- a/.github/workflows/test_pysteps.yml
+++ b/.github/workflows/test_pysteps.yml
@@ -1,10 +1,5 @@
 name: Test Pysteps
 
-env:
-  MINIMAL_DEPENDENCIES: cython numpy jsmin jsonschema matplotlib netCDF4 opencv pillow pyproj scipy dask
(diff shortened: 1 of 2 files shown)


dnerini (approved):
Ah this is real nice, I didn't know about a micromamba!

kmuehlbauer:
@aperezhortal Nice, I was just doing the same over at `h5netcdf`, but using the old style downloading via snakepit (as `wradlib` does). And then I see this new provisioning thing popping up in your commits. That saved me quite some keyboard time. Just wanted to say thanks :grinning:

dnerini:
@aperezhortal you wanna merge this or is it still work in progress?

PR #271, opened 2022-02-18
Revert some changes from #255

dnerini:
Some changes that were meant for #254 were wrongly included in #255. This PR tries to revert them.


aperezhortal (approved):
Good catch! It looks to go after the tests passed.

PR #272, opened 2022-02-22
Minor exporters adjustment

RubenImhoff:
For a deterministic run (n_ens_mem = 1), the shape of var_f is [n_timesteps, y, x]. The exporter assumed a shape of [n_ens_members, n_timesteps, y, x], which is no longer the case (changed a few releases ago). Hence, the indexation was incorrect (returning the shape of y instead of n_timesteps).

### DIFF
--- a/pysteps/io/exporters.py
+++ b/pysteps/io/exporters.py
@@ -804,7 +804,7 @@ def _export_netcdf(field, exporter):
         if exporter["num_ens_members"] > 1:
             var_f[:, var_f.shape[1], :, :] = field
         else:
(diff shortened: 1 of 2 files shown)


dnerini:
Hi @RubenImhoff, thanks for spotting the problem and suggesting the fix. Would it be possible to include a test that would have detected the issue in the first place?

RubenImhoff:
I think so, I'll have a look at it!

dnerini (approved):
Very nice, thanks a lot also for adding the new test!
This can be merged in my opinion.

PR #273, opened 2022-03-14
Place ensemble member number determination for blending inside forecast loop to prevent out of memory issues

RubenImhoff:
In the recently posted release, the blending code determines which NWP models will be combined with which nowcast ensemble members. It does this at once and creates a variable that contains [n_ens_members, n_timesteps, n_cascade_levels, y, x]. For a large number of time steps and ensemble members, this variable can become too big too handle. 

To overcome this issue, this PR tries to implement this procedure within the forecast loop (so per time step instead of all at once), which highly reduces the memory requirements.

### DIFF
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -16,14 +16,13 @@ consists of the following main steps:
        method to decompose and store the NWP model fields whenever a new NWP model
        field is present, is present in pysteps.blending.utils.decompose_NWP.
     #. Estimate AR parameters for the extrapolation nowcast and noise cascade.
-    #. Before starting the forecast loop, determine which NWP models will be
-       combined with which nowcast ensemble members. The number of output ensemble
-       members equals the maximum number of (ensemble) members in the input, which
-       can be either the defined number of (nowcast) ensemble members or the number
-       of NWP models/members.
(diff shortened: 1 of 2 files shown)


dnerini (approved):
@RubenImhoff I was able only to quickly scan through your changes, but unfortunately I could not look more in depth into it... Are there critical parts where you would need a more detailed feedback? Otherwise this looks to me, happy to see that this piece of code keeps improving!

ps: one small detail, can you maybe provide a more specific title to this PR? Thanks!

RubenImhoff:
> Are there critical parts where you would need a more detailed feedback?

Not really, all my own tests seem fine so far. The only small critical point is that it may slow down the forecast loop a bit (but probably not significantly). 

> one small detail, can you maybe provide a more specific title to this PR?

Will do so!

dnerini:
one last thought: since you mention the memory usage, would you care to look into a memory profiling for a single run? using `memory_profiler` would as easy as running 
```
mprof run python <your example script>.py
mprof plot
```

with this you could easily compare the impact of your new version.

RubenImhoff:
New code:
image

Old code:
Out-of-memory error, requested >100 GB according to the command line (I doubt it was that much, but defenitely too much for my laptop)

PR #274, opened 2022-03-21
Ensure clim window has same size as skill today and place forecast_prev update at end of extrapolation loop

RubenImhoff:
This PR makes two minor changes that impact the climatological skill calculation and the extrapolation loop in the steps blending code when subtimesteps are used. 

Changes:
- When the climatological skill window is saved, it first checks if the arrays have the same size as the skill that will be appended. Before, this resulted in an error when moving to the next day (skill clim window gets saved) when the new forecasts have a different number of ensemble members than the skill clim window. This has been solved now. 
- The update of "forecast_prev", the extrapolation and noise cascade that are fed into the extraplation loop, took place too early in the loop. This caused that the extrapolation loop already uses the next forecast when subtimesteps are used. This PR fixes this issue, too.

### DIFF
--- a/pysteps/blending/clim.py
+++ b/pysteps/blending/clim.py
@@ -60,8 +60,8 @@ def save_skill(
     current_skill,
     validtime,
     outdir_path,
-    n_models=1,
     window_length=30,
+    **kwargs,
 ):
     """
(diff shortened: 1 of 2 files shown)


dnerini (approved):
Looks good to me, thanks for the contribution!

PR #278, opened 2022-04-21
Additions to Seed et al. weights function to prevent numeric instability

RubenImhoff:
This (short) PR prevents possible numeric instabilities when the Seed et al. (2013) weights are implemented.

### DIFF
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -1435,17 +1435,23 @@ def calculate_weights_spn(correlations, cov):
     # Check if the correlations are positive, otherwise rho = 10e-5
     correlations = np.where(correlations < 10e-5, 10e-5, correlations)
 
-    if correlations.shape[0] > 1:
+    if correlations.shape[0] > 1 and len(cov) > 1:
         if isinstance(cov, type(None)):
             raise ValueError("cov must contain a covariance matrix")
         else:
+            print(cov)
             # Make a numpy matrix out of cov and get the inverse
             cov = np.where(cov == 0.0, 10e-5, cov)
-            if cov[0][0] != 1.0:
-                cov[0][0] = 1.0
-            if cov[1][1] != 1.0:
-                cov[1][1] = 1.0
+            # Make sure the determinant of the matrix is not zero, otherwise
+            # subtract 10e-5 from the cross-correlations between the models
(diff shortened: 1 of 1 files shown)


pysteps/blending/steps.py
```diff
@@ -1440,19 +1440,27 @@ def calculate_weights_spn(correlations, cov):
             raise ValueError("cov must contain a covariance matrix")
         else:
             # Make a numpy matrix out of cov and get the inverse
+            cov = np.where(cov == 0.0, 10e-5, cov)
+            if cov[0][0] != 1.0:
+                cov[0][0] = 1.0
+            if cov[1][1] != 1.0:
+                cov[1][1] = 1.0
```
dnerini: Can you briefly explain lines 1444-1447? Beside it, I guess the if statements are not really necessary? Just set the values to 1, no?

pysteps/blending/steps.py
```diff
@@ -1440,19 +1440,27 @@ def calculate_weights_spn(correlations, cov):
...
             cov_matrix = np.asmatrix(cov)
             cov_matrix_inv = cov_matrix.getI()
```
dnerini, line 1455: do we need to capture LinAlgError in case of a singular matrix or is this dealt somewhere else?

pysteps/blending/steps.py
RubenImhoff: Just setting it to 1.0, works as well! I noticed that in some cases, with e.g. hardly any rain in the NWP forecast, np.corrcoef does not return a correlation of 1.0 for the model with itself. This is to ensure that is the case.

pysteps/blending/steps.py
RubenImhoff: I'll double check that, good point.

dnerini: is this now solved by 27ed33a?

RubenImhoff: Yes!

dnerini (approved):

pysteps/blending/steps.py
dnerini: then I suggest to remove the "if" statements if they're not necessary (makes the code cleaner) and just set the diagonal to one in all cases?

PR #280, opened 2022-04-22
BUG: Fix T-DaTing labels when continuing tracking

ritvje:
Hi! We came across this bug when trying to use the tracking functions. When continuing the tracking from previously tracked list, the tracks were weird, for example when tracking 3 x 20-minute intervals:

Track

This is because when setting the unique IDs for the tracks, the IDs of the previous cells are not considered. This PR fixes that. Here are the tracks with this fix:

Track_fixed

### DIFF
--- a/pysteps/tests/test_tracking_tdating.py
+++ b/pysteps/tests/test_tracking_tdating.py
@@ -15,6 +15,54 @@ arg_values = [
     ("mch", True),
 ]
 
+arg_names_multistep = ("source", "len_timesteps")
+arg_values_multistep = [
+    ("mch", 6),
+]
+
+
+@pytest.mark.parametrize(arg_names_multistep, arg_values_multistep)
+def test_tracking_tdating_dating_multistep(source, len_timesteps):
+    pytest.importorskip("skimage")
+
+    input, metadata = get_precipitation_fields(
+        0, len_timesteps, True, True, 4000, source
+    )
+    input, __ = to_reflectivity(input, metadata)
+
+    timelist = metadata["timestamps"]
+
+    # First half of timesteps
+    tracks_1, cells, labels = dating(
+        input[0 : len_timesteps // 2], timelist[0 : len_timesteps // 2], mintrack=1
+    )
+    # Second half of timesteps
+    tracks_2, cells, _ = dating(
+        input[len_timesteps // 2 - 2 :],
+        timelist[len_timesteps // 2 - 2 :],
+        mintrack=1,
+        start=2,
+        cell_list=cells,
+        label_list=labels,
+    )
+
+    # Since we are adding cells, number of tracks should increase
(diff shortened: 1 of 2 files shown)


dnerini:
Very good, thanks a lot @ritvje for the contribution!

@feldmann-m do you want to have a look at this PR?

dnerini (commented):
The fix makes sense! Before merging, I wonder if we can also add a simple test that would have detected the bug in the first place? 

We already have a simple test in `pysteps/tests/test_tracking_tdating.py`, would you like to give it a try @ritvje ? Once done, I suggest that you try running the test without your fix to make sure that the test fails...

ritvje:
Sure, I can try to write the test! There are some clear indicators of the bug in my data, like that the track list is shorter at the second time interval than first and some tracks are not continuous in time, but I'll have to see if those repeat in the test data.

ritvje:
I now added a test. I tried to test that the logic in the result tracks is correct. However, even without the fix it won't fail for all datasets (for example,  the first timesteps of FMI test data), most likely if there are no unmatched cells in the second interval. Which makes sense, since the bug in setting the ID only affects unmatched cells. But it should fail for this configuration without the fix and pass with it.

Also, when I tried to run the tests with `tox` (as instructed in the [developer guide](https://pysteps.readthedocs.io/en/stable/developer_guide/test_pysteps.html)), it seems that the test (and also the other tdating test) is skipped because skimage is not installed in the environment. Is this fine or should I add skimage there somehow?

pysteps/tests/test_tracking_tdating.py
```diff
@@ -15,6 +15,54 @@
...
+arg_values_multistep = [
+    ("mch", 6),
+]
+
+
+@pytest.mark.parametrize(arg_names_multistep, arg_values_multistep)
+def test_tracking_tdating_dating_multistep(source, len_timesteps):
+    pytest.importorskip("skimage")
+
+    input, metadata = get_precipitation_fields(
```
dnerini: better to avoid redefining built-in method 'input', can your rename with something else, e.g. "input_fields" ?

dnerini (changes requested):
Very nice @ritvje , thanks a lot for adding the test! 
I guess it's fine to run it only with a single dataset that fits this particular case, so I'm happy with your current version, I only included a very minor suggestion.

Maybe let's give some more time to @feldmann-m to have a look into this PR (since she's the original author of the method) before merging into master.

Concerning tox, it's a good point, I'll think about it, but let's not address it in this PR. Thanks for pointing it out!

dnerini (approved):

feldmann-m:
Thank you @ritvje for raising this issue! The fix looks great to me, this should enable the tracking module to continue tracking from existing lists.

feldmann-m (approved):

PR #283, opened 2022-05-08
Cascade scale fix

### LINKED ISSUE #276: Spatial scale values in pySTEPS
Dear pySTEPS experts,

This may be a pySTEPS issue, but just as well it can be an issue with my maths :-) Either way, I would really appreciate if someone could help.

I am having a hard time in trying to understand the relationship between spatial scales and cascade level indices the way it has been implemented in pySTEPS. I try to explain with an example where my problem is.

According to Figure 1 in Pulkkinen et al. (2019) (https://gmd.copernicus.org/articles/12/4185/2019/) the spatial scales corresponding to cascade levels are as follows:

image

The last _Ratio_ column shows the ratio between two successive spatial scales. This ratio is constant for all other pairs of scales but for the first two. I also included the central wavenumbers in the table above (_central_wn_).

The relationship between the central wavenumbers and spatial scales is (except for level 0):

_Dim_ / _central_wn_ / 2

where _Dim_ is the maximum grid dimension. This makes some sense to me as the wavenumber is defined as the reciprocal of the wavelength and the spatial scale as half of the wavelength. But what should happen at level zero where the wavenuber is zero, and hence the wavelength and the spatial scale are infinite? Or not? Anyway, it is the zero cascade level that causes the ratio between subsequent spatial scales not to be constant, which poses me a problem as I will explain in the next paragraph.

My colleagues and myself at Aalto University have mainly used STEPS/pySTEPS in ‘simulation mode’ to create an ensemble of rain fields (see e.g. Niemi et al., 2016, DOI:10.1002/2015WR017521). Now the autocorrelative properties between consecutive rain fields cannot be estimated in the same manner as in nowcasting mode, but a relationship between the life-time and the spatial scale of rainfall structures needs to be established. Here we have followed equations 9-11 in Seed at al. (2014) 

https://www.researchgate.net/publication/264943060_Stochastic_simulation_of_space-time_rainfall_patterns_for_the_Brisbane_river_catchment

From the relationship between the correlation length and the spatial scale (equation 9) one can observe that the logarithm of the spatial scale and the logarithm of the correlation length are linearly related. Parameters a and b (equation 9) can hence be estimated as the intercept and slope of the line that forms when ln(correlation length) is plotted against ln(spatial scale). I attach below such figure for one event. 

image

The grey line represents the data where spatial scale values have been calculated as in Figure 1 in Pulkkinen et al. (2019) (see above). One can clearly see the “jump” as a longer horizontal distance between the two last points. Also, it seems that as the consequence of the “jump” the last point does not fall on the line. The orange line is otherwise exactly the same, except for the spatial scale value of the last point, which has been modified so that the ratio between spatial scale values for neighbouring cascades remains constant. In practice, in the orange line the second last spatial scale value of 44 (ln(44) = 3.78) has been multiplied with the constant scale ratio of 2.1 to yield for the largest scale a spatial scale value of 93.8 (ln(93.8) = 4.5) – instead of 264 (ln(264) = 5.6) given by the pySTEPS convention. This scale ratio (here 2.1), as discussed above, is in pySTEPS constant for all other cascade pairs but for the last two. Now all points lie neatly on the line.

In Alan Seed’s and Australian Bureau of Meteorology’s C++ STEPS version the spatial scales are computed as follows (abbreviated):

r4ScaleRatio = pow( 2.0/(double)i4CascadeSize, 1.0/(double)(i4NumberLevels-1)) ;

float r4Scale = i4CascadeSize*r4PixelSize;
for ( i4Level = 0; i4Level < i4NumberLevels; i4Level++ )
{
    r4Scale *= r4ScaleRatio;
}

So evidently the scale ratio is constant for all neighbouring cascade pairs.  This can also be seen from Equation 1 in Seed at al. (2013) (https://doi.org/10.1002/wrcr.20536).

So, I guess my questions are:

Do the spatial scale values in pySTEPS, as it seems, have a “jump” in the scale ratio between the two largest cascade scales? If yes, should it be so? If still yes, how and where has this been implemented (presumably in bandpass_filters.py in filter_gaussian method) in pySTEPS? As pySTEPS does not explicitly compute/report spatial scale values (as far as I know…) we have reverse engineered them from the central wavenumbers (returned by filter_gaussian) and Figure 1 in Pulkkinen et al. (2019) (see above).

Any help is greatly appreciated!

pulkkins:
It was pointed out in Issue #276 by that when using the cascade decomposition with the default parameters, the ratio between successive scales is not constant. There was a gap between the first scale and the remaining ones. However, this can be simply fixed by setting

l_0 = (0.5 * l) ** (1 / (n - 1))

So, I made this the new default value in `cascade.bandpass_filters.filter_gaussian`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 -   repo: https://github.com/psf/black
-    rev: 21.7b0
+    rev: 22.3.0
     hooks:
     - id: black
       language_version: python3
--- a/pysteps/cascade/bandpass_filters.py
+++ b/pysteps/cascade/bandpass_filters.py
@@ -104,7 +104,7 @@ def filter_gaussian(
         between successive spatial scales is constant. This value is
         l_0 = (0.5 * l)**(1 / (n-1)).
     gauss_scale: float
-        Optional scaling prameter. Proportional to the standard deviation of
+        Optional scaling parameter. Proportional to the standard deviation of
         the Gaussian weight functions.
     gauss_scale_0: float
         Optional scaling parameter for the Gaussian function corresponding to
--- a/pysteps/cascade/bandpass_filters.py
+++ b/pysteps/cascade/bandpass_filters.py
@@ -89,7 +89,14 @@ def filter_uniform(shape, n):
 
 
 def filter_gaussian(
-    shape, n, l_0=None, gauss_scale=0.5, gauss_scale_0=0.5, d=1.0, normalize=True
+    shape,
+    n,
+    l_0=None,
+    gauss_scale=0.5,
+    gauss_scale_0=0.5,
+    d=1.0,
+    normalize=True,
+    return_weight_funcs=False,
 ):
     """
     Implements a set of Gaussian bandpass filters in logarithmic frequency
@@ -118,6 +125,9 @@ def filter_gaussian(
     normalize: bool
         If True, normalize the weights so that for any given wavenumber
         they sum to one.
+    return_weight_funcs: bool
+        If True, add callable weight functions to the output dictionary with
+        the key 'weight_funcs'.
 
     Returns
     -------
@@ -188,6 +198,9 @@ def filter_gaussian(
     central_freqs[-1] = 0.5  # Nyquist freq
     central_freqs = 1.0 * d * central_freqs
     out["central_freqs"] = central_freqs
+
+    if return_weight_funcs:
+        out["weight_funcs"] = wfs
 
     return out
 
--- a/pysteps/cascade/bandpass_filters.py
+++ b/pysteps/cascade/bandpass_filters.py
@@ -63,6 +63,10 @@ def filter_uniform(shape, n):
     n: int
         Not used. Needed for compatibility with the filter interface.
 
+    Returns
+    -------
+    out: dict
+        A dictionary containing the filter.
     """
     del n  # Unused
 
--- a/pysteps/cascade/bandpass_filters.py
+++ b/pysteps/cascade/bandpass_filters.py
@@ -181,9 +181,11 @@ def filter_gaussian(
 
     for i in range(len(wfs)):
         if i == 0 and include_mean:
-            continue
-        weights_1d[i, 0] = 0.0
-        weights_2d[i, 0, 0] = 0.0
+            weights_1d[i, 0] = 1.0
+            weights_2d[i, 0, 0] = 1.0
+        else:
+            weights_1d[i, 0] = 0.0
+            weights_2d[i, 0, 0] = 0.0
 
     out = {"weights_1d": weights_1d, "weights_2d": weights_2d}
     out["shape"] = shape
--- a/pysteps/cascade/bandpass_filters.py
+++ b/pysteps/cascade/bandpass_filters.py
@@ -102,7 +102,7 @@ def filter_gaussian(
         Central frequency of the second band (the first band is always centered
         at zero). If set to None, l_0 is chosen automatically so that the ratio
         between successive spatial scales is constant. This value is
-        l_0 = (0.5 * l)**(1 / (n-1)).
+        l_0 = (0.5 * max(shape[0], shape[1])) ** (1 / (n-1)).
     gauss_scale: float
         Optional scaling parameter. Proportional to the standard deviation of
         the Gaussian weight functions.
--- a/pysteps/cascade/bandpass_filters.py
+++ b/pysteps/cascade/bandpass_filters.py
@@ -134,8 +134,10 @@ def filter_gaussian(
     except TypeError:
         height, width = (shape, shape)
 
+    max_length = max(width, height)
+
     if l_0 is None:
-        l_0 = (0.5 * np.min(shape)) ** (1 / (n - 1))
+        l_0 = (0.5 * max_length) ** (1 / (n - 1))
 
     rx = np.s_[: int(width / 2) + 1]
 
@@ -148,8 +150,6 @@ def filter_gaussian(
     dy = int(height / 2) if height % 2 == 0 else int(height / 2) + 1
 
     r_2d = np.roll(np.sqrt(x_grid * x_grid + y_grid * y_grid), dy, axis=0)
-
-    max_length = max(width, height)
 
     r_max = int(max_length / 2) + 1
     r_1d = np.arange(r_max)
--- a/pysteps/cascade/bandpass_filters.py
+++ b/pysteps/cascade/bandpass_filters.py
@@ -180,7 +180,7 @@ def filter_gaussian(
             weights_2d[k, :, :] /= weights_2d_sum
 
     for i in range(len(wfs)):
-        if include_mean:
+        if i == 0 and include_mean:
             continue
         weights_1d[i, 0] = 0.0
         weights_2d[i, 0, 0] = 0.0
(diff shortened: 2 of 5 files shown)


dnerini (commented):
I like the fix, quite elegant! Just wondering, do we have to expect any significant change in the behavior of the cascade decomposition with this? I'm running the example gallery on the branch to have a look at our examples, let's see if we can see any difference...

edit: couldn't spot any difference in the [rendered examples](https://pysteps.readthedocs.io/en/cascade_scale_fix/auto_examples/index.html).

Also, would probably make sense that also @tkokkone has a look at this changes before merging.

tkokkone:
Thanks to @pulkkins for addressing this, and sorry it has taken awhile for me to have a look. I am afraid I still have a slight hickup with understanding the spatial scale value related to the central wavenumber of zero. Let us consider a 512 x 512 grid with 6 cascade levels. This new, updated filter_gaussian gives then the following central wavenumbers (WN):
 
image


So if the relationship between wavenumbers and spatial scale values is as I presume in Issue https://github.com/pySTEPS/pysteps/issues/276, then the corresponding spatial scale values are as shown above in column Scale for levels 1-5. Now if the scale ratio is constant between all levels, the spatial scale value for level 0 (WN = 0) should be 256. This equals the grid dimension (512) divided by two. But in Figure 1 in Pulkkinen et al. (2019) (https://gmd.copernicus.org/articles/12/4185/2019/) the spatial scale for WN=0 equals the grid dimension. But should it be grid dimension divided by two?

I am much more an engineer than a mathematician, but still it would be nice to understand how the wavenumbers and spatial scale values are related to each other. Any help is appreciated!

pulkkins:
> Thanks to @pulkkins for addressing this, and sorry it has taken awhile for me to have a look. I am afraid I still have a slight hickup with understanding the spatial scale value related to the central wavenumber of zero. Let us consider a 512 x 512 grid with 6 cascade levels. This new, updated filter_gaussian gives then the following central wavenumbers (WN):
> 
> image
> 
> So if the relationship between wavenumbers and spatial scale values is as I presume in Issue #276, then the corresponding spatial scale values are as shown above in column Scale for levels 1-5. Now if the scale ratio is constant between all levels, the spatial scale value for level 0 (WN = 0) should be 256. This equals the grid dimension (512) divided by two. But in Figure 1 in Pulkkinen et al. (2019) (https://gmd.copernicus.org/articles/12/4185/2019/) the spatial scale for WN=0 equals the grid dimension. But should it be grid dimension divided by two?
> 
> I am much more an engineer than a mathematician, but still it would be nice to understand how the wavenumbers and spatial scale values are related to each other. Any help is appreciated!

I'm using this relation between the Fourier wavenumbers and the corresponding spatial scales:

`scale (km) = 0.5 * grid_size (pixels) / wavenumber * grid_resolution (km)`

For the first scale (wavenumber is 0), this is not defined. Since the zeroth Fourier wavenumber corresponds to the field mean, my interpretation is that it should correspond to the grid dimension, not the dimension divided by two. The latter would correspond to wavenumber 1.

pulkkins:
I tested another modification of the formula for l_0. By choosing
`l_0 = (0.5 * l / (2 ** (n - 2))) ** (1 / (n - 1))`
we get the following bandpass filters:
bandpass_filter_weights_fmi

In this case, the ratio between each successive scale is constant and the first scale is the grid dimension. I also suggest replacing the Gaussian function of the first filter with an indicator function f such that f(0)=1 and f(x)=0 for x > 0. All other filters are set to zero for the 0-th wavenumber. This choice is justified because the 0-th wavenumber corresponds to the field mean, which in my opinion should be treated separately. Does this make sense?

Edit: If we make the above modification, the first cascade level should not be included at all in the decomposition. It's only purpose would be to represent the field mean. This can be subtracted before doing the decomposition.

pulkkins:
Another option would be to completely omit the first filter centered at zero frequency. Following the approach of https://www.researchgate.net/publication/264943060_Stochastic_simulation_of_space-time_rainfall_patterns_for_the_Brisbane_river_catchment, I also modified the filter construction so that for each one we choose the midpoint of the corresponding interval as the central frequency. This results to the following filter configuration:
bandpass_filter_weights_fmi

@dnerini and @tkokkone What do you think about this?

dnerini:
Thanks Seppo for the interesting discussion. I quite like your last option where you chose to omit the zeroth wavenumber, which is anyways undefined. Would this assume that the field is centered before the decomposition?

tkokkone:
I also like this last alternative as we would not need to worry about the troublesome spatial scale related to the zeroth wavenumber but could simply use the relationship Seppo gave above.  BTW, should the y-axis in the figure above be scaled so that the weights sum up to one? So it seems when comparing with Figure 1 in https://gmd.copernicus.org/articles/12/4185/2019/. But this could be just a matter of presentation.

pulkkins:
> Thanks Seppo for the interesting discussion. I quite like your last option where you chose to omit the zeroth wavenumber, which is anyways undefined. Would this assume that the field is centered before the decomposition?

Yes. I think I will do this by subtracting the mean value before the decomposition and then adding it back after recomposing the cascade.

pulkkins:
> I also like this last alternative as we would not need to worry about the troublesome spatial scale related to the zeroth wavenumber but could simply use the relationship Seppo gave above. BTW, should the y-axis in the figure above be scaled so that the weights sum up to one? So it seems when comparing with Figure 1 in https://gmd.copernicus.org/articles/12/4185/2019/. But this could be just a matter of presentation.

Here is the same figure with normalized weights:
bandpass_filter_weights_fmi

dnerini:
I noticed that in [ca1c5c4](https://github.com/pySTEPS/pysteps/pull/283/commits/ca1c5c4a3c772fbdbe02ae1694d04f43e3d9b650) you adjusted the accuracy thresholds to pass the tests. I'm not against this, but I'd like to understand better the cause. Should we be concerned by this drop of quality? Or is it mostly caused by a lower spread in predictions because of the new cascade decomposition?

pulkkins:
> I noticed that in [ca1c5c4](https://github.com/pySTEPS/pysteps/pull/283/commits/ca1c5c4a3c772fbdbe02ae1694d04f43e3d9b650) you adjusted the accuracy thresholds to pass the tests. I'm not against this, but I'd like to understand better the cause. Should we be concerned by this drop of quality? Or is it mostly caused by a lower spread in predictions because of the new cascade decomposition?

Apart from those three cases, where I modified the thresholds, the CRPS was in fact slightly better with the new filter configuration. Anyway, we need to do a careful evaluation of any changes in the forecast skill before merging this pull request.

pulkkins:
I did some verification experiments with the old and new bandpass filters. Here are the results with the FMI data in the [repository](https://github.com/pySTEPS/pysteps-data) (nowcast start times between 16:00-16:30 UTC on September 28th 2016). The models were run with their default configurations (STEPS with velocity perturbations disabled).

It seems that the new filters give a slightly better forecast skill. However, as noted above, the skill with the new filters can also be slightly lower in some cases. Should we merge this pull request?

Deterministic metrics for S-PROG nowcasts:
ETS_0 1
ETS_1 0
ETS_5 0
MAE

CRPS for STEPS nowcasts:
CRPS

dnerini:
This looks great, thanks Seppo for the analysis! Yes, all good on my side, you can merge into master. Great work!

dnerini (approved):

PR #284, opened 2022-05-08
Fmi geotiff importer fixes

pulkkins:
This pull request contains the following fixes to the FMI GeoTIFF importer (`io.importers.import_fmi_geotiff`):
- remove unnecessary UTF-8 encoding lines
- import all gdal modules from osgeo instead of the global namespace (required by GDAL>=3.2)
- remove the GetScale() or GetOffset() lines because these are hard-coded in the files
- set the Z(R) relation parameters in the metadata to zr_a=223 and zr_b=1.53
- instead of None, set the unit to dBZ in the metadata
- set transform to 'dB' in the metadata
- add separate section for FMI GeoTIFF files in pystepsrc
- add tests for the FMI GeoTIFF importer
- add gdal as optional dependency in the user guide
- update black to version 22.3.0 in .pre-commit-config.yaml

### DIFF
--- a/pysteps/io/importers.py
+++ b/pysteps/io/importers.py
@@ -577,9 +577,12 @@ def import_fmi_geotiff(filename, **kwargs):
     f = gdal.Open(filename, gdalconst.GA_ReadOnly)
 
     rb = f.GetRasterBand(1)
-    precip = rb.ReadAsArray()
+    precip = rb.ReadAsArray().astype(float)
     mask = precip == 255
-    precip = precip.astype(float) * rb.GetScale() + rb.GetOffset()
+    if rb.GetScale() is not None:
+        precip *= rb.GetScale()
+    if rb.GetOffset() is not None:
+        precip += rb.GetOffset()
     precip = (precip - 64.0) / 2.0
     precip[mask] = np.nan
 
--- a/pysteps/io/importers.py
+++ b/pysteps/io/importers.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 """
 pysteps.io.importers
 ====================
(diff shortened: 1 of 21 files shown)


RubenImhoff (approved):

pysteps/io/importers.py
```diff
@@ -579,9 +577,12 @@ def import_fmi_geotiff(filename, **kwargs):
...
 
     rb = f.GetRasterBand(1)
-    precip = rb.ReadAsArray()
+    precip = rb.ReadAsArray().astype(float)
     mask = precip == 255
-    precip = precip.astype(float) * rb.GetScale() + rb.GetOffset()
+    if rb.GetScale() is not None:
+        precip *= rb.GetScale()
+    if rb.GetOffset() is not None:
+        precip += rb.GetOffset()
```
RubenImhoff: Would it be possible to add a test for these lines?

dnerini (approved):

requirements.txt
```diff
@@ -7,3 +7,4 @@ matplotlib
 jsmin
 jsonschema
 netCDF4
+gdal
```
dnerini: why adding gdal here? this file should include only the base dependencies. Also, are you sure that pip is the right way to install GDAL? 

For the tests, you need to list the dependency in the `ci/ci_test_env.yml` file, see [946c97f](https://github.com/pySTEPS/pysteps/pull/284/commits/946c97fca0ef9aa031323cfdcf051d5ba30e7ac8)

I also suggest that you add GDAL as an optional dependency for the geotiff FMI importer in the [user guide](https://github.com/pySTEPS/pysteps/blob/master/doc/source/user_guide/install_pysteps.rst).

requirements_dev.txt
```diff
@@ -18,6 +18,7 @@ h5py
 scikit-image
 pandas
 rasterio
+gdal
```
dnerini: same as above, I would not include it in here...

dnerini (changes requested):

requirements_dev.txt
pulkkins: You are right. I'll remove those.

dnerini (approved):

dnerini:
Very good, thanks, you can go ahead and merge into master once all tests are green.

PR #289, opened 2022-07-04
Allow Seed et al. weights to become negative

RubenImhoff:
We had fixed the Seed et al. weights to not become negative, but since we are calculating in a log-space anyway, this is possible and does not result in negative rainfall amounts. Allowing one of the weights to become negative is also more advantageous from the covariance matrix based weight determination perspective.

Tihs PR is just a minor change of two lines.


dnerini (approved):
LGTM!

PR #290, opened 2022-07-14
Extrapolation fixes

pulkkins:
This pull requests fixes inconsistencies and bugs related to handling of NaN values in the semi-Lagrangian extrapolation:
- Set the keyword argument allow_nonfinite_values automatically based on the input data when calling extrapolation.semilagrangian.extrapolate from the nowcasting methods. Previously this had to be set by the user. In some nowcasting methods (e.g. ANVIL) it was hard-coded to True.
- Fix a bug in extrapolation.semilagrangian.extrapolate. The bug caused NaN values advected from outside the domain to be replaced with the minimum value when interp_order>1.

### DIFF
--- a/pysteps/extrapolation/semilagrangian.py
+++ b/pysteps/extrapolation/semilagrangian.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 """
 pysteps.extrapolation.semilagrangian
 ====================================
--- a/pysteps/extrapolation/semilagrangian.py
+++ b/pysteps/extrapolation/semilagrangian.py
@@ -152,6 +152,8 @@ def extrapolate(
             precip = precip.copy()
             precip[~mask_finite] = 0.0
             mask_finite = mask_finite.astype(float)
+        else:
+            mask_finite = np.ones(precip.shape)
 
     prefilter = True if interp_order > 1 else False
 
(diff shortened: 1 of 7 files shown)


dnerini (approved):
Looks all good to me!  Can you just confirm my impression that this PR doesn't introduce any breaking change?

pulkkins:
> Looks all good to me! Can you just confirm my impression that this PR doesn't introduce any breaking change?

I can confirm that. I also tested all the currently implemented nowcasting methods using input data with and without NaN values. This is not included in the automatic tests, which raises a question. Should we extend the test coverage (although it might significantly increase the run time of the tests)?

dnerini:
> I also tested all the currently implemented nowcasting methods using input data with and without NaN values. This is not included in the automatic tests, which raises a question. Should we extend the test coverage (although it might significantly increase the run time of the tests)?

It’s a fair point, I’d generally go for more test coverage. If we have an issue with computation time, I think we could fairly easily make them more efficient by reducing the size of the test data. Should we take this into a separate issue or would fix that in here?

pulkkins:
> > > Looks all good to me! Can you just confirm my impression that this PR doesn't introduce any breaking change?
> > 
> > 
> > I can confirm that. I also tested all the currently implemented nowcasting methods using input data with and without NaN values. This is not included in the automatic tests, which raises a question. Should we extend the test coverage (although it might significantly increase the run time of the tests)?
> 
> It’s a fair point, I’d generally go for more test coverage. If we have an issue with computation time, I think we could fairly easily make them more efficient by reducing the size of the test data. Should we take this into a separate issue or would fix that in here?

I would take that into a separate issue. It seems that there are quite many gaps in the test coverage of the nowcasting methods.

PR #292, opened 2022-08-03
Improve documentation of rainfarm downscaling

dnerini:
Try to improve the documentation of the `pysteps.downscaling.rainfarm` module based on some recent feedback (see relative discussion on our pysteps slack channel).

### Breaking changes

- `pysteps.downscaling.rainfarm.downscale()`: keyword argument `ds_factor` becomes a positional argument without default value.

### DIFF
--- a/examples/rainfarm_downscale.py
+++ b/examples/rainfarm_downscale.py
@@ -57,11 +57,15 @@ precip[~np.isfinite(precip)] = metadata["zerovalue"]
 # -----------------
 #
 # To test our downscaling method, we first need to upscale the original field to
-# a lower resolution. We are going to use an upscaling factor of 16 x.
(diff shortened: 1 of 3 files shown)


dnerini:
I’m hesitant to merge this PR because of the (marginal?) breaking change listed above.

my feeling is that we can afford to introduce it since it would concern only edge cases (when using positional  arguments for keyword arguments) and it concerns a somewhat experimental module.

Any strong opinion against it here? @pySTEPS/core-devs

aperezhortal (approved):

aperezhortal:
> my feeling is that we can afford to introduce it since it would concern only edge cases (when using positional arguments for keyword arguments) and it concerns a somewhat experimental module.

I agree. We can also add a warning in the next release's notes explaining this.

PR #301, opened 2022-08-28
Fix future warning

feldmann-m:
fixed future warning from scikit-image


pierrejavelle:
The pull request is not describing the issue it is trying to solve.

feldmann-m:
Scikit image raises a future warning when using the keyword "selem" in the h_maxima function. This has been replaced with "footprint", which is the new name for this  keyword.

dnerini (changes requested):
Very nice, thanks a lot @feldmann-m !

One small detail: the new argument name won't work for scikit-image < 0.19, therefore I suggest that you specify this minimum version in the documentation:

in the file `pysteps/doc/source/user_guide/install_pysteps.rst`
please specify `scikit-image >= 0.19`

dnerini (approved):

PR #306, opened 2022-09-02
Fix nowcast importer in case of no rain

### LINKED ISSUE #295: Error when using import_netcdf_pysteps in case of no rain.
Dear developers,

I'm currently trying to deal with no-rain cases when using pySTEPS (blendig).
As you might be aware, there are some issues when running pySTEPS when the radar-images contain no precipitations (see also issue #282). 
As a temporary fix, I use in this 'no-rain case' only the NWP forecasts (basically no pySTEPS run). 
But I do export the results to a netCDF file using the pysteps exporter.
In this no-rain case, also the NWP forecast does not contain any rain, so basically what is in the netCDF file is a timeseries of zeros
When I try to open this netCDF file using ```import_netcdf_pysteps``` I get the following error:
```
import_netcdf_pysteps("./blended_nowcast_20220811141000.nc")
There was an error processing the file zero-size array to reduction operation fmin which has no identity
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Input In [41], in <cell line: 1>()
----> 1 a,b = import_netcdf_pysteps("./blended_nowcast_20220811141000.nc")

File /space/hpc-home/michielv/pylibs/pysteps/pysteps/decorators.py:91, in postprocess_import.<locals>._postprocess_import.<locals>._import_with_postprocessing(*args, **kwargs)
     88         mask = ~np.isfinite(precip)
     89         precip[mask] = _fillna
---> 91 return (precip.astype(_dtype),) + tuple(other_args)

AttributeError: 'NoneType' object has no attribute 'astype'
```

I have traced to error back to this line in the definition of ```import_netcdf_pysteps```:
```
metadata["threshold"] = np.nanmin(precip[precip > np.nanmin(precip)])
```
Which gives an error when ```precip[precip > np.nanmin(precip)]``` is an empty array in case all values are equal (zero in this case).
But this error is catched and a None is returned to the decorator which in turn gives an error.

This can probably be fixed easily with something like:
``` 
if np.nanmin(precip) == np.nanmax(precip):
  metadata["threshold"] = np.nanmin(precip)
else:
  np.nanmin(precip[precip > np.nanmin(precip)])
```

All the best,
Michiel

dnerini:
close #295

### DIFF
--- a/pysteps/io/nowcast_importers.py
+++ b/pysteps/io/nowcast_importers.py
@@ -204,7 +204,10 @@ def import_netcdf_pysteps(filename, onerror="warn", **kwargs):
         metadata["unit"] = unit
         metadata["transform"] = transform
         metadata["zerovalue"] = np.nanmin(precip)
-        metadata["threshold"] = np.nanmin(precip[precip > np.nanmin(precip)])
+        if metadata["zerovalue"] == np.nanmax(precip):
+            metadata["threshold"] = metadata["zerovalue"]
+        else:
+            metadata["threshold"] = np.nanmin(precip[precip > metadata["zerovalue"]])
 
         ds.close()
 
--- a/pysteps/tests/test_io_nowcast_importers.py
+++ b/pysteps/tests/test_io_nowcast_importers.py
@@ -1,0 +1,71 @@
+import numpy as np
+import pytest
+
+from pysteps import io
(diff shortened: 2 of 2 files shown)


RubenImhoff (approved):
Looks great, Daniele!

Issue #307, opened 2022-09-04
Minor bug/ disagreement in lagrangian_probability


aitaten:
The parameter **slope** in _forecast_ function has to be float but parameter **size** in __get_kernel_ function has to be integer. 

Size (actually called scale) is obtained in line 90 with:

`scale = timestep * slope`

Consequently, if slope is actually a float (in the existing test is always an integer), the _get_kernel function returns an error and the _forecast_ function fails without returning the forecast.

dnerini:
Close by #308.

PR #308, opened 2022-09-04
Fix wrong type when passing slope as float

aitaten:
This PR is trying to solve issue #307 by forcing **scale** to be an integer. A test (copy of a previous test) has been added with a float slope to show it is working and fixing the issue. Several other things could be addressed as the name of scale/size disagreement but I am not sure of the original purpose of these two different names so I am not changing that.


dnerini (approved):
Thanks a lot for the bug fix!

(I'll just give it a more precise name before merging)

Issue #287, opened 2022-06-24
RainFARM : Terzago et al. 2018 


simonbeylat:
I wanted to ask you if you could add the version of RainFARM improved by Terzago et al.(https://doi.org/10.5194/nhess-18-2825-2018). 

I would like to use it using Python. By the way, I am open to collaborate and help you to do it.

dnerini:
Hello @simonbeylat 

I'm not familiar with the work of Terzago et al, can you please briefly summarize what they improved with respect to the original RainFARM?

In any case, we would of course welcome your contribution! You can submit a draft of this new RainFARM version as a pull request following our [developer guidelines](https://pysteps.readthedocs.io/en/stable/developer_guide/contributors_guidelines.html). We will then support you during the review of your code, by suggesting code changes and contributing with code to make sure that your contribution will be properly integrated in pysteps. 

Before starting your implementation, it might be worth discussing some higher level details here. For example, should this become a separate method, an option in the existing RainFARM method, or simply replace some existing code? For this, I come back to my initial question: please try to describe in more detail what you'd like to do and how you propose to implement them.

simonbeylat:
Hello @dnerini, 
Thank you for your reply. 

Terzago has added 2 elements to RainFARM. The first is a smoothing operator at the end of the algorithm. 

The second (and most important) is a weight applied at the fine scale of the field, before the last step. 
These weights are calculated from the climatology. It allows RainFARM to reproduce the climatology of fine-scale precipitation. 
The aim is to provide RainFARM with complex orographic information in order to reproduce a realistic fine-scale precipitation field. 

I think we can follow what von Hardenberg (one of the authors of the RainFARM articles) has done in R and Julia and add an option to the current function. (https://github.com/jhardenberg/rainfarmr).

dnerini:
I like your proposition, @simonbeylat! If possible, I'd keep the two changes separate, hence work on two pull requests. The smoothing operator looks like a fairly straightforward addition, while the weighting is somewhat more involved, as it requires a climatological analysis. In this sense, I assume we would need to implement the methods to compute such weights based on a data archive?

Also, keep in mind that currently the rainfarm implementation in pysteps (courtesy of @jleinonen) doesn't include the temporal component, would you be willing to work on this missing part, too? 

Finally, I was  thinking we should get in touch with @jhardenberg, perhaps he's interested in contributing to a python version of his code within pysteps?

simonbeylat:
[...]

PR #311, opened 2022-09-21
Update rainfarm.py

simonbeylat:
Add smoothing operator

see #287

### DIFF
--- a/pysteps/downscaling/rainfarm.py
+++ b/pysteps/downscaling/rainfarm.py
@@ -22,6 +22,7 @@ import warnings
 import numpy as np
 from scipy.ndimage import convolve
 from scipy.ndimage import zoom
+from pysteps.utils import spectral
 
 
 def _log_slope(log_k, log_power_spectrum):
@@ -120,8 +121,48 @@ def _check_smooth_value(smooth):
         )
 
 
+def gaussianize(precip):
+    """
+    Gaussianize field using rank ordering
+
+    Parameters
+    ----------
+    precip : matrix
+        matrix containing the input field to be Gaussianized.
+
+    Returns
+    -------
+        The Gaussianized field with the same dimensions as the input field.
+
+    References
+    ----------
+    :cite:`DOnofrio2014`
+
+    """
+    m = np.shape(precip)[0]
+    n = np.shape(precip)[1]
+    nn = m * n
+    ii = np.argsort(precip.reshape(nn))
+    precip_gaussianize = np.zeros(nn)
+    precip_gaussianize[ii] = sorted(np.random.normal(0, 1, nn))
+    precip_gaussianize = precip_gaussianize.reshape(m, n)
+    sd = np.std(precip_gaussianize)
+    if sd == 0:
+        sd = 1
+
+    precip_gaussianize /= sd
+
+    return precip_gaussianize
+
+
 def downscale(
-    precip, ds_factor, alpha=None, threshold=None, return_alpha=False, smooth=None
+    precip,
+    ds_factor,
+    alpha=None,
+    threshold=None,
+    return_alpha=False,
+    smooth="Gaussian",
+    spectral_fusion=False,
 ):
     """
     Downscale a rainfall field by increasing its spatial resolution by
@@ -166,21 +207,25 @@ def downscale(
 
     """
     type_smooth = _check_smooth_value(smooth)
-
     orig_res = np.shape(precip)[1]
 
-    ki = np.fft.fftfreq(precip.shape[0])
-    kj = np.fft.fftfreq(precip.shape[1])
+    if spectral_fusion:
+        precip_bis = gaussianize(precip)
+    else:
+        precip_bis = precip
+
+    ki = np.fft.fftfreq(precip_bis.shape[0])
+    kj = np.fft.fftfreq(precip_bis.shape[1])
     k_sqr = ki[:, None] ** 2 + kj[None, :] ** 2
     k = np.sqrt(k_sqr)
 
-    ki_ds = np.fft.fftfreq(precip.shape[0] * ds_factor, d=1 / ds_factor)
-    kj_ds = np.fft.fftfreq(precip.shape[1] * ds_factor, d=1 / ds_factor)
+    ki_ds = np.fft.fftfreq(precip_bis.shape[0] * ds_factor, d=1 / ds_factor)
+    kj_ds = np.fft.fftfreq(precip_bis.shape[1] * ds_factor, d=1 / ds_factor)
     k_ds_sqr = ki_ds[:, None] ** 2 + kj_ds[None, :] ** 2
     k_ds = np.sqrt(k_ds_sqr)
 
     if alpha is None:
-        fp = np.fft.fft2(precip)
+        fp = np.fft.fft2(precip_bis)
         fp_abs = abs(fp)
         log_power_spectrum = np.log(fp_abs**2)
         valid = (k != 0) & np.isfinite(log_power_spectrum)
@@ -191,6 +236,21 @@ def downscale(
         warnings.simplefilter("ignore")
         fg *= np.sqrt(k_ds_sqr ** (-alpha / 2))
     fg[0, 0] = 0
+
+    if spectral_fusion:
+        fp = np.fft.fft2(precip_bis)
+        kmax = int(precip_bis.shape[0] / 2)
+        nx = fg.shape[0]
+        Pk0 = spectral.rapsd(precip_bis)[kmax - 1]
+        gk0 = spectral.rapsd(np.fft.ifft2(fg).real / np.fft.ifft2(fg).real.std())[
+            kmax - 1
+        ]
+        fg *= Pk0 / gk0
+        fg[0:kmax, 0:kmax] = fp[0:kmax, 0:kmax]
+        fg[nx - kmax :, 0:kmax] = fp[-kmax:, 0:kmax]
+        fg[0:kmax, nx - kmax :] = fp[0:kmax, -kmax:]
+        fg[nx - kmax :, nx - kmax :] = fp[-kmax:, -kmax:]
+
     g = np.fft.ifft2(fg).real
     g /= g.std()
     r = np.exp(g)
--- a/pysteps/downscaling/rainfarm.py
+++ b/pysteps/downscaling/rainfarm.py
@@ -23,6 +23,7 @@ import numpy as np
 from scipy.ndimage import zoom
 from scipy.signal import convolve
 from pysteps.utils.spectral import rapsd
+from pysteps.utils.dimension import aggregate_fields
 
 
 def _gaussianize(precip):
@@ -71,15 +72,17 @@ def _estimate_alpha(array, k):
     return alpha
 
 
-def _apply_spectral_fusion(array_low, array_high, freq_array_low, freq_array_high, ds_factor):
+def _apply_spectral_fusion(
+    array_low, array_high, freq_array_low, freq_array_high, ds_factor
+):
 
     nax, _ = np.shape(array_low)
     nx, _ = np.shape(array_high)
 
-    nax2 = nax // 2
+    k0 = nax // 2
 
-    array_low_k0 = rapsd(array_low, fft_method=np.fft)[nax2 - 1] * nax ** 2
-    array_high_k0 = rapsd(array_high, fft_method=np.fft)[nax2 - 1] * nx ** 2
+    array_low_k0 = rapsd(array_low, fft_method=np.fft)[k0 - 1] * nax**2
+    array_high_k0 = rapsd(array_high, fft_method=np.fft)[k0 - 1] * nx**2
 
     array_high *= np.sqrt(array_low_k0 / array_high_k0)
 
@@ -87,22 +90,26 @@ def _apply_spectral_fusion(array_low, array_high, freq_array_low, freq_array_hig
     fft_array_high = np.fft.fft2(array_high)
 
     fft_merged = np.zeros((nx, nx), dtype=np.complex128)
-    fft_merged[0:nax2, 0:nax2] = fft_array_low[0:nax2, 0:nax2]
-    fft_merged[nx - nax2:nx, 0:nax2] = fft_array_low[nax2:2*nax2, 0:nax2]
-    fft_merged[0:nax2, nx - nax2:nx] = fft_array_low[0:nax2, nax2:2*nax2]
-    fft_merged[nx - nax2:nx, nx - nax2:nx] = fft_array_low[nax2:2*nax2, nax2:2*nax2]
+    fft_merged[0:k0, 0:k0] = fft_array_low[0:k0, 0:k0]
+    fft_merged[nx - k0:nx, 0:k0] = fft_array_low[k0:2 * k0, 0:k0]
+    fft_merged[0:k0, nx - k0:nx] = fft_array_low[0:k0, k0:2 * k0]
+    fft_merged[nx - k0:nx, nx - k0:nx] = fft_array_low[k0:2 * k0, k0:2 * k0]
 
-    fft_merged[nax2, 0] = np.conj(fft_merged[nx - nax2, 0])
-    fft_merged[0, nax2] = np.conj(fft_merged[0, nx - nax2])
+    fft_merged[k0, 0] = np.conj(fft_merged[nx - k0, 0])
+    fft_merged[0, k0] = np.conj(fft_merged[0, nx - k0])
 
-    freq_i = np.tile(np.fft.fftfreq(array_high.shape[0], d=1 / ds_factor), nx).reshape((nx, nx))
+    freq_i = np.tile(np.fft.fftfreq(array_high.shape[0], d=1 / ds_factor), nx).reshape(
+        (nx, nx)
+    )
     freq_j = freq_i.T
 
     ddx = np.pi * (1 / nax - 1 / nx) / np.abs(freq_i[0, 1] - freq_i[0, 0])
-    fx2 = freq_array_high ** 2
-    fax2 = np.max(freq_array_low) ** 2
+    fx2 = freq_array_high**2
+    fax2 = freq_array_low[k0, k0] ** 2
 
-    fft_merged = fft_array_high * (fx2 > fax2) + fft_merged * (fx2 <= fax2) * np.exp(-1j * ddx * freq_i - 1j * ddx * freq_j)
+    fft_merged = fft_array_high * (fx2 > fax2) + fft_merged * (fx2 <= fax2) * np.exp(
+        -1j * ddx * freq_i - 1j * ddx * freq_j
+    )
 
     merged = np.real(np.fft.ifftn(fft_merged)) / len(fft_merged)
 
@@ -119,7 +126,7 @@ def _compute_kernel_radius(ds_factor):
 def _make_tophat_kernel(ds_factor):
     """Compute 2d uniform (tophat) kernel"""
     radius = _compute_kernel_radius(ds_factor)
-    (mx, my) = np.mgrid[-radius : radius + 0.01, -radius : radius + 0.01]
+    (mx, my) = np.mgrid[-radius:radius + 0.01, -radius:radius + 0.01]
     tophat = ((mx**2 + my**2) <= radius**2).astype(float)
     return tophat / tophat.sum()
 
@@ -235,8 +242,8 @@ def downscale(
 
         noise_field = _apply_spectral_fusion(
             precip_transformed, noise_field, freq_array, freq_array_highres, ds_factor
-        )            
-    else: 
+        )
+    else:
         noise_field /= noise_field.std()
         noise_field = np.exp(noise_field)
 
@@ -248,10 +255,16 @@ def downscale(
             f"available kernels: {list(_make_kernel)}"
         )
 
-    precip_zoomed = zoom(precip, ds_factor, order=1)
-    precip_smoothed = _balanced_spatial_average(precip_zoomed, kernel)
-    noise_smoothed = _balanced_spatial_average(noise_field, kernel)
-    precip_highres = noise_field * precip_smoothed / noise_smoothed
+    noise_lowres = aggregate_fields(noise_field, ds_factor, axis=(0, 1))
+
+    ca = precip / noise_lowres
+
+    cai = np.repeat(np.repeat(ca, ds_factor, axis=0), ds_factor, axis=1)
+
+    precip_highres = noise_field * cai
+
+    if kernel is not None:
+        precip_highres = _balanced_spatial_average(precip_highres, kernel)
 
     if threshold is not None:
         precip_highres[precip_highres < threshold] = 0
--- a/pysteps/downscaling/rainfarm.py
+++ b/pysteps/downscaling/rainfarm.py
@@ -21,6 +21,7 @@ import warnings
 
 import numpy as np
 from scipy.ndimage import convolve
+from scipy.ndimage import zoom
 
 
 def _log_slope(log_k, log_power_spectrum):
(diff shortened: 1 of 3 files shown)


pysteps/downscaling/rainfarm.py
```diff
@@ -44,7 +44,60 @@ def _balanced_spatial_average(x, k):
     return convolve(x, k) / convolve(ones, k)
 
 
-def downscale(precip, ds_factor, alpha=None, threshold=None, return_alpha=False):
+def _smoothconv(P,nas):
```
dnerini, line 47: We are trying to enforce python [naming conventions](https://realpython.com/python-pep8/#naming-styles) for variables names. In this case, instead of `P` we could try to use a more informative name and apply the lower_with_under case. It is a generally a good idea to have a look at the existing code: we used `precip` as variable name for the input field.

dnerini: We are trying to enforce python [naming conventions](https://realpython.com/python-pep8/#naming-styles) for variables names (I know it is still not the case everywhere, but we at least try to do it for new code...). Can you think of more explicit variable names for `P` and `nas`?

dnerini: we are also using the [black](https://github.com/psf/black) autoformatter throughout the code base (indeed, you'll notice that the black test failed).

pysteps/downscaling/rainfarm.py
```diff
@@ -44,7 +44,60 @@ def _balanced_spatial_average(x, k):
...
+
+    """
+    Parameters
```
dnerini: Very nice that you already included docstrings. Please have a look at the `downscale()` method below to see how we typically use blank lines and blank spaces to format the text. Let me know if something is not clear.

pysteps/downscaling/rainfarm.py
```diff
@@ -44,7 +44,60 @@ def _balanced_spatial_average(x, k):
...
+
+    Returns
+    -------
+
+    The smoothened field.
+
+    References
+    ----------
+
+    Terzago et al. 2018
```
dnerini: To make it a clickable reference, you'll need to include it in our [list of references](https://github.com/pySTEPS/pysteps/blob/master/doc/source/references.bib) and then use the ":cite:\`Terzago2018\`" tag. See the docstrings of the `downscale()` method below.

pysteps/downscaling/rainfarm.py
```diff
@@ -44,7 +44,60 @@ def _balanced_spatial_average(x, k):
...
+
+    """
+
+    print("smoothing")
```
dnerini: Perhaps provide a more verbose message? Or simply remove it if it is not that important...
Also, consider adding a `verbose` argument to your method which will silence the print messages if set to False.

pysteps/downscaling/rainfarm.py
```diff
@@ -119,8 +173,18 @@ def downscale(precip, ds_factor, alpha=None, threshold=None, return_alpha=False)
     tophat = ((mx**2 + my**2) <= rad**2).astype(float)
     tophat /= tophat.sum()
 
-    P_agg = _balanced_spatial_average(P_u, tophat)
```
dnerini, line 122: From what I understand, you could keep the calls to the `_balanced_spatial_average` function before the if block and simply call `_smoothconv` if `smooth` is True. This would avoid some code duplication.

pysteps/downscaling/rainfarm.py
```diff
@@ -44,7 +44,60 @@ def _balanced_spatial_average(x, k):
...
+
+    """
+    Parameters
+    ----------
+    
+    P: matrix
+    matrix with the input field to smoothen, with dimensions ns*ns
+    
+    nas : int
+    original size
```
dnerini: why do you need an argument for the original size? Can't you just get it from `P`?

pysteps/downscaling/rainfarm.py
```diff
@@ -44,7 +44,60 @@ def _balanced_spatial_average(x, k):
...
+    pf = np.real(np.fft.ifft2(fm*np.fft.fft2(P))) / np.sum(mask) 
+    if np.sum(indmask) > 0: 
+        P[~ indmask] = 1 
+        pf = pf / (np.real(np.fft.ifft2(fm*np.fft.fft2(P))) / np.sum(P) / len(fm))
+
+    pf[indmask]=np.nan
+    return pf 
+
+
+def downscale(precip, ds_factor, alpha=None, threshold=None, return_alpha=False , smooth=False):
```
dnerini: You forgot to document the argument `smooth` in the docstrings below.

dnerini (commented):
Thanks for this contribution, @simonbeylat !

First of all, I'd be curious to know if you already tried to run the code with some data? Do you see an improvement with the new smoothing operator?

I had a first look a your code and made few minor comments, mostly concerning  "aesthetics". These, however, are important to make the code base readable and usable by others (I like to think that we write code not for machines, but for other humans!). You can find some more details and guidelines in our dedicated developer page:
https://pysteps.readthedocs.io/en/stable/developer_guide/contributors_guidelines.html

You can also find more suggestions to improve your code in the codacy checks included in the test suite of this PR: https://app.codacy.com/gh/pySTEPS/pysteps/pullRequest?prid=10213794

More importantly, I was wondering about the difference between the smoothing operator from @jhardenberg (which you nicely reimplemented in this PR) and the `_balanced_spatial_average` already used in the code. Aren't they doing the same thing? I wouldn't mind replacing the existing method with a more sophisticated one, we just need to make sure that we are not applying the same thing twice.

Finally, your new code needs to be tested (you'll notice that the coverage would decrease after merging, since we are adding untested lines). I suggest that you have a look at the https://github.com/pySTEPS/pysteps/blob/master/pysteps/tests/test_downscaling_rainfarm.py file and try to add a new test that will run rainfarm with and without smoothing.

jhardenberg:
@simonbeylat  as I said, it is great that you are adding this.

> More importantly, I was wondering about the difference between the smoothing operator from @jhardenberg (which you nicely reimplemented in this PR) and the `_balanced_spatial_average` already used in the code. Aren't they doing the same thing? I wouldn't mind replacing the existing method with a more sophisticated one, we just need to make sure that we are not applying the same thing twice.

Indeed `_balanced_spatial_average` is a generic function which does convolution with a kernel (I also was going to suggest some pre-made convolution function from scipy or numpy to this end). So to implement Gaussian smoothing, it would be enough to define a Gaussian kernel instead of the tophat which is currently being used in the code and then call `_balanced_spatial_average`.  What is different in your new  `_smoothconv` function is also the treatment of missing values, which is indeed completely missing in `_balanced_spatial_average` (and which could in theory be added to this function). 

Both a tophat and a Gaussian are valid options and it would be fine if the user could select which one to use. In the original Rainfarm we usually also leave the option of no smoothing at all, just making sure that the aggregattion over boxes corresponding to the coarse resolution is conserved, but that is another issue.

Btw I notice that in your code if the smooth option is selected you call **both** the tophat smoothing and the Gaussian smoothing in sequence. That is a double smoothing, If we use the Gaussian smoothing we do not need to smooth before also with the tophat.

Another comment: in other languages we always implemented smoothing using the  equivalent of the `scipy.ndimage.convolve` option `mode='wrap'` to treat the boundaries because we implemented it dierctly in Fourier space. There is nothing wrong with the default 'reflect' used in  `_balanced_spatial_average` but obviously it will lead to different results.  The new implementation of `_smoothconv` implictly is using 'wrap' since it operates in Fourier space.

To be clear, my suggestion would indeed be to expand `_balanced_spatial_average` to treat also missing values and then call it either with a tophat or with a Gaussian kernel as needed. This is much more compact code. No need to have to functions, one should be enough.
 
A final comment regards speed: as far as I understand it `scipy.ndimage.convolve` which is used in `_balanced_spatial_average` does 'regular' convolution with matrix multiplication. The same done in Fourier space (as we do in other languase and as you are now doing in your new function) can be much more efficient. Actually `scipy.signal.convolve` as of version 0.19.0 chooses automatically the fastest option for you, so it might be better to use that instad of `scipy.ndimage.convolve` in `_balanced_spatial_average`. 

@simonbeylat have you tried comparing the speed of your new function to that of the current `_balanced_spatial_average` ? And with that of a `_balanced_spatial_average` using `scipy.signal.convolve`  ? (should be similar)

pysteps/downscaling/rainfarm.py
simonbeylat: Yes, I changed the names

simonbeylat: I am not familiar with black autoformatting, I will look into it further and try to use it too.

pysteps/downscaling/rainfarm.py
simonbeylat: It was a check when I was testing the code, I removed the printing, which is not useful.

pysteps/downscaling/rainfarm.py
simonbeylat: I get this `nas` on the low resolution field (the field to be reduced) on the `downscale()` method. The `P` used here already has a high resolution.

simonbeylat:
Hello 

First of all, thank you for your feedback! 

@dnerini , I also like the idea of having a clear and aesthetic code that could be easily understood by others. I will make some changes by looking at the different documentations you sent me! I will also use the test code before I add the change. 


> Both a tophat and a Gaussian are valid options and it would be fine if the user could select which one to use. In the original Rainfarm we usually also leave the option of no smoothing at all, just making sure that the aggregattion over boxes corresponding to the coarse resolution is conserved, but that is another issue.

I like @jhardenberg  idea where we let the users choose which technique they want to use. I think I can try to "update" the `_balanced_spatial_average` method to add the treatment of missing values. 

So I suggest changing the smooth argument and having 3 options: 

- `None`: no smooth operatore. 
- `'tophat'` (default): use the `_balanced_spatial_average ` method. 
- `'Gaussian'`: use `_smoothconv` or `_balanced_spatial_average` with a Gaussian kernel.


About testing the code and the speed of time. I tested the code using several data (I was using those data when I was using the R version of RainFARM). I didn't see a noticeable improvement but that may be because I was using `_balanced_spatial_average ` and `_smoothconv ` in a row. In terms of speed, both methods were quite fast. I didn't focus on that point. 



(out of context, I wanted to introduce myself quickly (as I have never done it before) I have just graduated in a master of data science/engineering. (which ended with an internship in downscaling and bais correction of climate models), and I am actively looking for a PhD in these (or similar) subjects. This development is also a way for me to learn and improve my skills. I wanted to thank you for the help and the feed back you are giving me.)

pysteps/downscaling/rainfarm.py
dnerini: very good, should be as easy as running (in your virtual environment)
```
conda install -c conda-forge black
black .
```
(note that the pysteps_dev environment already includes black)

dnerini:
Excellent, sounds all very good to me! I'll have another closer look once you push the next changes ;-)

And thanks for introducing yourself, always a pleasure to learn more about the person behind a simple name!

dnerini:
Thanks @simonbeylat for the new changes, well done!

I noticed that you still have both smoothing methods implemented: `_smoothconv` and `_balanced_spatial_average`. From what I gathered in our discussion above with @jhardenberg, we would rather join them into one single method which can accept any arbitrary kernel. In practice, we suggested that you start with the existing `_balanced_spatial_average` and expand it with the treatment of missing values (as in the new function). Then, based on a input argument `kernel`, you could build and pass the corresponding kernel function or just skip the smoothing part kernel is set to None.

Does it make sense?

simonbeylat:
Hi @dnerini, @jhardenberg 

I managed to add the 2nd pull request which is about the part added by D'onofrio et al in 2014 on spectral fusion. It took me a while but I found something that looks good. 

I look forward to your reviews.

dnerini:
hey @simonbeylat just wanted to quickly mention that I haven't forgot about this, it's just a bit of a busy period. I hope to be able to look into your latest changes soon!

simonbeylat:
Hello @dnerini , 

Will you have the time to review the new commit ? 

Or should I add the last commit (with the computation of weight) ?

dnerini:
hey @simonbeylat I just created a PR on your branch: https://github.com/simonbeylat/pysteps/pull/1 
let's continue discussing there for now, we'll come back here once those changes are discussed and merged.

simonbeylat:
I had the changes you made and I had small changes because it wasn't working. 

I was implementing the `_apply_spectral_fusion` function, I tried to translate @jhardenberg 's [julia version](https://github.com/jhardenberg/RainFARM.jl/blob/master/src/rf/mergespec_spaceonly.jl).

But I don't really understand how this version and the version written in D'onofrio's article are equivalent. Also, when I tried the code, it seems to be wrong and gives very large extremes. 

The first version I wrote (you should be able to see it by looking at this [commit](https://github.com/simonbeylat/pysteps/commit/388ba4560598153ec4125fc929c49a899e9d6c0e) )seemed to give good results and I tried to close to the mathematics written in the article, but I might be wrong. 

I think I misunderstood something and translated @jhardenberg's code wrong. 
I'll focus on this part.

dnerini:
Hello @simonbeylat thanks for the update, happy to see that we are making some progress on this!

Concerning the `_apply_spectral_fusion`, it would be perfectly fine for me to follow the maths from the original paper if these give more sensible results. We might then ask for help to @jhardenberg to reimplement his julia/R versions in python in a second step.

dnerini:
hey @simonbeylat, getting back on this PR I realized I might have misunderstood your last message -- are you still working on it or is it in principle complete?

simonbeylat:
Hello, sorry I have several important projects and a big problem with my personal computer where I was working on this project... (as I was doing this in my spare time).
I will work on this project as soon as I can !

dnerini:
Hey @simonbeylat , absolutely, no worries! Just wanted to make sure that you weren't waiting for a feedback from my side ;-)

simonbeylat:
Hi @dnerini, I'm very busy as I started my PhD and so I started many projects, I'm not sure if I can finish the development of rainfarm at the moment. I will try to do it when I have some free time, but I can't know when. 

I'm sorry for that.

ecasellas:
Hello @dnerini, @jhardenberg and @simonbeylat,

Firstly, thank you very much for providing this tool. We began working with the RainFARM implementation in PySTEPS, and upon discovering this pull request, we decided to try the spectral fusion update by Simon, as it represents an improvement in merging between coarse-scale precipitation fields and fine synthetic precipitation fields. The code we used is in this [commit](https://github.com/simonbeylat/pysteps/commit/388ba4560598153ec4125fc929c49a899e9d6c0e).

The appearance of the field seems to be more realistic compared to the current RainFARM implementation in PySTEPS. However, we have encountered some unrealistic precipitation values at certain grid boundaries. Attached is an example comparing the different RainFARM implementations: From left to right, Spectral fusion in Python using Simon's code (note the unrealistic precipitation at the boundary on the right), the original RainFARM in PySTEPS, and R's implementation of RainFARM.

rainfarm_example


We checked Simon's code, but so far, we haven’t identified anything that could lead to this problem. During the development of the code, did you encounter similar results or problems? We consulted resources online and found suggestions that this issue may be related to spectral leakage. However, we are not experts in this field. One proposed solution is to apply windowing methodologies before calculating the Fast Fourier Transform, although we have found that this does not always resolve the problem. We are not sure if there are additional insights or solutions we may be overlooking.

Thank you once again.

simonbeylat:
Hello @ecasellas,

I'd like to start by apologising for the lack of progress, 
I started updating RainFARM in my free time during my master's degree, but when I started my phd (on data assimilation), I didn't have time to get back to it. 

Actually the implementation was not completely finished, the initial plan was to come back with the same version of RainFARM as the Terzago et al. 2018 paper. 
but I'm quite surprised by the results (even with the boundary issue). Unfortunately I can't manage to finish this job (although I would have loved to)..

I'll be very happy to follow developments if you decide to update the code I've added. 

Best, 

Simon

dnerini:
Dear @ecasellas many thanks for following up on this! This PR is currently stale, which is a bit of a pity since it's probably not very far from being completed. Since the original author is not able to work on it anymore, I suppose this is now open to anyone motivated to continue and finalize the work!

Concerning the actual issue reported above, the windowing approach that you mention makes sense to me. Another thing worth trying might be padding the field with zeros to mitigate such boundary effects, have you already tried something in this direction? 

edit: just out of curiosity, could you add a plot of the original low res field that you downscaled in your example?

ecasellas:
Dear @simonbeylat and @dnerini,

Thank you very much for your comments. We're attaching a new figure with the original low-resolution field. We've tried the downscaling procedure with different events, and the problem seems to persist, especially at the right and bottom boundaries, as you can see in the new example.

MicrosoftTeams-image (1)

MicrosoftTeams-image

As you suggested, we tried padding the field with zeros (we applied it before the FFT calculation, we hope we did it in the correct place), but it didn’t improve. Perhaps we didn’t apply it correctly. We have already experimented with some windowing approaches, but sometimes they work (completely removing the boundary effects), and sometimes they yield very implausible results, appearing very granular.

We'll take another look into the code and return here as soon as we have made progress.

Just to introduce ourselves a little bit, I'm working on this together with Blai Domingo, who is doing an internship at the Meteorological Service of Catalonia, and we'd like to continue with this pull request, although we're not sure if we're going to succeed.

Thank you once again.

simonbeylat:
Dear @ecasellas, 

Thank you for the updates, it would be great if you could continue! 

Here are the problems linked to these pr :https://github.com/pySTEPS/pysteps/issues/287. 
If you want more context, I think there's the initial plan. and I apologise again for not having completed the code correctly. 

Best,

Simon

ecasellas:
Dear @dnerini,

Together with @blaiDoAr, we made some progress and believe we now have a Python version equivalent to R's RainFARM implementation. The rare bands previously reported seem to have disappeared, and the results appear to match those obtained directly with R, as you can see in the figure below. (from left to right: original low res field, PySTEPS implementation, RainFARM R's implmentation with smoothing set to None, and Simon's implementation with changes and without smoothing)

MicrosoftTeams-image (3)

Our starting point was the last commit of this PR, from December 6th, and we made some changes using a translation to Python from some of R's implementation. The version we implemented is not completely adapted to the current version of PySTEPS, since, if we didn’t misunderstand R's implementation, the latter works with wavenumbers and PySTEPS with frequencies. Perhaps we should make some changes to unify both versions and attempt to modify the least of the current PySTEPS implementation.

Should we commit the current changes to this PR as a starting point for the possible new version, or should we proceed differently?

Thank you

dnerini:
Hi @ecasellas and @blaiDoAr 
Exciting news and congratulations on the great achievement!

So you can either decide to commit your changes to this PR, which is a fork that belongs to @simonbeylat (not sure you'll have the right permissions though, unless you already got invited to the associated repo), or decide to "fork the fork" so to have full control on the new version. In this scenario, you'll have to open a new PR to pysteps and we'll close this one.

Does this help? I look forward to include your changes (including the excellent work from @simonbeylat ) into the main repo! Let me know if you have any other question.

simonbeylat:
Hi @ecasellas and @blaiDoAr

I've invited you on the fork, and I'll give you all the rights. I think that keeping the PR can help keep all the information in one place (and therefore make it easier for others to follow).
But it's up to you to decide what you want to do to continue the work.
Very good Job, 
Thank to you two ! 

Best, 

Simon

ecasellas:
Hi @dnerini and @simonbeylat,

Thank you very much for your comments. Since @simonbeylat granted us push permissions to the associated repository, we'll commit the changes to this PR. As Simon mentioned, we also think it's better to keep everything in the same place and to maintain the history of changes and all the great work done by Simon.

We'll return here once the changes are pushed.

Thank you once again.

ecasellas:
Hello @dnerini and @simonbeylat,

We've made some updates to this PR, which include:

- Implementation of spectral fusion, translated from R's RainFARM version. Initially, we used a custom function for radially averaged spectra in Python, but later discovered that the `raspd` function was already implemented in PySTEPS, as @simonbeylat had done.
- Adjustment of the consistency condition between the synthetic field and the original field. In the previous version, the original field was zoomed into the synthetic field shape before applying the consistency condition. Here, we aggregate the synthetic field to the original shape, interpolate it to match the synthetic field shape (using `np.kron`), and finally apply the consistency condition.
- Added an option to skip smoothing. When aggregating a high-resolution field, it matches the original field if no smoothing kernels (already implemented by @simonbeylat) are applied. This is not the case if any of the available smoothing kernels are used.
- If `spectral_fusion` set to `True`, the noise field is normalized using the shape of the field rather than its standard deviation. This is the same as in R, we tried using standard deviation, but then the results differ.
- The `alpha` used in generating the synthetic field is divided by 2, as in the referenced articles. However, in R and Julia implementations, it’s calculated as `(alpha + 1) / 4`. Dividing `alpha` by 2 yields results consistent with R, although we're uncertain why this discrepancy exists.
- Tests for precipitation high-resolution aggregation and spectral fusion. The latter was integrated into the existing `test_rainfarm_shape` function, with additional parameters included.

We appreciate the excellent work done by @simonbeylat and @dnerini, as it’s much easier to approach a problem when it’s already laid out.

We look forward to your reviews, thank you once again.

RAINFARM

jhardenberg:
Hi folks, this is a great development, congrats to everybody!

ecasellas:
Hello,

We tried this new version on several episodes and found that something might not be entirely correct in the changes we made. Upon reviewing the code, we realized that we misunderstood the calculation of the synthetic field amplitude. In the papers, `alpha` is divided by 2, and in our previous implementation, we used:

`noise_field_complex = white_noise_field_complex * np.sqrt(
    freq_array_highres**-alpha / 2
)`

However, the factor of 1/2 is already accounted for by `np.sqrt`. So adding `1/2` was resulting in a double `np.sqrt`. In the new commit, we removed the "`/ 2`". We then checked other precipitation events and compared the performance of Python RainFARM and R RainFARM, and the results were equivalent.

P.S. Sorry @jhardenberg for not mentioning you in the previous message, but we also thank you very much for providing RainFARM in R and Julia.

dnerini:
hi @ecasellas, thanks for the important fix! From my perspective, this looks good to go. I wonder if @simonbeylat and @jhardenberg are planning to review this PR?

edit: @ecasellas the formatting test is failing, can you run black on your code and push the changes to this branch?

ecasellas:
Hello @dnerini, black formatter run on rainfarm.py, hope it'll pass the test!

pysteps/downscaling/rainfarm.py
```diff
@@ -20,38 +20,211 @@
...
+
+    # Fuse in the frequency domain
+    mask_high = freq_squared_high > freq_squared_low_center
+    mask_low = ~mask_high
+    fft_merged = fft_high * mask_high + fft_merged * mask_low * np.exp(
+        -1j * ddx * freq_i - 1j * ddx * freq_j
+    )
+
+    # Inverse FFT to obtain the merged array in the spatial domain
+    merged = np.real(np.fft.ifftn(fft_merged)) / len(fft_merged)
```
dnerini: shouldn't it be `fft_merged.size` instead? `len(fft_merged)` will return the length of the first dimension of that array or `fft_merged.size[0]`, not the total number of elements in it

dnerini:
hello @ecasellas apologies for the slow action on this PR... I had a closer look today and pushed mostly cosmetic changes to make the code (hopefully) more readable. I also left one question above if you could look at it? once resolved that conversation I think we'll be ready to merge

pysteps/downscaling/rainfarm.py
ecasellas: yes, completely true, we missed this when translating from R to Python because `length` in R doesn't have the same meaning as `len` in Python. Still, we reran the code with this change, and the result doesn't seem to change that much. We'll upload a new commit with this change.

dnerini (approved):

ecasellas:
hi @dnerini, the code is far more readable now! Thank you very much for your reviews and changes, and for the opportunity to contribute to PySTEPS!

dnerini:
Thank you for completing the work of @simonbeylat and providing a great contribution to pysteps! OSS spirit at its best! 

I'll need to look into a problem with the docs that are currently failing to compile, but I hope to be able to release a new version with your work soon!

PR #322, opened 2023-03-13
Fix l2 parameter in SAL verification #321

### LINKED ISSUE #321: location parameter L2 in the SAL verification
In the function _sal_l2_param in pysteps.verification.salscores, obs_r and forc_r is obtained by multiplying output from _sal_weighted_distance with the meanvalue of model field. Why this multiplication? It makes it possible for l2 to exceed 1, which is different from what is said in the docstring. The source article, referred to as WPHF2008 in the docstring does not multiply with the model field mean value when they calculate obs_r and forc_r.

Best,

Matilda

EsmailGhaemi:
Remove the multiplication factor in l2 parameter, see #321

### DIFF
--- a/pysteps/verification/salscores.py
+++ b/pysteps/verification/salscores.py
@@ -302,10 +302,10 @@ def _sal_l2_param(prediction, observation, thr_factor, thr_quantile, tstorm_kwar
     )
     obs_r = (
         _sal_weighted_distance(observation, thr_factor, thr_quantile, tstorm_kwargs)
-    ) * (np.nanmean(observation))
+    )
     forc_r = (
         _sal_weighted_distance(prediction, thr_factor, thr_quantile, tstorm_kwargs)
-    ) * (np.nanmean(prediction))
+    ) 
     location_2 = 2 * ((abs(obs_r - forc_r)) / maximum_distance)
     return float(location_2)
 


pysteps/verification/salscores.py
```diff
@@ -302,10 +302,10 @@ def _sal_l2_param(prediction, observation, thr_factor, thr_quantile, tstorm_kwar
     )
     obs_r = (
```
dnerini, line 303: please simplify the syntax by removing the unnecessary brackets and keeping it on one line, ie
```
obs_r = _sal_weighted_distance(observation, thr_factor, thr_quantile, tstorm_kwargs)
```

pysteps/verification/salscores.py
```diff
@@ -302,10 +302,10 @@ def _sal_l2_param(prediction, observation, thr_factor, thr_quantile, tstorm_kwar
...
         _sal_weighted_distance(observation, thr_factor, thr_quantile, tstorm_kwargs)
-    ) * (np.nanmean(observation))
+    )
     forc_r = (
         _sal_weighted_distance(prediction, thr_factor, thr_quantile, tstorm_kwargs)
-    ) * (np.nanmean(prediction))
+    )
```
dnerini, line 308: same comment as above, note that the extra blank character at line 308 is causing the code formatting check to fail -> remember to run `black .` before committing your changes (see section "Using Black auto-formatter" in our [Contributor guidelines](https://pysteps.readthedocs.io/en/stable/developer_guide/contributors_guidelines.html#code-style)).

dnerini (changes requested):
Thanks again @EsmailGhaemi for the PR, need just to fix the minor issues with the code formatting before we can merge this into master

pysteps/verification/salscores.py
EsmailGhaemi: It is done!

dnerini:
duplicate of #323

PR #326, opened 2023-04-22
Update references

dnerini:
Update references with recent publication from @RubenImhoff. 

Note that this will have to be soon updated with volume, number, pages, and year.


RubenImhoff (approved):
Thanks for adding this, Daniele! Looks good to go, I'll add the volume and issue numbers, etc., once I have them.

PR #327, opened 2023-04-22
Fixes following pandas 2

dnerini:
Minor adjustments following the release of pandas 2:

* in `tdating.py` use `pd.concat` instead of `pd.DataFrame().append` which was effectively removed ([ref](https://pandas.pydata.org/docs/whatsnew/v2.0.0.html#deprecations)).
* in `salscores.py`: Division by zero returns `-inf`, `nan`, or `inf` depending on the numerator, instead of raising (but I wasn't able to understand exactly the root cause of this breaking change...)

Additionally,
* Fix scipy.ndimage's DeprecationWarnings 
* Fix pandas' SettingWithCopyWarnings

### DIFF
--- a/pysteps/tracking/tdating.py
+++ b/pysteps/tracking/tdating.py
@@ -311,7 +311,7 @@ def couple_track(cell_list, max_ID, mintrack):
         for t in range(len(cell_list)):
             mytime = cell_list[t]
             mycell = mytime[mytime.ID == n]
-            cell_track = cell_track.append(mycell)
+            cell_track = pd.concat([cell_track, mycell], ignore_index=True)
 
         if len(cell_track) < mintrack:
             continue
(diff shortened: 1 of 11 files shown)


pysteps/tracking/tdating.py
```diff
@@ -211,12 +211,12 @@ def tracking(cells_id, cells_id_prev, labels, V1, max_ID):
                 size.append(len(x))
             biggest = np.argmax(size)
             new_ID = new_ID[biggest]
-            cells_id_new.ID.iloc[ID] = new_ID
+            cells_id_new.loc[index, "ID"] = new_ID
```
dnerini, line 214: using `.loc[index, "ID"]` instead of `.ID.iloc[ID]` avoids the warning by removing the chaining operation. In this case, `.loc` should be more suited than `.iloc` since `index`refers to the index label and not the index position (although the two are in fact the same value...).

PR #328, opened 2023-04-25
Update blending references

RubenImhoff:


RubenImhoff:
No volume and issue number so far. Who knows, maybe it will be added later. But if not, this should be the full reference.

dnerini (approved):

PR #330, opened 2023-06-08
Add missing "bias" method to the interface of the verification module

Joep1999:
The BIAS method was not available with the 'get method' function, because it was missing from the name list. Therefore I added it.


dnerini (approved):
Hi @Joep1999, very good, many thanks for the valuable contribution!

PR #340, opened 2023-12-19
Geogrid fix

dbkhout:
While only used for visualization, the `get_geogrid`  and `reproject_geodata` functions seem to be slightly off.

The functions use `np.linspace` to retrieve the lower boundary for each pixel based on the number of pixels and the extent of the image (e.g. `x1`, `x2`). In doing so, `x2` should be excluded from the interval as it represents the upper boundary of the image.

___
Illustration:

`np.linspace(3, 10, 7)` returns `3.,  4.16666667,  5.33333333,  6.5,  7.66666667, 8.83333333, 10.` while `3., 4., 5., 6., 7., 8., 9.` is expected. 

`np.linspace(3, 10, 7, endpoint=False)` does return the expected result.

___

Note that this has been accounted for correctly (yet with different approaches) in other pysteps functions:

https://github.com/pySTEPS/pysteps/blob/8dca14afbf74f47072a810317a8bddefa33e3e60/pysteps/utils/dimension.py#L410

https://github.com/pySTEPS/pysteps/blob/8dca14afbf74f47072a810317a8bddefa33e3e60/pysteps/io/exporters.py#L336

### DIFF
--- a/pysteps/visualization/utils.py
+++ b/pysteps/visualization/utils.py
@@ -237,8 +237,14 @@ def reproject_geodata(geodata, t_proj4str, return_grid=None):
     # Reproject grid on fall-back projection
     if return_grid is not None:
         if return_grid == "coords":
-            y_coord = np.linspace(y1, y2, shape[0]) + geodata["ypixelsize"] / 2.0
-            x_coord = np.linspace(x1, x2, shape[1]) + geodata["xpixelsize"] / 2.0
+            y_coord = (
+                np.linspace(y1, y2, shape[0], endpoint=False)
+                + geodata["ypixelsize"] / 2.0
+            )
+            x_coord = (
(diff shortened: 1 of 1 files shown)


dnerini:
hi @dbkhout thanks for the PR! Tests are failing for macos and windows, but it's probably unrelated from your changes. I'll have a closer look asap and will let you know

dnerini:
@dbkhout could you please merge the latest version of pysteps/master in your branch? This should fix the errors in the CI pipelines

dnerini (approved):
Looking good and all tests are green now (we can ignore the coverage in this case)! Thanks again for the nice fix :-)

dbkhout:
You're welcome, thanks for the quick reaction!

PR #345, opened 2023-12-22
Return total scaled volume in SAL metric

### LINKED ISSUE #344: Structure parameter in SAL verification metric
When calculating scaled volume (V(R)), it should be proportional to the second moment of the precipitation field, not the first.

EsmailGhaemi:
Fixed #344 (S parameter edited, L parameter modified for dealing with nan values)

Now the scaled volume function directly returns total scaled volume (instead of scaled voulme for each object). Additionally, for the weighted distance function, np.nansum() is used to avoid the effect of nan values.

### DIFF
--- a/pysteps/verification/salscores.py
+++ b/pysteps/verification/salscores.py
@@ -360,10 +360,10 @@ def _sal_detect_objects(precip, thr_factor, thr_quantile, tstorm_kwargs):
         }
     _, labels = tstorm_detect.detection(precip, **tstorm_kwargs)
     labels = labels.astype(int)
-    precip_objects = pd.DataFrame(regionprops_table(
-        labels, intensity_image=precip, properties=REGIONPROPS
-    ))
-    return (precip_objects)
+    precip_objects = pd.DataFrame(
+        regionprops_table(labels, intensity_image=precip, properties=REGIONPROPS)
+    )
+    return precip_objects
 
 
 def _sal_scaled_volume(precip_objects):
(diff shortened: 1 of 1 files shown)


pysteps/verification/salscores.py
```diff
@@ -155,11 +155,11 @@ def sal_structure(
     observation_objects = _sal_detect_objects(
         observation, thr_factor, thr_quantile, tstorm_kwargs
     )
-    prediction_volume = _sal_scaled_volume(prediction_objects).sum()
-    observation_volume = _sal_scaled_volume(observation_objects).sum()
+    prediction_volume = sal_scaled_volume(prediction_objects)
```
dnerini: why did you rename the calls to `_sal_scaled_volume`? this is causing the NameErrors in the tests, can you revert this change?

pysteps/verification/salscores.py
```diff
@@ -155,11 +155,11 @@ def sal_structure(
...
+    observation_volume = sal_scaled_volume(observation_objects)
     nom = prediction_volume - observation_volume
     denom = prediction_volume + observation_volume
-    return np.divide(nom, (0.5 * denom))
+    return nom / (0.5 * denom)
```
dnerini: I changed this to `np.divide()` in #327 because we want this function to raise an error instead of returning `Inf`, any particular reason to change it back?

dnerini (changes requested):
Hi @EsmailGhaemi! Very nice seeing you back here :-) Thanks a lot for the fix! I have few questions on some changes you introduced.

Tests are currently failing because of NameErrors (see my comment below). Let me know if you need any help on this.

pysteps/verification/salscores.py
EsmailGhaemi: Hi @dnerini, thank you.
You are right. I've applied black and edited these lines.
Merry Christmas :)

dnerini:
Thanks @EsmailGhaemi for the changes. Could you just spend few more words on what you did? It's not obvious to me what you changed, why, and what kind of impact we should expect. This will help to describe the changes in the release notes.

EsmailGhaemi:
Yes, of course.
In the previous version, the scaled volume was calculated for each object separately, and then in sal_srtucture function we summed all of them. However, when calculating total volume (V_R in equation 8, Wernli et al., 2008), we should multiply each object's scaled volume (v_n) to the amount of precipitation for that object (R_n) as well.
Now the scaled volume function directly returns total scaled volume (instead of scaled voulme for each object). Additionally, for the weighted distance function, np.nansum() is used to avoid the effect of nan values.

dnerini:
Brilliant, thanks, I've update the description above.

dnerini (approved):

Issue #346, opened 2024-02-06
No reproducibility of steps blended nowcast when using noise_stddev_adj='auto'


mpvginde:
When running 2 identical blended nowcasts with `pysteps.blending.steps.forecast` using  
a fixed  `seed`  
and setting  
`noise_stddev_adj = 'auto'`,  
the results (sometimes) differ.

I suspect this behavior is coming from the 
`pysteps.noise.utils.compute_noise_stddev_adjs` function.

This function takes the `seed` argument, but inside the function the `seed` is always set to `None`:
https://github.com/pySTEPS/pysteps/blob/be8eea432239f77bdfb7edf51a9bfaa94d012752/pysteps/noise/utils.py#L102-L106

Leading to different normalization weights for every run and impacting the final results, preventing reproducibility.

Kind regards,
@mpvginde

dnerini:
Hi @mpvginde thanks for reporting this issue! Do you want to go ahead and submit a pr (looks like a simple fix)? Otherwise I'll try to have a look on it

mpvginde:
Sure @dnerini. I'll have a look at it.

mpvginde:
A small update:
The issue might be more complex as I initially thought.
When I fix passing the seed in the `pysteps.noise.utils.compute_noise_stddev_adjs` results become reproducable for the `pysteps.blending.steps.forecast` routine.
However, for `pysteps.nowcast.steps.forecast` results still differ between identical runs. 
Only when I disable the dask parallel workers in the `_update` helper routine in `nowcasts.steps` the results become reproducable. 
I suspect this might be linked to issue #337, I will investigate further.

mpvginde:
See https://github.com/pySTEPS/pysteps/issues/337#issuecomment-1935702000

Should be fixed with PR #347

Issue #337, opened 2023-08-17
fft_method pyfftw causes unexpected noise additions when using multi-threading


RubenImhoff:
If I run an ensemble using multiple cores (so, dask will parallellize the ensemble members over the cores), it seems that the ensemble order is lost, resulting in weird transitions (due to noise from a different member that ends up in that member). See for instance (these are three 15-min instance in the forecast):
![2023-08-17 15_21_02-P_fc in 202308170600_blended_nowcast_olddask](https://github.com/pySTEPS/pysteps/assets/31476760/88e5a5f3-edbb-49a1-bbfb-219191fe3ac4)

![2023-08-17 15_21_19-P_fc in 202308170600_blended_nowcast_olddask](https://github.com/pySTEPS/pysteps/assets/31476760/52e43872-c4e1-4abd-bb53-79084f0cc088)

![2023-08-17 15_21_30-P_fc in 202308170600_blended_nowcast_olddask](https://github.com/pySTEPS/pysteps/assets/31476760/f38046ad-e5a6-442a-8d89-0fd20d998f24)

Or maybe even clearer:
![MicrosoftTeams-image](https://github.com/pySTEPS/pysteps/assets/31476760/ac936688-cca6-4309-b38e-60416e0e4a2a)

If I run on 1 core, this problem does not occur, so it gives the impression that this has to do with parallellizing using dask. After having contact with @mpvginde, I could not reproduce the error with the Belgian test case that we have (the figures above are with our Dutch data and setup). Only difference between our setups turned out to be the setting that 'fft_method' = pyfftw instead of numpy in the Belgian setup. After changing this, the problems disappeared when running multi-threaded. This gives the impression that pyfftw should not be used for nowcasting and blended forecasting when using more than 1 worker/thread. 
Is this a familiar issue to you and is there anything we can do about it?

RubenImhoff:
Update on this possible bug: I see that it still comes back in the STEPS nowcasts (not the blending anymore) with both pyfftw and numpy as fft_methods. Seems that in the newer Dask versions, we have to more explicitly pin the activities to specific cores and keep track of it. For the nowcast loop this still goes well multi-threaded, provided that numpy is used as method, but for all other multi-threaded processes (cascade decomposition, noise initialization, etc.) it seems to go wrong. We should either fix those options to one worker or find a different solution, I'm afraid.

dnerini:
[...]

PR #347, opened 2024-02-09
Bugfix reproducibility & ensemble member order with dask

mpvginde:
This PR (hopefully) fixes 2 issues:
#337: Strange artifacts showing up in the STEPS nowcasting due to the random order of the ensemble members in the numpy array when using dask multi-threading

#346: No reproducibility  of the (blended) STEPS nowcast results even when a fixed seed argument is given

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -217,7 +217,7 @@ latex_preamble = r"""
 latex_elements = {
     "papersize": "a4paper",
     "pointsize": "10pt",
-    "preamble": latex_preamble
+    "preamble": latex_preamble,
     # Latex figure (float) alignment
     #
     # 'figure_align': 'htbp',
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -514,6 +514,7 @@ def forecast(
     )
 
     # 2. Initialize the noise method
+    np.random.seed(seed)
     pp, generate_noise, noise_std_coeffs = _init_noise(
         precip,
         precip_thr,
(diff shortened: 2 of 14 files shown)


dnerini:
Nice job @mpvginde! This looks fantastic and I'm curious to hear from @RubenImhoff about #337.

I was wondering, should we add an explicit test on reproducibility?

RubenImhoff:
Great work, @mpvginde! I will have a look at the PR later this week. :)

RubenImhoff (commented):
Hi @mpvginde, great work! This indeed solves the issue. I know that the same issues occurs when using the LINDA nowcasting model. Would a similar fix work there, too?

dnerini:
Hi @mpvginde should we try to push your fixes quickly to master and release them? Do you need any support of the issues that are still open or you think this is now ready to be merged?

mpvginde:
Hi @dnerini, 
I noticed last week that in some cases, the result was also dependent on the number of dask workers used.
This should be fixed now with the last commit that I added.
I think this can be merged now.
I'm not sure why the Black check is failing though, If somebody could help me with that.

I did notice that the method used for generating the random numbers is using some deprecated numpy-functions.
If I find some more time I will try to update the generation of random number to the latest best practices.
But for now this fix should be ok.

dnerini:
> I think this can be merged now.

Brilliant! Just wondering if we should include the fix for the `linda` method too, as mentioned by @RubenImhoff?

> I'm not sure why the Black check is failing though, If somebody could help me with that.

maybe you're using a older version of black locally? during tests we are using the latest available version

dnerini:
alright, I'm merging this, thanks a lot @mpvginde for the nice contribution!

PR #350, opened 2024-03-14
Bug fix: ignore 0 in TDaTing cell match

ritvje:
This PR fixes a bug in the `pysteps.tracking.tdating.match` function. As far as I can tell, in this bug the label 0 was included in possible match IDs, even though 0 denotes pixels that are not part of any cell. So there could be cases, where the advected cell would have e.g. 55% of 0 and 45% of some cell and it would not be matched to that cell. So this fix will first check if the advected cell contains only 0-labeled pixels, it is initiated as a new track. If not, then 0 is removed from the possible ID labels and the matching is done as previously.


dnerini:
sorry for taking so long, I think we can merge this bugfix if @feldmann-m has no objections to this?

dnerini (approved):

Issue #332, opened 2023-06-28
STEPS-blending: Losing small scale features in the first time steps and other issues


mpvginde:
Hi everyone,

In setting up pySTEPS for a blended nowcast with STEPS, I have noticed that during the first time steps some of the radar image small-scale features are immediately lost. 
This becomes particularly apparent when the blended STEPS nowcast is compared to the non-blended version of the STEPS nowcast.

An example over the Belgian radar domain can be seen here:
![blendVSnoblend_t0](https://github.com/pySTEPS/pysteps/assets/81358394/640f8d06-8650-4e37-b019-2308915a2e19)
*Fig. 1: Identical starting conditions (i.e. radar field) for the blenden nowcast (left) and non-blended nowcast (right)*

![blendVSnoblend_t1](https://github.com/pySTEPS/pysteps/assets/81358394/68c4b338-22cd-4c1d-9b94-b8bad6da2c06)
*Fig. 2: The blended (left) and non-blended (right) nowcast after 1 time step of 5 minutes*

It's clear that even after the first time step (Fig, 2), the blended nowcast field has lost some fine-scale structures w.r.t. the non-blended nowcast (i.e. the blended fields seem smoother). In addition, there also seem to be more broad structures with more intense rainfall in the blended nowcast.

When investigating this phenomenon I came across several issues/points of attention in the blending code.\
I already discussed some of them with @RubenImhoff, but I will try to explain them here too so everything is well documented.
\
\
In order to explore the problem I plotted the time evolution of the Power (displayed as standard deviation [St. Dev.]) of the different cascade levels for the different fields.\
For blending nowcast there are 2 relevant fields per level:
- the precipitation field
- the noise field

Both are evolved separately using an AR(2) model\
In addition I also plotted the time evolution of their respective blending weights with which they are combined. I have deliberately omitted the NWP-component as its weight is typically very small during the first time steps. 

For the non-blending nowcast there is only 1 relevant field as the noise is directly incorporated in the AR(2) process.

Finally, also included is a figure of the actual fields themselves at a specific time step.
For the blended nowcast:
- the AR(2)-evolved precipitation field
- the AR(2)-evolved noise field
- both fields combined using their respective blending weights

For the non-blended nowcast:
- the AR(2) with noise term-evolved precipitation field

![overview_l6_t1](https://github.com/pySTEPS/pysteps/assets/81358394/dae25e84-957b-4c6b-bca4-abac51cb0889)
[...]

Issue #309, opened 2022-09-13
How to handle no rain in pysteps?


ladc:
Dear developers,

As mentioned in #282 and #295, the STEPS nowcasting component of pysteps (with/without blending with NWP) cannot really handle the case where there is no rain in the radar image / QPE. The problems are not only technical but also scientific: how to determine optical flow and AR/scaling parameters from a field of zeroes? And subsequently, how to extrapolate/blend?
We see 2 possible problem scenarios:

- QPE is zero and NWP is nonzero
  Here, as a workaround, one could use the "past" NWP forecasts as if they were observed QPE fields, and compute optical flow and other parameters on these fields. One could do a blending, but it's not clear how, as the AR parameters of the QPE are undefined.
- QPE is zero and NWP is zero:
  This is a bigger problem since there's simply no rain to calculate any optical flow or noise parameters. One could immediately use the NWP instead of blending, but that does not yet give us a way to generate variability in the ensemble forecast. A possible solution is to use climatological values (or values of the past X days) to determine the noise parameters. That way, you can generate some noise in the ensemble members (to generate variability when it actually starts to rain in the NWP forecast). 

 What are your thoughts on this?

Cheers,

Michiel and Lesley

dnerini:
Dear Michiel and Lesley

Many thanks for starting this important discussion.
(let's ping @pulkkins @RubenImhoff @loforest directly here too)

For radar only nowcasts, the problem is somewhat simpler. Some routines will have no problems in dealing with all zeros: e.g. the LK optical flow will simply return a field of zero displacement. Others, the steps nowcasts for example, will fail (as noted in #282). 

I guess that in the past we have mostly dealt with this before calling the main nowcasting routine: if no precipitation is observed by radar in the recent past, do not run the nowcast as it will be all zeros anyway. I acknowledge that it would be better to deal with such cases within the nowcasting method itself (if input is all zeros, return a forecast with all zeros), or at least raise a clear error message. Clearly, this starts by adding tests cases for nowcasting with input without precipitation.

For blending this gets more complicated, as you nicely summarized above. Maybe we should ask directly Alan? Surely he had to find a solution for Australia ...

RubenImhoff:
Hi Michiel, Lesley and Daniele,

Very relevant points that we come across often in operational systems..

[...]

PR #351, opened 2024-03-27
Hackathon July 2023

mpvginde:
This pull request brings all developments made during the July 2023 pySTEPS-hackathon at the RMIB.  
Many thanks to all contributors: @ladc, @RubenImhoff, Simon De Kock and Ricardo  Reinoso-Rondinel.

The work was focussed on the STEPS blending routine.

**NEW FEATURES**
1. Scientificaly correct handling of following cases when using STEPS blending (linked with #309)
- No rain in radar fields | No rain in NWP fields
- No rain in rader fiels | Rain in NWP fields

**REFACTORING** (all linked with #332)
1. Initialisation of the noise cascade with random noise (with correct spatial correlation) to have correct power 
2. Renormalize radar extrapolation cascade after AR(2)-evolution to not lose power
3. STEPS Blending extrapolation component:
- Precipiation and noise cascade are recomposed before advection
- Both fields are advected
- They are decomposed again for blending with NWP cascade
This removes the loss of power in the smallest scales due to the interpolation needed for the semilagrangian advection.  
In addition computational cost is reduced because the number of extrapolation is greatly reduced.

**BUGFIXES**
- Decoding issue for loading older (RMI) ODIM HDF5-files
- Avoid accidentally masking precip arrays by passing a copy to spatial_correlation

### DIFF
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -970,7 +970,6 @@ def forecast(
                         precip_cascade[j][i] = autoregression.iterate_ar_model(
                             precip_cascade[j][i], PHI[i, :]
                         )
-
                     else:
                         # use the deterministic AR(p) model computed above if
                         # perturbations are disabled
@@ -2223,8 +2222,8 @@ def _compute_initial_nwp_skill(
     """Calculate the initial skill of the (NWP) model forecasts at t=0."""
     rho_nwp_models = [
         blending.skill_scores.spatial_correlation(
-            obs=R_c[0, :, -1, :, :],
-            mod=precip_models[n_model, :, :, :],
+            obs=R_c[0, :, -1, :, :].copy(),
+            mod=precip_models[n_model, :, :, :].copy(),
             domain_mask=domain_mask,
         )
         for n_model in range(precip_models.shape[0])
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -869,6 +869,8 @@ def forecast(
                 )
 
             # the nowcast iteration for each ensemble member
+            R_f_ = [None for _ in range(n_ens_members)]
+
             def worker(j):
                 # 8.1.2 Determine the skill of the nwp components for lead time (t0 + t)
                 # Then for the model components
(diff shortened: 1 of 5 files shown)


pysteps/blending/steps.py
```diff
@@ -560,336 +570,659 @@ def forecast(
...
 
-    # 4. Estimate AR parameters for the radar rainfall field
-    PHI = _estimate_ar_parameters_radar(
-        precip_cascade, ar_order, n_cascade_levels, MASK_thr
-    )
+    # 2.3 Check for zero input fields in the radar and NWP data.
+    zero_precip_radar = blending.utils.check_norain(precip, precip_thr, norain_thr)
+    # The norain fraction threshold used for nwp is the default value of 0.0,
+    # since nwp does not suffer from clutter.
+    zero_model_fields = blending.utils.check_norain(precip_models_pm, precip_thr)
```
RubenImhoff: I have noticed that there can be some artifacts when there is very little rainfall, so I would suggest to use the same norain_thr for the NWP model fields.

RubenImhoff (commented):
Great to see the PR, @mpvginde! :)

dnerini:
some of the tests were failing for the same reason as in #352 , @mpvginde can you please merge master into this branch to see if that solves the issue?

there are however many other failing tests, are you looking into these?

pysteps/blending/steps.py
```diff
@@ -560,336 +570,659 @@ def forecast(
...
+            n_cascade_levels,
+            bp_filter,
+            decompositor,
+            fft,
+            noise_method,
+            noise_kwargs,
+            noise_stddev_adj,
+            measure_time,
+            num_workers,
+            seed,
```
dnerini, line 681: @mpvginde I added the missing `seed` parameter here, do we also need to set the seed with numpy before the call to `_init_noise` as done elsewhere in this file?

RubenImhoff:
I made some changes which should (hopefully) fix the issues that came from the tests. @ladc, which tests did you still want to add? We might be good to go after that.

RubenImhoff:
The tests on Ubuntu fail due to a problem with uploading the Codecov report - not sure what is causing it, but it seems unrelated to the PR.

dnerini:
yes I agree, we can ignore the failing test in the context of this PR. If @ladc has no further comments or changes to add, I think we can merge it!

dnerini (approved):

pysteps/blending/steps.py
dnerini: was this resolved in the meantime?

RubenImhoff: Yes, I added it in my commit.

RubenImhoff (approved):

mpvginde:
Thanks everyone for helping with this PR!

PR #368, opened 2024-07-03
Fix: make sure that precip_models variable has the correct length

RubenImhoff:
Fix to ensure that the `_check_inputs` function in `/blending/steps.py` looks for the correct `precip_models` length when a list of timesteps instead of an integer is provided.

### DIFF
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -1808,7 +1808,7 @@ def _check_inputs(
     if isinstance(timesteps, list) and not sorted(timesteps) == timesteps:
         raise ValueError("timesteps is not in ascending order")
     if isinstance(timesteps, list):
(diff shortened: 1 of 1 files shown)


dnerini (approved):
Looking good!

ladc:
Thanks! I haven't used custom timesteps so far, so just to check -I thought it was allowed to have non-integer timesteps if you specify custom ones, so what happens in that case? (maybe requires a ceil?)

RubenImhoff:
That is a good point, @ladc! I have not tested that (yet), but I can imagine that a ceil would solve that potential issue. Should we make sure that the number gets rounded up in that case to ensure that the non-integer timestep always falls within the total forecast length?

dnerini:
should this be added to the hackathon?

PR #369, opened 2024-07-03
fix: make sure netcdf exporters can handle list of timesteps

RubenImhoff:
This PR ensures that the netcdf exporters can handle a list of timesteps next to just an integers number of timesteps. One of pysteps' functionalities is that you can give it specific output timesteps for which the nowcasts should be stored, but it was not yet possible to store these output timesteps as such in a netCDF. This PR tries to fix that.

### DIFF
--- a/pysteps/io/exporters.py
+++ b/pysteps/io/exporters.py
@@ -592,7 +592,7 @@ def initialize_forecast_exporter_netcdf(
     var_time = ncf.createVariable("time", int, dimensions=("time",))
     if isinstance(n_timesteps, list):
         if incremental != "timestep":
-            var_time[:] = n_timesteps * timestep * 60
+            var_time[:] = np.array(n_timesteps) * timestep * 60
     else:
         if incremental != "timestep":
             var_time[:] = [i * timestep * 60 for i in range(1, n_timesteps + 1)]
--- a/pysteps/io/exporters.py
+++ b/pysteps/io/exporters.py
@@ -461,7 +461,8 @@ def initialize_forecast_exporter_netcdf(
             + "'timestep' or 'member'"
         )
 
-    if isinstance(n_timesteps, list):
+    n_timesteps_is_list = isinstance(n_timesteps, list)
+    if n_timesteps_is_list:
         num_timesteps = len(n_timesteps)
     else:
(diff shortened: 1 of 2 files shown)


pysteps/io/exporters.py
```diff
@@ -584,8 +590,12 @@ def initialize_forecast_exporter_netcdf(
         var_ens_num.units = ""
 
     var_time = ncf.createVariable("time", int, dimensions=("time",))
-    if incremental != "timestep":
-        var_time[:] = [i * timestep * 60 for i in range(1, n_timesteps + 1)]
+    if isinstance(n_timesteps, list):
+        if incremental != "timestep":
```
dnerini: I suggest to move `if incremental != "timestep":` to the upper level of this nested if, this way we'd need it only once

pysteps/io/exporters.py
```diff
@@ -635,7 +645,8 @@ def initialize_forecast_exporter_netcdf(
     exporter["timestep"] = timestep
     exporter["metadata"] = metadata
     exporter["incremental"] = incremental
-    exporter["num_timesteps"] = n_timesteps
+    exporter["num_timesteps"] = num_timesteps
+    exporter["timesteps"] = n_timesteps
```
dnerini, line 649: don't we need to build the list of timesteps in case `n_timesteps` is not a list?

pysteps/io/exporters.py
```diff
@@ -460,8 +461,13 @@ def initialize_forecast_exporter_netcdf(
             + "'timestep' or 'member'"
         )
 
+    if isinstance(n_timesteps, list):
```
dnerini: this condition is checked multiple times later in the code, consider using a variable `n_timesteps_is_list = isinstance(n_timesteps, list)`

dnerini (changes requested):
Great contribution @RubenImhoff ! I left few very minor comments

pysteps/io/exporters.py
RubenImhoff: It does a check later on whether this variable contains a list or an integer, so that should be covered.

dnerini (approved):

PR #379, opened 2024-07-16
Add option to smooth radar mask

### LINKED ISSUE #377: Create smooth transition between cells within and outside the radar domain during blending

sidekock:
Solution for #377 

Addition done by adding a cv2 based method in blending.utils.py called compute_smooth_dilated_mask(...).

This function can be used in the blending.steps.forecast by changing the create_smooth_radar_mask parameter to True, standard this is put to False to not create conflicts with existing code.

### DIFF
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -90,7 +90,7 @@ def forecast(
     conditional=False,
     probmatching_method="cdf",
     mask_method="incremental",
-    create_smooth_radar_mask=False,
+    smooth_radar_mask_range=0,
     callback=None,
     return_output=True,
     seed=None,
@@ -211,9 +211,10 @@ def forecast(
       'obs' = apply precip_thr to the most recently observed precipitation intensity
       field, 'incremental' = iteratively buffer the mask with a certain rate
       (currently it is 1 km/min), None=no masking.
-    create_smooth_radar_mask: bool
-      If set to True, this generates a smooth mask for the radar that removes
-      hard edges between NWP and radar in the final blended product.
+    smooth_radar_mask_range: int, Default  is 0.
+      If 0 this generates a normal forecast. To create a smooth mask, this range
+      should be a positive value. The smooth radar mask for the radar removes
+      the hard edges between NWP and radar in the final blended product.
     probmatching_method: {'cdf','mean',None}, optional
       Method for matching the statistics of the forecast field with those of
       the most recently observed one. 'cdf'=map the forecast CDF to the observed
@@ -1455,12 +1456,13 @@ def forecast(
                         # forecast outside the radar domain. Therefore, fill these
                         # areas with the "..._mod_only" blended forecasts, consisting
                         # of the NWP and noise components.
-                        if create_smooth_radar_mask is True:
+
+                        nan_indices = np.isnan(R_f_new)
+                        if smooth_radar_mask_range != 0:
                             # Compute the smooth dilated mask
-                            nan_indices = np.isnan(R_f_new)
                             new_mask = blending.utils.compute_smooth_dilated_mask(
                                 nan_indices,
-                                100,
+                                max_padding_size_in_px=smooth_radar_mask_range,
                                 gaussian_kernel_size=9,
                                 inverted=False,
                                 non_linear_growth_kernel_sizes=False,
@@ -1484,22 +1486,28 @@ def forecast(
                                 ],
                                 axis=0,
                             )
-                            # Finally, fill the remaining nan values, if present, with
-                            # the minimum value in the forecast
-                            R_f_new[np.isnan(R_f_new)] = np.nanmin(R_f_new)
+
+                            nan_indices = np.isnan(R_pm_blended)
+                            R_pm_blended = np.nansum(
+                                [
+                                    R_pm_blended * mask_radar,
+                                    R_pm_blended_mod_only * mask_model,
+                                ],
+                                axis=0,
+                            )
                         else:
-                            nan_indices = np.isnan(R_f_new)
                             R_f_new[nan_indices] = R_f_new_mod_only[nan_indices]
                             nan_indices = np.isnan(R_pm_blended)
                             R_pm_blended[nan_indices] = R_pm_blended_mod_only[
                                 nan_indices
                             ]
-                            # Finally, fill the remaining nan values, if present, with
-                            # the minimum value in the forecast
-                            nan_indices = np.isnan(R_f_new)
-                            R_f_new[nan_indices] = np.nanmin(R_f_new)
-                            nan_indices = np.isnan(R_pm_blended)
-                            R_pm_blended[nan_indices] = np.nanmin(R_pm_blended)
+
+                        # Finally, fill the remaining nan values, if present, with
+                        # the minimum value in the forecast
+                        nan_indices = np.isnan(R_f_new)
+                        R_f_new[nan_indices] = np.nanmin(R_f_new)
+                        nan_indices = np.isnan(R_pm_blended)
+                        R_pm_blended[nan_indices] = np.nanmin(R_pm_blended)
 
                         # 8.7.2. Apply the masking and prob. matching
                         if mask_method is not None:
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -211,11 +211,13 @@ def forecast(
       'obs' = apply precip_thr to the most recently observed precipitation intensity
       field, 'incremental' = iteratively buffer the mask with a certain rate
       (currently it is 1 km/min), None=no masking.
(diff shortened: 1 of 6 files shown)


pysteps/blending/steps.py
```diff
@@ -1451,16 +1455,51 @@ def worker(j):
...
-                        R_f_new[nan_indices] = R_f_new_mod_only[nan_indices]
-                        nan_indices = np.isnan(R_pm_blended)
-                        R_pm_blended[nan_indices] = R_pm_blended_mod_only[nan_indices]
-                        # Finally, fill the remaining nan values, if present, with
-                        # the minimum value in the forecast
-                        nan_indices = np.isnan(R_f_new)
-                        R_f_new[nan_indices] = np.nanmin(R_f_new)
-                        nan_indices = np.isnan(R_pm_blended)
-                        R_pm_blended[nan_indices] = np.nanmin(R_pm_blended)
+                        if create_smooth_radar_mask is True:
```
dnerini:
```suggestion
                        if create_smooth_radar_mask:
```

pysteps/blending/steps.py
```diff
@@ -1451,16 +1455,51 @@ def worker(j):
...
+                            # Compute the smooth dilated mask
+                            nan_indices = np.isnan(R_f_new)
```
dnerini: I would not replicate in the if..else.. block any line that is common to both cases, as for example the this line which could go before the if statement

pysteps/blending/steps.py
```diff
@@ -1451,16 +1455,51 @@ def worker(j):
...
+                            new_mask = blending.utils.compute_smooth_dilated_mask(
+                                nan_indices,
+                                100,
+                                gaussian_kernel_size=9,
+                                inverted=False,
+                                non_linear_growth_kernel_sizes=False,
+                            )
+
+                            # Ensure mask values are between 0 and 1
```
dnerini, line 1471: this comment is a bit off, isn't it?

pysteps/blending/steps.py
```diff
@@ -90,6 +90,7 @@ def forecast(
     conditional=False,
     probmatching_method="cdf",
     mask_method="incremental",
+    create_smooth_radar_mask=False,
```
dnerini: can we add a test case setting this flag as True?

dnerini (changes requested):

sidekock:
@RubenImhoff  suggested to add some figures of the effect of the smooth mask on the final product. Below I attach a figure with 3 version: one without a smooth edge (current product), one with a smooth edge with a maximum range of 50 px (=50km) and one with a max range of 80km.
Screenshot from 2024-07-17 16-41-04

pysteps/blending/steps.py
sidekock: This is fixed in the new version

pysteps/blending/steps.py
sidekock: This is fixed in the new version

pysteps/blending/steps.py
sidekock: This is fixed in the new version

pysteps/blending/steps.py
sidekock: Some tests still need to be written.

pysteps/blending/steps.py
```diff
@@ -210,6 +211,10 @@ def forecast(
       'obs' = apply precip_thr to the most recently observed precipitation intensity
       field, 'incremental' = iteratively buffer the mask with a certain rate
       (currently it is 1 km/min), None=no masking.
+    smooth_radar_mask_range: int, Default  is 0.
+      If 0 this generates a normal forecast. To create a smooth mask, this range
```
dnerini: Add somewhere that the range is in units of pixels

pysteps/blending/steps.py
```diff
@@ -1451,10 +1456,52 @@ def worker(j):
                         # forecast outside the radar domain. Therefore, fill these
                         # areas with the "..._mod_only" blended forecasts, consisting
                         # of the NWP and noise components.
+
                         nan_indices = np.isnan(R_f_new)
-                        R_f_new[nan_indices] = R_f_new_mod_only[nan_indices]
-                        nan_indices = np.isnan(R_pm_blended)
-                        R_pm_blended[nan_indices] = R_pm_blended_mod_only[nan_indices]
+                        if smooth_radar_mask_range != 0:
```
dnerini, line 1464: What happens if the value is negative?

pysteps/blending/steps.py
```diff
@@ -1451,10 +1456,52 @@ def worker(j):
...
+                            # Compute the smooth dilated mask
+                            new_mask = blending.utils.compute_smooth_dilated_mask(
+                                nan_indices,
+                                max_padding_size_in_px=smooth_radar_mask_range,
+                                gaussian_kernel_size=9,
+                                inverted=False,
+                                non_linear_growth_kernel_sizes=False,
+                            )
+
+                            # Ensure mask values are between 0 and 1
```
dnerini, line 1471: To ensure that values are between 0 and 1 you would have to do something like np.clip

pysteps/blending/steps.py
```diff
@@ -1451,10 +1456,52 @@ def worker(j):
...
                         nan_indices = np.isnan(R_f_new)
-                        R_f_new[nan_indices] = R_f_new_mod_only[nan_indices]
-                        nan_indices = np.isnan(R_pm_blended)
-                        R_pm_blended[nan_indices] = R_pm_blended_mod_only[nan_indices]
+                        if smooth_radar_mask_range != 0:
+                            # Compute the smooth dilated mask
+                            new_mask = blending.utils.compute_smooth_dilated_mask(
+                                nan_indices,
+                                max_padding_size_in_px=smooth_radar_mask_range,
+                                gaussian_kernel_size=9,
```
dnerini: So this value doesn't need to be changed by the user?

dnerini (commented):
Looking good! I have few more minor comments. Some simple testing and then I think this I will be ready to go!

pysteps/blending/steps.py
sidekock: Which value are you talking about here? The kernel size or de range? In the case of the first, it can be changed but should be experimented with according to the domain and radar observation. In the case of nr 2, the method should provide this as a parameter.

pysteps/blending/steps.py
sidekock: I added this to the new version

pysteps/blending/steps.py
sidekock: A  raise ValueError("max_padding_size_in_px must be greater than or equal to 0.") is thrown in the bledning.util.compute_smooth_dilated_mask( ... ) function

pysteps/blending/steps.py
sidekock: I added this to the new version

pysteps/blending/steps.py
dnerini: I was referring to `gaussian_kernel_size` which is hardocoded

RubenImhoff: Just discussed with @sidekock, the `gaussian_kernel_size` ensures that the nan-fields are large enough to start the smoothing. Without it, you would also apply this to local nan-values in the radar domain (if these occur - which is the case for the Belgian data). The fixed value does the trick. It is the default in the `blending.utils.py` functionality, so it does not need to be repeated here (hence, `gaussian_kernel_size = 9` can be removed here in my opinion).

pysteps/blending/utils.py
```diff
@@ -557,3 +565,86 @@ def check_norain(precip_arr, precip_thr=None, norain_thr=0.0):
...
+    inverted : bool, optional
+        If True, invert the original mask before processing. Default is False.
+    non_linear_growth_kernel_sizes : bool, optional
+        If True, use non-linear growth for kernel sizes. Default is False.
+
+    Returns
+    -------
+    initial_mask : array_like
+        The initial binary mask after Gaussian blur and thresholding.
+    new_mask : array_like
```
RubenImhoff: It returns final_mask instead of the two variables mentioned above. Are they the same or different?

RubenImhoff (changes requested):

pysteps/blending/steps.py
dnerini: OK thanks @RubenImhoff , I agree, let's remove the hardcoded value and rely on the default

RubenImhoff:
I think we're almost there - I'm working on adding some extra tests. :)

RubenImhoff (approved):
I think all issues have been solved now and we have sufficient test coverage. We can merge this PR now. Nice contribution! :)

dnerini (approved):
Well done!

Issue #353, opened 2024-04-23
Add Kalman-filter based post-processing to blending scheme


RubenImhoff:
As discussed at the start of our blending implementation and re-discussed during this year's EGU, we can still improve the post-processing that takes place in the blending procedure. Right now, the post-processing procedures from the nowcasting module are used and adjusted to work within the blending module. This is functional, but has as limitation that the post-processing is based on (1) having everything in Lagrangian coordinates and that is not the case in the blending procedure (we have incorporated that difference, but is suboptimal), and (2) it depends on the weights assigned to the NWP and extrapolation component, which could reduce extremes. 

A Kalman-filter like method, such as in Nerini et al. (2019; https://doi.org/10.1175/MWR-D-18-0258.1), could do the trick. I think we should see if we can implement this in the blending module anytime soon. :)

dnerini:
Thanks for initiating this @RubenImhoff ! Can you provide any hint on where we should make changes in the current code? I can then try to make a first draft.

RubenImhoff:
Sorry for completely forgetting to respond to this, @dnerini! I will dive into the code and come back to your question early next week. Would be great to put this feature into the code. :)

RubenImhoff:
Hi @dnerini, 
To finally come back to your question:

This is where the probability matching is called in the code: [https://github.com/pySTEPS/pysteps/blob/953f799f932760dd5775ce4f9f0fc771e4d0a67d/pysteps/blending/steps.py#L1494](url).

The CDF probmatching method still calls this original function: [https://github.com/pySTEPS/pysteps/blob/953f799f932760dd5775ce4f9f0fc771e4d0a67d/pysteps/postprocessing/probmatching.py#L54](url)

And maybe most important, as the fields are not in Lagrangian coordinates (whereas this is the case in the nowcasting module until the probability matching part), we perform the post-processing on a blended extrapolation-NWP field (that is, excluding the noise cascade) to get some sort of indication of where the rainfall fields would be at that timestep and what they would look like given the original input files and no added noise: [https://github.com/pySTEPS/pysteps/blob/953f799f932760dd5775ce4f9f0fc771e4d0a67d/pysteps/blending/steps.py#L1431](url). 

Is this something you can work with? Let me know how I can assist, happy to help!

RubenImhoff:
[...]

Issue #370, opened 2024-07-11
Create deterministic blended "control member" for visualisation purposes


ladc:
As mentioned in issue #333, there's currently no way to do a "control" or deterministic run (unperturbed) for the blended nowcast. However, this would be useful (or even necessary) for the visualisation in the app. The ensemble mean is too "blurry" and a single ensemble member might be too distorted.

dnerini:
Would this be a simple extrapolation with linear blending?

mrtnrey:
I find this a very relevant feature request. In fact, I'd like to add the idea of a "worst case" and "least case" ensemble member. I know each of the ensemble members is equally likely, but some of them contain more precipitation than others. I'd like to know whether you could order the members by severeness and if it is then possilbe to show the two extremes to your users (together with the "control member").

dnerini:
>  I'd like to add the idea of a "worst case" and "least case" ensemble member. 

Some time ago, I looked into this idea of "band depth", which goes in the direction of ranking multi-dimensional data and therefore could be relevant for your use case. https://doi.org/10.1198/jasa.2009.0108

There is even an implementation in pysteps, ready to be used and experimented with: https://pysteps.readthedocs.io/en/stable/generated/pysteps.postprocessing.ensemblestats.banddepth.html 

This, however, would not solve the problem of having a "control" member, or did I get your comment wrong?

mrtnrey:
Great! Is there also a notion of "median" of the band then, and could such a median then be used as "control member"?

dnerini:
Yes and in fact the lowest band depth corresponds to the median, while larger values indicate more and more outliers (in both negative and positive senses).

Not sure this corresponds to a control run, strictly speaking, but it could be something worth exploring.

Another option could be to look at the total accumulations for the whole forecast so to consider also the advection speed as a risk factor (typically, the faster the rain cell moves, the lower the total accumulation at a given point on the ground). (By the way, do you perturb the motion field in your setup @ladc ?)

It's a very interesting problem, how to present the ensemble information to the public!

ladc:
To (attempt to) answer your questions:

> Would this be a simple extrapolation with linear blending?

Possibly, but the question is what the weights would be. One could take the weights of the second cascade level as we do to blend the velocity fields. Then still the question remains how to deal with the weight of the third, non-existing noise cascade. I would propose to use the NWP and radar weights as-is (don't try to renormalize them or anything) and just blend them without noise. 

[...]

PR #390, opened 2024-07-18
Add resample distributions function for probability matching

RubenImhoff:
First attempt to add a resample distributions function for probability matching, see also issue #353.

- [x] Add resampling function to `postprocessing/probmatching.py`.
- [x] Use the resampled distribution instead of the `R_pm_blended` variable in `blending/steps.py`.
- [x] If `probmatching_method ` is `cdf` or `mean` in `blending/steps.py`, add resampling prior to using the probability matching as an optional functionality.
- [x] Add tests.

### DIFF
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -1549,6 +1549,7 @@ def forecast(
                         # both the extrapolation cascade and the model (NWP) cascade and use
                         # that for the probability matching
                         if probmatching_method is not None and resample_distribution:
+                            # resample weights are based on cascade level 2
                             R_pm_resampled = probmatching.resample_distributions(
                                 first_array=R_pm_ep[t_index],
                                 second_array=precip_models_pm_temp[j],
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -1550,11 +1550,9 @@ def forecast(
                         # that for the probability matching
                         if probmatching_method is not None and resample_distribution:
                             R_pm_resampled = probmatching.resample_distributions(
-                                a=R_pm_ep[t_index],
-                                b=precip_models_pm_temp[j],
-                                weight=weights_pm_normalized[
-                                    0
-                                ],  # Use the weight from the extrapolation cascade here
+                                extrapolation_cascade=R_pm_ep[t_index],
+                                model_cascade=precip_models_pm_temp[j],
+                                weight_extrapolation=weights_pm_normalized[0],
                             )
                         else:
                             R_pm_resampled = R_pm_blended.copy()
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -53,7 +53,7 @@ from pysteps import extrapolation
 from pysteps import noise
 from pysteps import utils
 from pysteps.nowcasts import utils as nowcast_utils
-from pysteps.postprocessing import probmatching, resample_distributions
+from pysteps.postprocessing import probmatching
 from pysteps.timeseries import autoregression, correlation
 from pysteps import blending
 
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -1549,10 +1549,15 @@ def forecast(
                         # both the extrapolation cascade and the model (NWP) cascade and use
                         # that for the probability matching
                         if probmatching_method is not None and resample_distribution:
-                            # resample weights are based on cascade level 2
+                            # deal with missing values
+                            arr1 = R_pm_ep[t_index]
+                            arr2 = precip_models_pm_temp[j]
+                            arr2 = np.where(np.isnan(arr2), np.nanmin(arr2), arr2)
+                            arr1 = np.where(np.isnan(arr1), arr2, arr1)
+                            # resample weights based on cascade level 2
                             R_pm_resampled = probmatching.resample_distributions(
-                                first_array=R_pm_ep[t_index],
-                                second_array=precip_models_pm_temp[j],
+                                first_array=arr1,
+                                second_array=arr2,
                                 probability_first_array=weights_pm_normalized[0],
                             )
                         else:
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -90,6 +90,7 @@ def forecast(
     conditional=False,
     probmatching_method="cdf",
     mask_method="incremental",
+    resample_distribution=False,
     smooth_radar_mask_range=0,
     callback=None,
     return_output=True,
@@ -205,25 +206,32 @@ def forecast(
       If set to True, compute the statistics of the precipitation field
       conditionally by excluding pixels where the values are below the threshold
       precip_thr.
-    mask_method: {'obs','incremental',None}, optional
-      The method to use for masking no precipitation areas in the forecast field.
-      The masked pixels are set to the minimum value of the observations.
-      'obs' = apply precip_thr to the most recently observed precipitation intensity
-      field, 'incremental' = iteratively buffer the mask with a certain rate
-      (currently it is 1 km/min), None=no masking.
-    smooth_radar_mask_range: int, Default is 0.
-      Method to smooth the transition between the radar-NWP-noise blend and the NWP-noise
-      blend near the edge of the radar domain (radar mask), where the radar data is either
-      not present anymore or is not reliable. If set to 0 (grid cells), this generates a normal forecast without smoothing. To create a smooth mask, this range
-      should be a positive value, representing a buffer band of a number of pixels
-      by which the mask is cropped and smoothed. The smooth radar mask removes
-      the hard edges between NWP and radar in the final blended product. Typically, a value between 50 and 100 km can be used. 80 km generally gives good results.
     probmatching_method: {'cdf','mean',None}, optional
       Method for matching the statistics of the forecast field with those of
       the most recently observed one. 'cdf'=map the forecast CDF to the observed
       one, 'mean'=adjust only the conditional mean value of the forecast field
       in precipitation areas, None=no matching applied. Using 'mean' requires
       that mask_method is not None.
+    mask_method: {'obs','incremental',None}, optional
+      The method to use for masking no precipitation areas in the forecast field.
+      The masked pixels are set to the minimum value of the observations.
+      'obs' = apply precip_thr to the most recently observed precipitation intensity
+      field, 'incremental' = iteratively buffer the mask with a certain rate
+      (currently it is 1 km/min), None=no masking.
+    resample_distribution: bool, optional
+        Method to resample the distribution from the extrapolation and NWP cascade as input
+        for the probability matching. Not resampling these distributions may lead to losing
+        some extremes when the weight of both the extrapolation and NWP cascade is similar.
+        Defaults to False.
+    smooth_radar_mask_range: int, Default is 0.
+      Method to smooth the transition between the radar-NWP-noise blend and the NWP-noise
+      blend near the edge of the radar domain (radar mask), where the radar data is either
+      not present anymore or is not reliable. If set to 0 (grid cells), this generates a
+      normal forecast without smoothing. To create a smooth mask, this range should be a
+      positive value, representing a buffer band of a number of pixels by which the mask
+      is cropped and smoothed. The smooth radar mask removes the hard edges between NWP
+      and radar in the final blended product. Typically, a value between 50 and 100 km
+      can be used. 80 km generally gives good results.
     callback: function, optional
       Optional function that is called after computation of each time step of
       the nowcast. The function takes one argument: a three-dimensional array
@@ -1404,7 +1412,6 @@ def forecast(
                         # latest extrapolated radar rainfall field blended with the
                         # nwp model(s) rainfall forecast fields as 'benchmark'.
 
-                        # TODO: Check probability matching method
                         # 8.7.1 first blend the extrapolated rainfall field (the field
                         # that is only used for post-processing steps) with the NWP
                         # rainfall forecast for this time step using the weights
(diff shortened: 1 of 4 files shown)


pysteps/blending/steps.py
```diff
@@ -205,25 +206,32 @@ def forecast(
...
       The method to use for masking no precipitation areas in the forecast field.
       The masked pixels are set to the minimum value of the observations.
       'obs' = apply precip_thr to the most recently observed precipitation intensity
       field, 'incremental' = iteratively buffer the mask with a certain rate
       (currently it is 1 km/min), None=no masking.
+    resample_distribution: bool, optional
+        Method to resample the distribution from the extrapolation and NWP cascade as input
+        for the probability matching. Not resampling these distributions may lead to losing
+        some extremes when the weight of both the extrapolation and NWP cascade is similar.
+        Defaults to False.
```
dnerini: why default should be False?

pysteps/blending/steps.py
```diff
@@ -1538,19 +1545,35 @@ def worker(j):
...
+                        # If probmatching_method is not None, resample the distribution from
+                        # both the extrapolation cascade and the model (NWP) cascade and use
+                        # that for the probability matching
+                        if probmatching_method is not None and resample_distribution:
+                            R_pm_resampled = probmatching.resample_distributions(
+                                a=R_pm_ep[t_index],
+                                b=precip_models_pm_temp[j],
+                                weight=weights_pm_normalized[
+                                    0
+                                ],  # Use the weight from the extrapolation cascade here
```
dnerini: should we specify somewhere that the weights are based on level 2?

dnerini (commented):
looking good! do we have any example result? is this working?

pysteps/blending/steps.py
RubenImhoff: Good question, True is also fine with me. I was just thinking of not breaking any results for our users, but in principle, this functionality should be the way to go.

pysteps/blending/steps.py
RubenImhoff: Good call, I will add it!

RubenImhoff:
Apologies for the ugly plots, but this was the quick and dirty way to test three versions:

This is a blended forecast up to 12 hours ahead (1 member) for 16 July 2024 in the Netherlands. The left column shows a forecast without the resampling and also no smoothing as introduced by @sidekock yesterday. The middle column does both the smoothing and resampling prior to the probability matching and the right column is a check with only smoothing and no resampling. Overall, I would say that the smoothing already makes a differences on the edges of the radar domain (@sidekock gave a better example in his PR), but the resampling clearly results in less suppression of the extremes in the NWP (you kind of see a dissapation for the longer lead times in the left and right columns, but not in the middle one). I think this is what we are looking for, although one example is of course a bit limited.

resampling_test

pysteps/blending/steps.py
dnerini: I don't think it's a breaking change (the code won't stop work) provided that the results won't dramatically get worst. If we believe this is the way to go, I would rather label this as an improvement and as a user I'll be happy to get it out of the box without having to know much about it.

pysteps/postprocessing/probmatching.py
```diff
@@ -259,6 +260,49 @@ def _get_error(scale):
     return shift, scale, R.reshape(shape)
 
 
+def resample_distributions(extrapolation_cascade, model_cascade, weight_extrapolation):
```
dnerini: mmmh I'm not sure about the new argument names, this method is rather agnostic with respect to what arrays are passed as long as they have the same shape, don't you think?

RubenImhoff: True, but just a and b made it very unclear what the function is doing and what it needs as inputs from the `steps.py` code. What would you suggest?

dnerini: in the steps code you should leave the same variable names, but here I would keep it very general, something like`first_array` and `second_array` would do, no?

pysteps/postprocessing/probmatching.py
```diff
@@ -259,6 +260,49 @@ def _get_error(scale):
...
+    to effectively combine two distributions for the probability matching without losing extremes.
+    Parameters
+    ----------
+    extrapolation_cascade: array_like
+        One of the two arrays from which the distribution should be samples (e.g. the extrapolation
+        cascade).
+    model_cascade: array_like
+        One of the two arrays from which the distribution should be samples (e.g. the NWP (model)
+        cascade).
+    weight_extrapolation: float
```
dnerini: we should make clear that this specifies the probability of sampling the first array

```suggestion
    probability_first_array: float
```

pysteps/postprocessing/probmatching.py
RubenImhoff: That could work, though the method is very tailored to blending at the moment. Happy to change it otherwise.

pysteps/postprocessing/probmatching.py
RubenImhoff: Thanks! Added it in the commit below :)

RubenImhoff:
I think the PR is ready to go, but definitely feel free to give it a last thorough review. :)

pysteps/postprocessing/probmatching.py
```diff
@@ -259,6 +260,57 @@ def _get_error(scale):
...
+    csort: array_like
+        The output distribution as first_array drawn binomial distribution from the input arrays first_array and second_array.
+    """
+    # Make sure that the probability is never: <0, >1 or nan
+    if probability_first_array < 0.0:
+        probability_first_array = 0.0
+    elif probability_first_array > 1.0:
+        probability_first_array = 1.0
+    elif np.isnan(probability_first_array):
+        probability_first_array = 0.0
```
dnerini: why this? if it can happen that the level 2 cascade weight is nan, this special case should be handled in `blending/steps.py` imho

pysteps/postprocessing/probmatching.py
```diff
@@ -259,6 +260,57 @@ def _get_error(scale):
...
+
+    # First make sure there are no nans in the input data
+    # Where the extrapolation cascade is nan (outside the radar domain), we fill it up with the model data
```
dnerini: here to, I think this should be handled in `blending/steps.py`

dnerini (commented):
I added some changes directly to the code and left few comments that I think it'd be better to clarify before we merge

pysteps/postprocessing/probmatching.py
RubenImhoff: Your change to `probability_first_array = np.clip(probability_first_array, 0.0, 1.0)` is neat, better to do it that way. The nan was an extra check, but is (I think) not necessary anymore, as nans should not take place in the blending code (but weights above or below 1 and 0 are in theory possible).

RubenImhoff:
Looks good like this to me, @dnerini, thanks for the help!

pysteps/postprocessing/probmatching.py
dnerini: Is this part necessary @RubenImhoff ?

RubenImhoff: Not necessarily this way, but we should account for nans in the 2D input arrays. My suggestion would be that if there is a nan at a location in one of the two input arrays, you automatically sample the non-nan value in the array that contains a value at that cell. A cleaner way to account for this would be good, I think.

dnerini: OK I see, a bit the problem is that this function is global in that it doesn't compare values at locations, but only considers their ranks. (that is, the arrays are sorted before being resampled). the  logic you describe is in my opinion something different that is specific to steps blending and therefore it should be done before calling this function. Or am I misunderstanding something?

RubenImhoff: That is certainly true, although nans will almost always occur with the datasets that are used for pysteps. So, the question is if we would like to account for it, or if we just want to leave this up to the user and thus put it in the `blending/steps.py` code (fine with me!).

dnerini: OK please have a loot at [a46e62a](https://github.com/pySTEPS/pysteps/pull/390/commits/a46e62a58aa0af64e7421bc8d273f0696974f389) and let me know if this works for you

RubenImhoff: Great, thanks, Daniele! This looks good to me.

dnerini:
@ladc @RubenImhoff @sidekock 
the effect of this new approach seems to be fairly strong, see for example the example in the gallery:

| v1.10 | latest |
|----------|----------|
| image    | image    |

aitaten:
Isn't that what we were looking for? I think the problem in left example (60, 90 minutes) is that the mismatching precipitation bring lower precipitation values than the actual distributions either from NWC or NWP. And these resampling guarantees a similar distribution. In the right example, the values look similar in all the lead-times whereas in the left it is clear that the smoothing (lost of high values) is happening in the middle lead-times where I am assuming the weight it is closer to 0,5. Just let me know what you think :smile:

dnerini:
totally agree with you, just wanted to make it clear that this has a significant impact on the output.

RubenImhoff:
Thanks for sharing, @dnerini! Based on the input radar data (see below), I'd agree that the changes seem to improve the distribution. Nevertheless, would be nice to see some more examples at some point (I'm sure some will come up from both KNMI and RMI soon). 
image

ladc:
Yes, I agree that this is actually better since you don't lose the extremes as much. Thanks for the examples. It might also fix a problem I had with the control member but I still need to run some tests, sorry for the delay.

ladc:
Applying it on the sample dataset for the hackathon I get a strange result, but maybe I'm using a bad combination of parameter settings. Maybe because the NWP intensities are so high, the nowcast intensity increases immediately at timestep 1 (ens 0 and mean are the images on the top right and bottom left, respectively). Any idea what's going on?
control_ens_mean_obs

Top left is the "control run" by the way.

RubenImhoff:
Hi @ladc, 
Thanks for sharing this example! Did you also turn on the smoothing by @sidekock? I think the top right and bottom left do, in principle, look quite good, but I agree that it is weird to have such an intensity jump. This almost gives the feeling that the weight for NWP compared to the nowcast is too high, but that should go well, I think.. 

Something else that could play a role is the following in the `blending/steps.py: code`
```
  if probmatching_method is not None and resample_distribution:
      # deal with missing values
      arr1 = R_pm_ep[t_index]
      arr2 = precip_models_pm_temp[j]
      arr2 = np.where(np.isnan(arr2), np.nanmin(arr2), arr2)
      arr1 = np.where(np.isnan(arr1), arr2, arr1)
      # resample weights based on cascade level 2
      R_pm_resampled = probmatching.resample_distributions(
          first_array=arr1,
          second_array=arr2,
          probability_first_array=weights_pm_normalized[0],
      )
```

We fill up the 'outside radar domain' extrapolation values with the NWP values to have no nans (which I initially thought was a good idea). But maybe that skews the distribution in the wrong direction? So maybe we should change that back to actually allowing for nans in the extrapolation component?

RubenImhoff:
@ladc, would the above be worth the test for your case?

ladc:
Thanks @RubenImhoff. I didn't turn on the smoothing, this was on purpose. 
The main issues for me is indeed the intensity jump. Could this be because the NWP has very high intensities, but as these are not present in the blended forecast, the latter is unrealistically scaled up *overall*? 
Do you suggest not filling *anything* outside the radar mask (so we have nans in the final forecast)? Or just for the resampling?

RubenImhoff:
I think just the resampling for now. This would at least keep the radar distribution lower in a case like this (because you push the distribution up by filling the nans with the values of the NWP forecast outside the radar domain).

aitaten:
Hi! Sorry to jump late for this discussion. I have a small question; how is it looking before doing the resampling/cdf? Because I am not sure if the problem is coming from the resample or from the merging of the cascades with different weights in different areas? @ladc , I did run your example data; maybe you could update or send me the snippet for recreating the figure so I can have a more detail look. :)

ladc:
Hi @aitaten you can find the plot_nwc.py and plot_radar.py script here - apologies for the messy code:
https://github.com/ladc/rmi-scripts

ladc:
I tried not filling the nan values (or rather not including them in the pmatching) - but I was a bit rushed so it's possible that the code contains errors:

control_ens_mean_obs

Warning: I will be fully offline in the coming week and still on vacation until 14 August.

The code is in my branch mentioned in  #370

aitaten:
For me it is not working either ... I had to change a couple of things in the merged code (#370 ) because `noise = None` was throwing an error and not creating any output. After, making sure the noise was 0, I obtained (as expected) the smoothing of the unpredictable scales. Not sure how @ladc you are making it without losing the small scales :disappointed: . Here my example:
forecast_202107041600

RubenImhoff:
@ladc, your second attempt looks already better to me! Should I change the code (back) to accepting nans outside the radar domain? Then we're good to go, I hope.

RubenImhoff:
And seeing the vacation of @ladc, what do the others think?

dnerini:
> And seeing the vacation of @ladc, what do the others think?

one remaining issue might be that using the weights from level 2 gives too much credit to NWP already from the first lead time, while you would expect a more progressive transition from the radar distribution...

RubenImhoff:
Based on the figure we made for the QJRMS paper, that should not be too much of an issue at that scale level:

image
image

(the lead time is in hours going from 0 to 12 hr)

RubenImhoff:
@dnerini, any idea on how to nicely fix this? Right now the nans outside the radar domain are filled with the NWP values, which changes the distribution of the extrapolation cascade data. We could go back to our earlier version where nans were allowed, or maybe we can think of a more elegant solution to handle the nans outside the radar domain without changing the distribution?

dnerini:
I wonder if we could simply "blend in" the NWP values outside the radar domain, also improving the seamlessness at t0. In practice we could try replacing the radar mask with zeros and see if that is enough. Alternatively, one could treat the part outside the radar domain separately and use a simple blending function to introduce the NWP values. Both approaches should avoid the jump at t0 while the resampling approach can stay the same. What do you think?

RubenImhoff:
That would be worth the try. Would we introduce any statistical problems in the resampling if we introduce zeros (or `np.nanmin(precip_radar)`) in the extrapolation cascade outside the radar domain? If not, that might be an easy solution while keeping the code nearly the same.

PR #397, opened 2024-07-19
make dimension.py xarray compatible

### LINKED ISSUE #400: Convert dimmension.py to xarray
This is the first pull request in which I rewrote a new module to xarray, namely the dimension.py module. This module actually could benefit a lot from the xarray datamodel, so I rewrote this to actually make use of the xarray methods.

mats-knmi:
Converted the methods in `dimension.py` to work with the new xarray datamodel. Since these methods can largely be replaced by xarray functionality, I updated these methods to make use of this.


pysteps/utils/dimension.py
```diff
@@ -14,26 +14,25 @@
...
 
-def aggregate_fields_time(R, metadata, time_window_min, ignore_nan=False):
+def aggregate_fields_time(dataset: xr.Dataset, time_window_min, ignore_nan=False):
     """Aggregate fields in time.
 
     Parameters
     ----------
-    R: array-like
-        Array of shape (t,m,n) or (l,t,m,n) containing
+    dataset: Dataset
```
dnerini: I would still mention here that this is an xarray dataset, even if it's already obvious from the type hints

dnerini: also, the methods expect certain attributes, not just any kind of dataset. These should be mentioned somehow. See for example the description of "metadata" where we referenced the documention in pysteps.io.importers.

pysteps/utils/dimension.py
```diff
@@ -100,36 +79,23 @@ def aggregate_fields_time(R, metadata, time_window_min, ignore_nan=False):
...
-def aggregate_fields_space(R, metadata, space_window, ignore_nan=False):
+def aggregate_fields_space(dataset: xr.Dataset, space_window, ignore_nan=False):
     """
     Upscale fields in space.
 
     Parameters
     ----------
-    R: array-like
-        Array of shape (m,n), (t,m,n) or (l,t,m,n) containing a single field or
+    dataset: Dataset
```
dnerini: need to update the name in the rest of the docstrings (see the text for space_window)

pysteps/utils/dimension.py
```diff
@@ -144,214 +110,154 @@ def aggregate_fields_space(R, metadata, space_window, ignore_nan=False):
 
     Returns
     -------
-    outputarray: array-like
-        The new array of aggregated fields of shape (k,j), (t,k,j) or (l,t,k,j),
-        where k = m*ypixelsize/space_window[1] and j = n*xpixelsize/space_window[0].
-    metadata: dict
-        The metadata with updated attributes.
+    dataset: Dataset
+        The new dataset.
```
dnerini, line 131: Is it self-explanatory how the fields are aggregate with a given spatial window?

pysteps/utils/dimension.py
```diff
@@ -14,26 +14,25 @@
...
+
+from pysteps.converters import compute_lat_lon
 
 _aggregation_methods = dict(
     sum=np.sum, mean=np.mean, nanmean=np.nanmean, nansum=np.nansum
 )
 
 
-def aggregate_fields_time(R, metadata, time_window_min, ignore_nan=False):
+def aggregate_fields_time(dataset: xr.Dataset, time_window_min, ignore_nan=False):
```
dnerini: need to specify the type of the output, as it's not detailed in the docstrings

dnerini (commented):
Fantastic work @mats-knmi ! Code is really looking much better (and this PR has a negative impact on the number of lines, brilliant!) and I know that the code you had to adapt is not easy to navigate... so well done! 

I only had few minor comments, mostly concerning documentation - how are we going to make sure that the structure of the dataset is clearly documented? perhaps it could also be a good idea to add some basic input validator with a decorator, what do you think?

pysteps/utils/dimension.py
mats-knmi: Yes indeed a documentation similar to the "metadata" description (but for the dataset structure), which we can link to from these docstrings, would be very a very useful addition. I will take a look at this.

mats-knmi:
```suggestion
    dataset: xarray.Dataset
```
Is this what you mean?

mats-knmi:
@dnerini I have added a ticket to the [project backlog](https://github.com/orgs/pySTEPS/projects/4?pane=info), for a [basic input validator with a decorator](https://github.com/pySTEPS/pysteps/issues/412).

mats-knmi:
@dnerini I believe I have also taken care of the remaining comments you had, let me know what you think

PR #405, opened 2024-07-19
Adapt pysteps to allow for postprocessing plugins

joeycasey87:
Added an infrastructure.py file to the postprocessing folder which should operate in the same way as the interface file does for the importers in the io folder. The postprocessor file is currently effectively empty as no postprocessor plugins have been added yet. A line has been added to the __init__ file so that it should act similar to the __init__ file in the io folder.


pysteps/postprocessing/__init__.py
```diff
@@ -2,3 +2,4 @@
 """Methods for post-processing of forecasts."""
 
 from . import ensemblestats
+from postprocessors import *
```
dnerini:
```suggestion
from .postprocessors import *
```

pysteps/postprocessing/postprocessors.py
dnerini, line 1: just wondering if "postprocessors" is a good name? sounds a bit too vague, particularly as a submodule of "postprocessing" ... can we try to be a bit more specific? perhaps use "diagnostic" instead? what do you think?  @ladc ?

ladc: Yes good point - "diagnostic" makes sense if we also want to add other postprocessors in the future which are not purely diagnostic (such as bias correction methods).

joeycasey87:
Hi @dnerini, could you please review this pull request when you have the time? The extended version of the cookiecutter plugin template which can be used to create importer, diagnostic postprocessing, and nowcast plugins is nearly finished but I require the postprocessing interface to be merged to ensure that everything is behaving correctly when tested. Cheers

ladc:
Thanks for this contribution! We discussed how to make it more generic and @joeycasey87 will try to refactor it tomorrow to allow for various kinds of post-processors including new ensemblestats and bias correction methods, for example. To be continued.

dnerini:
what's the status of this PR? should we mark it as a draft?

ladc:
Apologies for the force-push, but I have done a rebase to the latest master version. The plan is now to clean up some of the dummy code (maybe move this into test, but most likely remove it entirely), add the necessary test(s) in test_plugins_support.py and hopefully merge it into the master branch soonish, thanks in advance to @FelixE91 #nopressure

For the tests, we're now only checking the importer cookiecutter template. The new precipitation type plugin was created using the diagnostics template (I suppose), so we should add a test for this new template ( see also https://github.com/pySTEPS/pysteps/blob/e3325854dae006b1e78eb43a6d2203f2b0b71560/pysteps/tests/test_plugins_support.py#L17C1-L17C20 ) 

Question: do we really need a separate template for  all the plugin types? (Diagnostics, ensemblestats ... which are all in fact postprocessors)? Could we have one for any kind of postprocessors? Or even better, one cookiecutter template for all kinds of plugins which could even expose multiple functions (e.g. diagnostics and visualisation)? Not urgent at all.

pysteps/postprocessing/diagnostics.py
ladc, line 1: Could we just delete this file?

pysteps/postprocessing/ensemblestats.py
```diff
@@ -177,3 +177,11 @@ def banddepth(X, thr=None, norm=False):
         depth = (depth - depth.min()) / (depth.max() - depth.min())
 
     return depth
+
+
+# def ensemblestat_example1(filename, **kwargs):
+#     return "Hello, I am an example of postprocessing ensemble statistics."
+
```
ladc: To remove: lines 180-187

pysteps/postprocessing/interface.py
```diff
@@ -0,0 +1,277 @@
...
+
+    Parameters
+    ----------
+    name: str
+        Name of the method. The available options are:\n
+
+        diagnostics:
+
+        .. tabularcolumns:: |p{2cm}|L|
+
```
ladc: A bit too much noise in the documentation maybe?

ladc (commented):
Almost done, some final cleanup needed. There's also still some duplicate code but that will take a bit more time to simplify.

ladc:
Also, by updating the cookiecutter template I just realised that we broke the standard pysteps tests, so best to merge these changes in quite fast.

pysteps/postprocessing/diagnostics.py
FelixE91: I don't think so, at least not until we change the structure. The objective was to create a structure similar to what exists for the importers with `io.importers.py` and for the ensemblestats with `postprocessing.ensemblestats.py`.
The postprocessing `interface.py` and `__init__.py` currently read from diagnostics.py to check for available pre-installed functions[^1] - that's consistent with the way pre-installed ensemblestats functions are loaded and similar to the `io.__init__.py` and `io.interface.py` to load pre-installed importers.

New functions found in the *entry_points* are added to the available methods in the methods dictionary.

The (`io` and `postprocessing`) interface also reads from the module (`importers.py` and `ensemblestats.py`|`diagnostics.py`, respectively) **BUT** it would not add the functions to the methods dictionary. Instead, they must be added explicitly to the  `interface.py` methods dictionary  (*_importer_methods* and *_ensemblestats_methods*|*_diagnostics_methods*, respectively) to avoid an error in the *get_method()* function.
This might be fixed to add pre-installed functions from the module automatically to the methods dictionary.

[^1]: At the moment there are no pre-installed diagnostics. This might change in the future, unless we always use the cookiecutter to create plugins

FelixE91:
Additional tests are successful. CodeCov's getting closer to the target value. Still some work to do before we can merge into main.

pysteps/postprocessing/interface.py
```diff
@@ -0,0 +1,254 @@
...
+    get_method
+"""
+import importlib
+
+from pysteps.postprocessing import diagnostics, ensemblestats
+from pprint import pprint
+import warnings
+
+_diagnostics_methods = dict()
```
FelixE91, line 25: There is still a discrepancy here, as the dictionary in the `io.interface` is called *_importer_methods*, i.e. without “s” on importer. However, I would leave it as it is, as the use of plural is consistent with the module name.

pysteps/postprocessing/interface.py
```diff
@@ -0,0 +1,254 @@
...
+    )
+    pprint([(short_name, f.__name__) for short_name, f in interface_methods.items()])
+
+    module_methods_set = set(module_methods)
+    interface_methods_set = set(interface_methods.keys())
+
+    difference = module_methods_set ^ interface_methods_set
+    if len(difference) > 0:
+        # print("\nIMPORTANT:")
+        _diff = module_methods_set - interface_methods_set
```
FelixE91, line 138: Since we want to avoid differences, it should not be a problem that these lines with `if len(difference) > 0` are not considered in the tests - this means that the plugin implementation works well.

ci/test_plugin_support.py
```diff
@@ -13,15 +13,11 @@
...
 
-assert "abc_xxx" in io.interface._importer_methods
-assert "abc_yyy" in io.interface._importer_methods
+assert "institution_name" in io.interface._importer_methods
 
-from pysteps.io.importers import import_abc_xxx, import_abc_yyy
-
-import_abc_xxx("filename")
-import_abc_yyy("filename")
+from pysteps.io.importers import import_institution_name
```
FelixE91, line 20: Should similar lines be added to test the diagnostics plugin? Or not needed here in ci?

pysteps/io/interface.py
```diff
@@ -107,7 +113,7 @@ def importers_info():
 
     difference = available_importers ^ importers_in_the_interface
     if len(difference) > 0:
-        print("\nIMPORTANT:")
+        # print("\nIMPORTANT:")
```
FelixE91: Delete?

pysteps/postprocessing/diagnostics.py
```diff
@@ -0,0 +1,22 @@
...
+Available Diagnostics Postprocessors
+------------------------
+
+.. autosummary::
+    :toctree: ../generated/
+
+"""
+
+# Add your diagnostic_ function here AND add this method to the _diagnostics_methods
+# dictionary in postprocessing.interface.py
```
FelixE91, line 22: Let's see how we proceed with this file. I think we must keep it as mention in _Originally posted by @ladc in https://github.com/pySTEPS/pysteps/pull/405#discussion_r1972575936_

pysteps/postprocessing/interface.py
```diff
@@ -0,0 +1,254 @@
...
+.. currentmodule:: pysteps.postprocessing.interface
+
+.. autosummary::
+    :toctree: ../generated/
+
+    get_method
+"""
+import importlib
+
+from pysteps.postprocessing import diagnostics, ensemblestats
```
FelixE91, line 21: **Codacy Static Code Analysis** does not like the import of *diagnostics* as it isn't explicitly called. However, it is needed to add diagnostics plugins as `globals()[module]` in *add_postprocessor()* needs the diagnostics module to be imported.

tox.ini
```diff
@@ -65,9 +65,8 @@ commands =
     # Test the pysteps plugin support
     pip install cookiecutter
     cookiecutter -f --no-input https://github.com/pySTEPS/cookiecutter-pysteps-plugin -o {temp_dir}/
-    pip install {temp_dir}/pysteps-importer-abc
+    pip install {temp_dir}/pysteps-importer-institution-name
```
FelixE91, line 69: We must remember to update all these default names when we further develop or simplify the cookiecutter.

Maybe we can introduce an environment variable that sets the default name? Then we would only need to change it in one place.

pysteps/postprocessing/interface.py
```diff
@@ -0,0 +1,254 @@
+# -*- coding: utf-8 -*-
+"""
+pysteps.postprocessing.interface
+====================
+
+Interface for the postprocessing module.
+
+Support postprocessing types:
```
FelixE91, line 8: I have added that we currently support these two types of postprocessing plugins.
We might need to add some tests for the *ensemblestats* type later.

Note that we have removed the *ensemblestats* type from the [cookiecutter](https://github.com/pySTEPS/cookiecutter-pysteps-plugin) at the moment.

FelixE91 (commented):
Let's decide together whether we want to merge it like this or make some of the recommended changes first. @dnerini @ladc

ci/test_plugin_support.py
ladc: Not urgently needed here I think

pysteps/io/interface.py
```diff
@@ -53,8 +53,10 @@ def discover_importers():
 
     importlib.reload(pkg_resources)
 
-    for entry_point in pkg_resources.iter_entry_points(
-        group="pysteps.plugins.importer", name=None
+    for entry_point in list(
+        pkg_resources.iter_entry_points(group="pysteps.plugins.importer", name=None)
+    ) + list(
+        pkg_resources.iter_entry_points(group="pysteps.plugins.importers", name=None)
```
dnerini, line 59: thanks for adding this! could you please just add a short comment to explain this was necessary? I'm sure we'll forget otherwise ;)

pysteps/postprocessing/interface.py
dnerini: just ignore codacy

tox.ini
dnerini: do you suggest to introduce it in this PR?

pysteps/io/interface.py
ladc: Less is more.

tox.ini
ladc: I would suggest to leave it like this for now, only maybe adding a comment to avoid repeating the scenario where we were desperately looking for the source of an error when running the tests on the plugin side, not realising that those tests triggered tests in another repo that assume certain default naming conventions.

pysteps/io/interface.py
ladc: Done in c1dee1e

pysteps/io/interface.py
ladc: Done in c1dee1e

tox.ini
ladc: Added comment in f40a41c

ladc (approved):
For me, this is ready to merge

PR #407, opened 2024-07-20
Replace numpy.matrix with numpy.ndarray

### LINKED ISSUE #406: Deprecation Warning in steps blending: Replace numpy.matrix with numpy.ndarray
Our code currently uses the numpy.matrix subclass, which is generating a PendingDeprecationWarning as it is not the recommended way to represent matrices or deal with linear algebra in NumPy. We need to replace instances of numpy.matrix with numpy.ndarray to ensure future compatibility and eliminate the warning.

For example [here](https://github.com/pySTEPS/pysteps/actions/runs/10010421529/job/27671561775):
```
tests/test_blending_steps.py: 5184 warnings
  /home/runner/micromamba/envs/test_environment/lib/python3.11/site-packages/numpy/matrixlib/defmatrix.py:70: PendingDeprecationWarning: the matrix subclass is not the recommended way to represent matrices or deal with linear algebra (see https://docs.scipy.org/doc/numpy/user/numpy-for-matlab-users.html). Please adjust your code to use regular ndarray.
    return matrix(data, dtype=dtype, copy=False)
```
I think this warning is triggered in `blending/steps.py` [lines 1726-1744](https://github.com/pySTEPS/pysteps/blob/d77fe738a98ac3f647d977e998a3c218192228ca/pysteps/blending/steps.py#L1726) where we use numpy.matrix through the np.asmatrix and .getI() methods, which is likely triggering the warning. To fix this, we should replace the use of numpy.matrix with numpy.ndarray and use the appropriate functions from numpy and scipy for matrix operations.

dnerini:
closes #406

### DIFF
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -1732,14 +1732,15 @@ def calculate_weights_spn(correlations, cov):
             weights = np.nan_to_num(
                 weights, copy=True, nan=10e-5, posinf=10e-5, neginf=10e-5
             )
+            weights_dot_correlations = np.dot(weights, correlations)
             # If the dot product of the weights with the correlations is
             # larger than 1.0, we assign a weight of 0.0 to the noise (to make
             # it numerically stable)
-            if np.dot(weights, correlations) > 1.0:
+            if weights_dot_correlations > 1.0:
                 noise_weight = np.array([0])
             # Calculate the noise weight
             else:
-                noise_weight = np.sqrt(1.0 - np.dot(weights, correlations))
+                noise_weight = np.sqrt(1.0 - weights_dot_correlations)
             # Convert weights to a 1D array
             weights = np.array(weights).flatten()
             # Ensure noise_weight is a 1D array before concatenation
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -46,6 +46,7 @@ consists of the following main steps:
 import time
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


pysteps/blending/steps.py
```diff
@@ -1722,26 +1723,29 @@ def calculate_weights_spn(correlations, cov):
...
+            cov_matrix_inv = inv(cov_matrix)
+            # The component weights are the dot product between cov_matrix_inv and cor_vec
+            weights = np.dot(cov_matrix_inv, correlations)
             weights = np.nan_to_num(
                 weights, copy=True, nan=10e-5, posinf=10e-5, neginf=10e-5
             )
             # If the dot product of the weights with the correlations is
             # larger than 1.0, we assign a weight of 0.0 to the noise (to make
             # it numerically stable)
-            if weights.dot(correlations) > 1.0:
```
sidekock, line 1738: Would it not be more efficient to compute np.dot(weights, correlations) once and then do the check as you need this variable both in the if and else statement.

sidekock (changes requested):

pysteps/blending/steps.py
dnerini: good catch! see [0919ba2](https://github.com/pySTEPS/pysteps/pull/407/commits/0919ba26f2b4ea801eac52958a1de01be01e81ff)

RubenImhoff (approved):
This looks good to me, nice work!

Issue #376, opened 2024-07-15
Optimizing the parameters for the optical flow calculation


ladc:
[RMI/ KNMI / site specific] Is more better, and how to check this?
Number of frames, other parameters

ladc:
This can be done using ray as suggested by @dnerini 

see example: https://colab.research.google.com/drive/1pMAJZ8AZe7gUDgCogenfdlc-ayX6kVVM?usp=sharing

ladc:
Lukas-Kanade from openCV provides some uncertainty information about the u,v components, can we use this?

dnerini:
> see example: https://colab.research.google.com/drive/1pMAJZ8AZe7gUDgCogenfdlc-ayX6kVVM?usp=sharing

if possible, one by product of this issue could be adding this example as a tutorial to the pysteps documentation

aitaten:
Hi!! As I mentioned yday ... I implemented in this example the INCA motion which uses the errors of previous nowcating to correct the latest kind-of-TREC motion vector computation. It uses numba but I am not sure this is required as a new library in pySTEPS but maybe we can think if these errors are useful for task 8 (velocity perturbation settings) or can help in optimizing the optical flow calculation. Here is the link to the [example](https://github.com/aitaten/pysteps/blob/zamg/examples/plot_optical_flow_with_new_inca_motion.py)

dnerini:
> Hi!! As I mentioned yday ... I implemented in this example the INCA motion which uses the errors of previous nowcating to correct the latest kind-of-TREC motion vector computation. It uses numba but I am not sure this is required as a new library in pySTEPS but maybe we can think if these errors are useful for task 8 (velocity perturbation settings) or can help in optimizing the optical flow calculation. Here is the link to the [example](https://github.com/aitaten/pysteps/blob/zamg/examples/plot_optical_flow_with_new_inca_motion.py)

wouldn't this be rather a new feature to be implemented?

aitaten:
> > Hi!! As I mentioned yday ... I implemented in this example the INCA motion which uses the errors of previous nowcating to correct the latest kind-of-TREC motion vector computation. It uses numba but I am not sure this is required as a new library in pySTEPS but maybe we can think if these errors are useful for task 8 (velocity perturbation settings) or can help in optimizing the optical flow calculation. Here is the link to the [example](https://github.com/aitaten/pysteps/blob/zamg/examples/plot_optical_flow_with_new_inca_motion.py)
> 
[...]

PR #408, opened 2024-07-20
Example notebook for optimizing number of frames as described in issue #376

aitaten:
This is a notebook optimizing the number of frames for different lead-times for the Optical Flow (Lucas-Kanade) motion vector computation. The exampls is related to issue #376 but I think the second example commented in that issue should be also added.


dnerini:
Thanks @aitaten , this is looking very good! I will have a closer look, but I was wondering if it wouldn't be better to split this PR into two? Here I would rather keep the tuning tutorial, while the second should be about the INCA motion field -- by the way, I'm not sure why this is implemented in the example only, shouldn't we have it as an additional method in the motion module?

aitaten:
Thanks @dnerini for the review. Including the other example was a mistake so I have removed the unwanted file from the PR (addressed in commit [3f0c4ed](https://github.com/pySTEPS/pysteps/pull/408/commits/3f0c4edc9e3a2eeece1b745738e5a101ec237c1f)). Sorry about that.

I will do, as we talk, a separated PR and branch about the INCA motion field/TREC-like.

dnerini:
(I'm working on this, sorry, just going very slow...)

PR #409, opened 2024-07-23
Inca motion vector (#376)

aitaten:
Introduction of the correlation based motion vector computation as is currently computed in INCA. The error motion computation, and the NWP vector wind correction has not been included. However, this correlation based computation is similar to the original TREC computation adding a new OF method. The original idea is coming from #376

### DIFF
--- a/pysteps/motion/correlation.py
+++ b/pysteps/motion/correlation.py
@@ -1,0 +1,313 @@
+# -*- coding: utf-8 -*-
+"""
+pysteps.motion.correlation
+========================
+
+Implementation of the classical correlation based motion vector.
+
+.. autosummary::
+    :toctree: ../generated/
+
+    correlation
+"""
+
+import numpy as np
+import math
+
+from pysteps.decorators import check_input_frames
+from pysteps.utils.cleansing import detect_outliers
+from pysteps import utils
+
+#To delete
+import pdb
+
+#Check if numba can be imported
+try:
+    from numba import jit
+
+    NUMBA_IMPORTED = True
+except ImportError:
+    NUMBA_IMPORTED = False
+
+
+@check_input_frames(2)
+def correlation(
+    input_images,
+    settings={},
+    interp_method="idwinterp2d",
+    interp_kwargs=None,
+    dense=True,
+    nr_std_outlier=3,
+    k_outlier=30,
+    verbose=False,
+):
+
+    """
+    Run the correlation based optical flow method and interpolate the motion
+    vectors. This is using NUMBA to speed up the correlation computation.
+
+    The sparse motion vectors are finally interpolated to return the whole
+    motion field.
+
+    Parameters
+    ----------
+    input_images: ndarray_
+        Array of shape (T, m, n) containing a sequence of *T* two-dimensional
+        input images of shape (m, n). The indexing order in **input_images** is
+        assumed to be (time, latitude, longitude).
+
+        *T* = 2 is the minimum required number of images.
+        With *T* > 2, all the resulting obtained vectors are pooled together for
+        the final interpolation on a regular grid.
+
+    settings: dict, optional
+        Optional dictionary containing keyword arguments for the correlation
+        based algorithm.
+
+    interp_method: {"idwinterp2d", "rbfinterp2d"}, optional
+      Name of the interpolation method to use. See interpolation methods in
+      :py:mod:`pysteps.utils.interpolate`.
+
+    interp_kwargs: dict, optional
+        Optional dictionary containing keyword arguments for the interpolation
+        algorithm. See the documentation of :py:mod:`pysteps.utils.interpolate`.
+
+    dense: bool, optional
+        If True, return the three-dimensional array (2, m, n) containing
+        the dense x- and y-components of the motion field.
+
+        If False, return the sparse motion vectors as 2-D **xy** and **uv**
+        arrays, where **xy** defines the vector positions, **uv** defines the
+        x and y direction components of the vectors.
+
+    nr_std_outlier: int, optional
+        Maximum acceptable deviation from the mean in terms of number of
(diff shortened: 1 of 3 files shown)


pysteps/motion/correlation.py
```diff
@@ -0,0 +1,330 @@
...
+
+    References
+    ----------
+    Haiden, T., A. Kann, C. Wittmann, G. Pistotnik, B. Bica, and C. Gruber, 2011: The
+    Integrated Nowcasting through Comprehensive Analysis (INCA) System and Its
+    Validation over the Eastern Alpine Region. Wea. Forecasting, 26, 166–183.
+    """
+
+    if not NUMBA_IMPORTED:
+        raise ImportError(
```
dnerini: raise an MissingOptionalDependency instead
(need to import it: `from pysteps.exceptions import MissingOptionalDependency`)

```suggestion
        raise MissingOptionalDependency(
```

pysteps/motion/correlation.py
dnerini, line 1: "correlation" sounds a bit too vague, can we use a more precise name? is this approach equivalent to "TREC"?

pysteps/motion/correlation.py
```diff
@@ -0,0 +1,330 @@
...
+    uv = np.vstack([Uana[boole_], Vana[boole_]]).T
+    xy = np.vstack([XX[boole_], YY[boole_]]).T
+
+    # detect and remove outliers
+    outliers = detect_outliers(uv, nr_std_outlier, xy, k_outlier, verbose)
+    xy = xy[~outliers, :]
+    uv = uv[~outliers, :]
+
+    if verbose:
+        print("--- LK found %i sparse vectors ---" % xy.shape[0])
```
dnerini:
```suggestion
        print("--- found %i sparse vectors ---" % xy.shape[0])

```

pysteps/motion/correlation.py
```diff
@@ -0,0 +1,330 @@
...
+                                if (abs(isho) < nsh) and (abs(jsho) < nsh):
+                                    IS_tmp[int(j / nthin), int(i / nthin)] = -isho
+                                    JS_tmp[int(j / nthin), int(i / nthin)] = -jsho
+                            else:
+                                IS_tmp[int(j / nthin), int(i / nthin)] = 0
+                                JS_tmp[int(j / nthin), int(i / nthin)] = 0
+
+        return IS_tmp, JS_tmp
+
+else:
```
dnerini, line 326: don't need this since the function won't be executed without numba

pysteps/motion/correlation.py
```diff
@@ -0,0 +1,330 @@
...
+    input_images: ndarray_
+        Array of shape (T, m, n) containing a sequence of *T* two-dimensional
+        input images of shape (m, n). The indexing order in **input_images** is
+        assumed to be (time, latitude, longitude).
+
+        *T* = 2 is the minimum required number of images.
+        With *T* > 2, all the resulting obtained vectors are pooled together for
+        the final interpolation on a regular grid.
+
+    settings: dict, optional
```
dnerini, line 62: where are these settings documented? can't we convert this dictionary into optional arguments?

pysteps/motion/correlation.py
```diff
@@ -0,0 +1,330 @@
...
+        Optional dictionary containing keyword arguments for the correlation
+        based algorithm.
+
+    interp_method: {"idwinterp2d", "rbfinterp2d"}, optional
```
dnerini, line 66: out of curiosity, what approach uses INCA to densify the motion field?

pysteps/motion/correlation.py
```diff
@@ -0,0 +1,330 @@
...
+    if verbose:
+        print("--- total time: %.2f seconds ---" % (time.time() - t0))
+
+    return uvgrid
+
+
+if NUMBA_IMPORTED:
+
+    @jit(nopython=True)
+    def compute_motion_numba(image1, image2, NI, NJ, nthin, IS_tmp, JS_tmp):
```
dnerini, line 211: can't we simply have a non accelerated version in case numba is not installed instead of throwing an error?

pysteps/motion/correlation.py
```diff
@@ -0,0 +1,330 @@
...
+        interp_kwargs = dict()
+
+    if verbose:
+        print("Computing the motion field with the classical correlation-based method.")
+        t0 = time.time()
+
+    nr_fields = input_images.shape[0]
+    domain_size = (input_images.shape[1], input_images.shape[2])
+
+    PMIN = settings.get("PMIN", 0.05)
```
dnerini, line 139: we usually use ALL_CAPS names for global constants only.

pysteps/tests/test_motion.py
dnerini, line 1: need to add numba to the test requirements

dnerini (commented):
Thanks @aitaten this looks very cool! I left few initial comments to get started with the review

pysteps/motion/correlation.py
aitaten: TREC stands for tracking radar echo by cross-correlation (TREC). So I would say it is quite equivalent. I will change the name accordingly

pysteps/motion/correlation.py
aitaten: I does a cubic interpolation. I can also implemented if you want ... but again, It is done with numba :(

pysteps/motion/correlation.py
aitaten: Thanks ... I did some changes on these and other variables

pysteps/motion/correlation.py
dnerini: do you think it could be added as an additional interpolator in the pysteps/utils/interpolate.py module?  (I wouldn't mind the numba optional requirement)

PR #410, opened 2024-07-24
Performance fix

gjm174:
After profiling some changes which improve the overall performance, especially the loading of NWP files.

### DIFF
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -57,6 +57,8 @@ from pysteps.postprocessing import probmatching
 from pysteps.timeseries import autoregression, correlation
 from pysteps import blending
 
+from copy import deepcopy
+
 try:
     import dask
 
@@ -694,15 +696,11 @@ def forecast(
         # 5. Repeat precip_cascade for n ensemble members
         # First, discard all except the p-1 last cascades because they are not needed
         # for the AR(p) model
-        precip_cascade = [
-            precip_cascade[i][-ar_order:] for i in range(n_cascade_levels)
-        ]
 
-        precip_cascade = [
-            [precip_cascade[j].copy() for j in range(n_cascade_levels)]
-            for i in range(n_ens_members)
-        ]
-        precip_cascade = np.stack(precip_cascade)
+        precip_cascade = np.stack(
+            [[precip_cascade[i][-ar_order:].copy() for i in range(n_cascade_levels)]]
+            * n_ens_members
+        )
 
         # 6. Initialize all the random generators and prepare for the forecast loop
         randgen_prec, vps, generate_vel_noise = _init_random_generators(
@@ -773,8 +771,10 @@ def forecast(
             timestep_type = "list"
 
         extrap_kwargs["return_displacement"] = True
-        forecast_prev = precip_cascade
-        noise_prev = noise_cascade
+
+        forecast_prev = deepcopy(precip_cascade)
+        noise_prev = deepcopy(noise_cascade)
+
         t_prev = [0.0 for j in range(n_ens_members)]
         t_total = [0.0 for j in range(n_ens_members)]
 
--- a/pysteps/extrapolation/semilagrangian.py
+++ b/pysteps/extrapolation/semilagrangian.py
@@ -173,7 +173,7 @@ def extrapolate(
 
     if xy_coords is None:
         x_values, y_values = np.meshgrid(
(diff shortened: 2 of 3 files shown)


pysteps/extrapolation/semilagrangian.py
```diff
@@ -173,7 +173,7 @@ def extrapolate(
 
     if xy_coords is None:
         x_values, y_values = np.meshgrid(
-            np.arange(velocity.shape[2]), np.arange(velocity.shape[1])
+            np.arange(velocity.shape[2]), np.arange(velocity.shape[1]), copy=False
```
dnerini, line 176: I can see how this provides better performance, but are we 100% sure that there are no side effects?

pysteps/blending/steps.py
```diff
@@ -694,15 +694,11 @@ def forecast(
...
-            precip_cascade[i][-ar_order:] for i in range(n_cascade_levels)
-        ]
-
-        precip_cascade = [
-            [precip_cascade[j].copy() for j in range(n_cascade_levels)]
-            for i in range(n_ens_members)
-        ]
-        precip_cascade = np.stack(precip_cascade)
+
+        precip_cascade = np.stack(
```
dnerini, line 700: In this new version, you use a repeated reference to the same sliced data, which is more memory efficient but can lead to unintended side effects if the data is modified later. The previous version used deep copies to ensure that modifications to one ensemble member do not affect the others. Did you consider such aspects?

dnerini (commented):
Thanks @gjm174 for this contribution! I was curious to get an idea of the performance improvement, do you have some numbers? 

Also, I'm a bit skeptical on some of these changes: you removed copies of the data, which surely ensures a lower memory footprint. Are we sure these copies were not needed to avoid side effects later on in the code? All tests are green, which is good, but I wouldn't rely too much on our test suite to detect such unintended effects.

perhaps @RubenImhoff could also have a look since you worked more closely on the steps blending code?

gjm174:
> Thanks @gjm174 for this contribution! I was curious to get an idea of the performance improvement, do you have some numbers?
> 
> Also, I'm a bit skeptical on some of these changes: you removed copies of the data, which surely ensures a lower memory footprint. Are we sure these copies were not needed to avoid side effects later on in the code? All tests are green, which is good, but I wouldn't rely too much on our test suite to detect such unintended effects.
> 
> perhaps @RubenImhoff could also have a look since you worked more closely on the steps blending code?

The main performance improvement occurs in the reading of NWP files. The other minor changes save approximately 6-7 seconds. The loading of the NWP files now takes only one-fifth of the time it did previously. I have ensured that no unintended effects will occur by not making copies.

dnerini:
Very good thanks! I'll wait for a second opinion from @RubenImhoff and then I'm happy to merge this PR

pysteps/extrapolation/semilagrangian.py
dnerini: perhaps @pulkkins as the original author of this code can jump in here and give us a feedback on whether those copies were intentional or not?

RubenImhoff:
I will have a look at it after this weekend. Great work so far, @gjm174!

pysteps/blending/steps.py
RubenImhoff: I expect that this is still fine here, as this takes places prior to the dask-based parellization of the ensemble members.

pysteps/blending/steps.py
RubenImhoff, line 1: Talking about deep copies, should:
```
forecast_prev = precip_cascade
noise_prev = noise_cascade
```

get a deep copy?

pysteps/extrapolation/semilagrangian.py
RubenImhoff: Is it likely that the velocity shape changes? If not, I expect that the copy is not needed.

RubenImhoff (commented):
@gjm174, nice work! I left a few comments, but other than, it seems good to go. Nice work!

RubenImhoff (approved):

pysteps/extrapolation/semilagrangian.py
ladc: I think not, since the `np.arange(...)` creates a new array, and it's this new array that is no longer copied. When copy=True, the reference to the brand-new array at some point goes out of scope and the memory will be garbage collected anyway. I think this just avoids unnecessarily creating and throwing away an intermediate array.

PR #411, opened 2024-07-24
Fix support for timesteps list in case model has rain but radar does not

mats-knmi:


RubenImhoff (approved):

RubenImhoff:
Thanks for this extra fix related to #368, @mats-knmi!

PR #418, opened 2024-08-08
Adapt pysteps to allow for nowcast plugins.

Loickemajou:
Creating a DGMR plugin to be intergrated into the pysteps nowcasts methods.


pysteps/nowcasts/interface.py
```diff
@@ -90,6 +187,11 @@ def get_method(name):
     |  sseps          | short-space ensemble prediction system (SSEPS).       |
     |                 | Essentially, this is a localization of STEPS          |
     +-----------------+-------------------------------------------------------+
+    |  dgmr           | a deep generative model for the probabilistic    .    |
```
dnerini: why would you need to add this to the base pysteps package? Anything about DGMR should go within its plugin

pysteps/nowcasts/__init__.py
```diff
@@ -1,3 +1,8 @@
 """Implementations of deterministic and ensemble nowcasting methods."""
 
-from pysteps.nowcasts.interface import get_method
+from pysteps.nowcasts.interface import *
+
+try:
+    from dgmr_module_plugin import dgmr
+except ImportError:
+    dgmr = None
```
dnerini: why this block? the dgmr plugin, if installed, will be discovered below in `discover_nowcasts()`

dnerini (commented):
apologies for the slow action on this, I left few comments to clarify certain parts, but otherwise this is looking good!

So the actual plugin would be https://github.com/pySTEPS/pysteps-dgmr-nowcasts right? is everything ready on that side?

pysteps/nowcasts/__init__.py
dnerini: oh I see, is it because the function is called `forecast` and it should be added to a module named `dgmr`?  https://github.com/pySTEPS/pysteps-dgmr-nowcasts/blob/c8903bfa689951c555cfb275fb2100e36a683e40/setup.py#L86

pysteps/nowcasts/interface.py
Loickemajou: Goodmorning @dnerini , thanks for reviewing the pull request.

Well received, will remove that now.

Loickemajou:
> apologies for the slow action on this, I left few comments to clarify certain parts, but otherwise this is looking good!
> 
> So the actual plugin would be https://github.com/pySTEPS/pysteps-dgmr-nowcasts right? is everything ready on that side?

Hello @dnerini , thanks for reviewing the code.

Concerning the plugin, everything is ok, just left to include a test for the plugin.

pysteps/nowcasts/__init__.py
Loickemajou: > oh I see, is it because the function is called `forecast` and it should be added to a module named `dgmr`? https://github.com/pySTEPS/pysteps-dgmr-nowcasts/blob/c8903bfa689951c555cfb275fb2100e36a683e40/setup.py#L86

Also, in the get_method, actually when the plugin is installed, the dgmr is discovered by pysteps and it is possible to 

_dgmr=get_method('dgmr')_.

Though is not quite a nice way, included this block so that when the plugin is installed we could called it directly by doing 

_from pysteps.nowcasts import dgmr_.

Had no idea on how to do that differently and will be very delighted to have some suggestions.

ladc:
I think this can soon be merged, although it shows that the current way the plugins are structured is hitting its limits. It leads to a lot of duplicate code and maintainability issues, and also it's not so flexible. I'll open a separate issue about that. 

These modifications allow the pysteps-nowcast-dgmr plugin and other future nowcast plugins to be detected if installed.

dnerini:
> I think this can soon be merged, although it shows that the current way the plugins are structured is hitting its limits. It leads to a lot of duplicate code and maintainability issues, and also it's not so flexible. I'll open a separate issue about that.

Hi @ladc, could you please elaborate on the issues that you mentioned? anything we can do to improve it or should we rather go ahead with this PR as it is now and improve it in a separate effort?

pysteps/nowcasts/__init__.py
dnerini: I see you removed that part, @Loickemajou , does it mean that you found an alternative solution?

Loickemajou: > I see you removed that part, @Loickemajou , does it mean that you found an alternative solution?

Hello @dnerini 

I have removed the conflicting dependencies and model weights to ensure that they do not interfere with PySTEPS during execution.

However, I have not yet found an alternative solution and would appreciate any suggestions or insights you might have to help improve the integration process.

ladc:
I think the current state is fine, it enables nowcast plugins in pysteps, opening the door for all kinds of data driven methods. The reason why I said it's hitting its limits is because another internship student, @viktor40, wanted to create a plugin which included both post-processing and visualisation functionality, but the way the plugin system is currently designed only allows a single functionality type to be added. I suppose it's not too difficult to extend it but currently the internships are mostly finished so we'll have to do that at a later time.
I would also need to add a test for the nowcast loader, but for that we need to update the cookiecutter - so I propose to go ahead and merge this in, if it's ok for everyone.

pysteps/nowcasts/interface.py
```diff
@@ -54,6 +60,91 @@
...
+            setattr(nowcasts, nowcast_module_name, _module)
+
+
+def nowcasts_info():
+    """Print all the available nowcast methods."""
+
+    # nowcasts methods available in the `nowcasts` package
+    available_nowcasts = [
+        attr.split(".")[0]
+        for attr in os.listdir(" ".join(nowcasts.__path__))
```
ladc, line 109: @Loickemajou Is there a more elegant way to detect all the nowcast methods that doesn't use os.listdir?

Loickemajou: Thanks @ladc, for the remark. 

I will take time to look into alternative approaches and see if I can find a more elegant way to achieve this.

Issue #416, opened 2024-08-05
Issue with intensity jump when resampling distributions


dnerini:
The main issues for me is indeed the intensity jump. Could this be because the NWP has very high intensities, but as these are not present in the blended forecast, the latter is unrealistically scaled up *overall*? 
Do you suggest not filling *anything* outside the radar mask (so we have nans in the final forecast)? Or just for the resampling?

_Originally posted by @ladc in https://github.com/pySTEPS/pysteps/issues/390#issuecomment-2247776140_

dnerini:
I'm opening a new issue here to better track the problem described by @ladc in #390. 

@RubenImhoff did I understand correctly that you're working on it?

RubenImhoff:
Great, thanks, @dnerini! I'm not actively working on it, but happy to make some code changes. I think I prefer to have data outside the radar mask, but not filling anything outside the radar mask for resampling seems the way to go. What do you think?

ladc:
Should be fixed in #428.

PR #419, opened 2024-08-09
fix: change extrapolation values for resampling outside radar domain to zero

RubenImhoff:
Fix for issue #416.

The small fix introduced here, which sets all extrapolation component values to zero outside the radar domain (instead of to the NWP value as in the code now), gives the following results for the case indicated in #390:

Radar observations (up to 60 minutes ahead):
radar_observations

Blended forecast for ensemble member 0, without the smoothing mask as introduced in #379:
forecast_withoutsmoothing

Blended forecast for ensemble member 0, with the smoothing mask:
forecast_with_smoothing

See also #390 for the previous results. The first lead time is still not perfect, but we got a lot closer already. What do you think?

### DIFF
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -1545,7 +1545,7 @@ def forecast(
                             arr1 = R_pm_ep[t_index]
                             arr2 = precip_models_pm_temp[j]
                             arr2 = np.where(np.isnan(arr2), np.nanmin(arr2), arr2)
-                            arr1 = np.where(np.isnan(arr1), arr2, arr1)
+                            arr1 = np.where(np.isnan(arr1), np.nanmin(arr1), arr1)
                             # resample weights based on cascade level 2
                             R_pm_resampled = probmatching.resample_distributions(
                                 first_array=arr1,


RubenImhoff:
@ladc, what are your thoughts here?

ladc:
I also tried to fix this issue in the commit c7b7e0c471 (when addressing issue #370) but in a more complicated way. I changed the resampling to allow nan values in one of the arrays, and disregard the corresponding pixels in the resampling procedure. The results are different for the control member, but I cannot see a difference for a random member. Here's the animation comparing 447260cb (first, Ruben's fix applied to my control member branch) and 7ef8a32 (last, my control member branch with modifications to resample_distributions and nonparam_match_empirical_cdf to allow nans). 
In figure 1, there's a reduction in intensity in the control member for timestep 1 (maybe due to the explicit addition of zeroes in the resampling?) which is not present in figure 2, and I prefer the latter. I'm going to go through my code again to check if I haven't made any errors. 

control_ens_mean_obs_447260c
control_ens_mean_obs_7ef8a32

RubenImhoff:
@ladc, I agree that your implementation would even be cleaner. If you're confident about it, could you push those resampling changes to this branch?

ladc:
Hi everyone,
Sorry for the delay and the proliferation of branches. I've pushed my code with a ton of tests and you can find the result in the corresponding pull request #428

dnerini:
Are we closing this in favour of #428?

ladc:
That's OK for me.

PR #428, opened 2024-08-23
Modify probability matching code to allow ignoring a part of the domain

ladc:
The resampling now allows the presence of nans. For pixel locations with a nan, both arrays are ignored.
The probability matching has an extra argument ignore_indices which are left untouched. 

This functionality is added so that the values outside the radar mask, where there is only NWP data, don't affect the distribution of rainfall inside the mask.
In the case of high intensity in NWP but not in the radar for example, you'd otherwise get a huge jump in intensity due to the extremes affecting the radar nowcast area. 

Result is in the top right and bottom left:
control_ens_mean_obs

### DIFF
--- a/pysteps/postprocessing/probmatching.py
+++ b/pysteps/postprocessing/probmatching.py
@@ -111,10 +111,9 @@ def nonparam_match_empirical_cdf(initial_array, target_array, ignore_indices=Non
         p = np.percentile(target_array, 100 * (1 - war))
         target_array[target_array < p] = zvalue_trg
 
-    # flatten the arrays
-    arrayshape = initial_array_copy.shape
-    target_array = target_array.flatten()
-    initial_array_copy = initial_array_copy.flatten()
+    # flatten the arrays without copying them
+    target_array.reshape(-1)
+    initial_array_copy.reshape(-1)
 
     # rank target values
     order = target_array.argsort()
--- a/pysteps/postprocessing/probmatching.py
+++ b/pysteps/postprocessing/probmatching.py
@@ -113,8 +113,8 @@ def nonparam_match_empirical_cdf(initial_array, target_array, ignore_indices=Non
 
     # flatten the arrays without copying them
     arrayshape = initial_array_copy.shape
-    target_array.reshape(-1)
-    initial_array_copy.reshape(-1)
+    target_array = target_array.reshape(-1)
+    initial_array_copy = initial_array_copy.reshape(-1)
 
     # rank target values
     order = target_array.argsort()
--- a/pysteps/postprocessing/probmatching.py
+++ b/pysteps/postprocessing/probmatching.py
@@ -52,7 +52,7 @@ def compute_empirical_cdf(bin_edges, hist):
     return cdf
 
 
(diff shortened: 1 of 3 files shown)


RubenImhoff:
Fantastic, @ladc! I will have a look at it after the weekend. :)

ladc:
Thanks! There's still somewhat of a jump, but not sure if we can entirely avoid it.

dnerini:
~~Quick thought: would it perhaps help to simply "fade in" in few time steps after t0 from zeros to the NWP values outside the radar mask? This way, you would ensure a smooth transition~~

Edit: on a second thought and more careful read, you're approach achieves the same in a more elegant way, way to go!

pysteps/postprocessing/probmatching.py
```diff
@@ -64,55 +64,63 @@ def nonparam_match_empirical_cdf(initial_array, target_array):
     initial_array: array_like
         The initial array whose CDF is to be matched with the target.
     target_array: array_like
-        The target array.
+        The target array
+    ignore_indices: array_like, optional
```
dnerini, line 68: is this meant to be an array of booleans or an array of integers?

pysteps/postprocessing/probmatching.py
```diff
@@ -64,55 +64,63 @@ def nonparam_match_empirical_cdf(initial_array, target_array):
...
         p = np.percentile(target_array, 100 * (1 - war))
-        target_array = target_array.copy()
         target_array[target_array < p] = zvalue_trg
 
     # flatten the arrays
-    arrayshape = initial_array.shape
+    arrayshape = initial_array_copy.shape
     target_array = target_array.flatten()
-    initial_array = initial_array.flatten()
+    initial_array_copy = initial_array_copy.flatten()
```
dnerini: wondering if making another copy through `.flatten()` here is necessary or if we can simply use `initial_array_copy.shape = (-1,)`, what do you think?

dnerini: same goes for `target_array` actually, since this is already a copy of the input array from the very beginning

dnerini (approved):
Some minor comments, looking good otherwise!

RubenImhoff (approved):
Great work, @ladc! Besides @dnerini's comments, I've nothing to add. I think this will do it for now. :)

pysteps/postprocessing/probmatching.py
ladc: Both work, in fact. I've updated the documentation

ladc: Sorry, I clicked a wrong button and started a new review.

pysteps/postprocessing/probmatching.py
ladc: Fixed, thank you!

ladc (commented):
If all tests are green it's also good for me.

ladc:
Here's just a final animation with and without the smoothing between nwc and nwp (mask boundary fix by Simon)
row 1: "control" (other branch, not yet finished) - member - mean (without smoothing)
row 2: observation - member with smoothing - mean with smoothing

control_ens_mean_obs_enssmooth_meansmooth

PR #432, opened 2024-09-03
Added member and time dimension

gjm174:
Added member and time dimension support into the convert_input_to_xarray_dataset method.

### DIFF
--- a/pysteps/xarray_helpers.py
+++ b/pysteps/xarray_helpers.py
@@ -83,6 +83,7 @@ def convert_input_to_xarray_dataset(
     precip: np.ndarray,
     quality: np.ndarray | None,
     metadata: dict[str, str | float | None],
+    startdate: datetime | None = None,
 ) -> xr.Dataset:
     """
     Read a precip, quality, metadata tuple as returned by the importers
@@ -99,6 +100,8 @@ def convert_input_to_xarray_dataset(
     metadata: dict
         Metadata dictionary containing the attributes described in the
         documentation of :py:mod:`pysteps.io.importers`.
+    startdate: datetime, None
+        Datetime object containing the start date and time for the nowcast
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


mats-knmi:
Looks good, I just have a few questions, which I will put in line

pysteps/xarray_helpers.py
```diff
@@ -142,25 +169,37 @@ def convert_input_to_xarray_dataset(
...
                 "standard_name": attrs["standard_name"],
                 "long_name": attrs["long_name"],
                 "grid_mapping": "projection",
-                "transform": metadata["transform"],
-                "accutime": metadata["accutime"],
-                "threshold": metadata["threshold"],
-                "zerovalue": metadata["zerovalue"],
-                "zr_a": metadata["zr_a"],
-                "zr_b": metadata["zr_b"],
+                "xpixelsize": xpixelsize,
```
mats-knmi: I don't think we need to add the `xpixelsize` and `ypixelsize` as metadata, since the dataset has an evenly spaced `x` and `y` dimension. So I added these variables there. You can retrieve them from the dataset by doing `dataset.x.attrs.stepsize` and `dataset.y.attrs.stepsize`

pysteps/xarray_helpers.py
```diff
@@ -142,25 +169,37 @@ def convert_input_to_xarray_dataset(
...
+
+    metadata_keys = [
+        "transform",
+        "accutime",
+        "threshold",
+        "zerovalue",
+        "zr_a",
+        "zr_b",
+        "cartesian_unit",
+        "yorigin",
```
mats-knmi: I also don't think the `yorigin` needs to be kept as attribute any longer (at least not as attribute of the precip var), since at the end of this method we sort the dataset in the `x` and `y` dimension so the `yorigin` will always be `lower`.

gjm174: Done

pysteps/xarray_helpers.py
gjm174: Done

pysteps/xarray_helpers.py
```diff
@@ -142,25 +169,34 @@ def convert_input_to_xarray_dataset(
...
     }
+
+    metadata_keys = [
+        "transform",
+        "accutime",
+        "threshold",
+        "zerovalue",
+        "zr_a",
+        "zr_b",
+        "cartesian_unit",
```
mats-knmi: Oh actually one more thing: the cartesian unit is also already present in the attributes of the x and y dimension

gjm174: Done

PR #435, opened 2024-09-24
allow for significantly less memory usage in steps blending

mats-knmi:
By delaying the decomposition of the model data into cascade levels and allowing for model data to be delivered in float32, it seems to be possible to run the blending with almost 10x less memory than before in some tests I did.

### DIFF
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -119,24 +119,37 @@ def forecast(
       ordered by timestamp from oldest to newest. The time steps between the
       inputs are assumed to be regular.
     precip_models: array-like
-      Array of shape (n_models,timesteps+1) containing, per timestep (t=0 to
-      lead time here) and per (NWP) model or model ensemble member, a
-      dictionary with a list of cascades obtained by calling a method
-      implemented in :py:mod:`pysteps.cascade.decomposition`. In case of one
-      (deterministic) model as input, add an extra dimension to make sure
-      precip_models is five dimensional prior to calling this function.
-      It is also possible to supply the original (NWP) model forecasts containing
-      rainfall fields as an array of shape (n_models,timestep+1,m,n), which will
-      then be decomposed in this function. Note that for an operational application
-      or for testing with multiple model runs, it is recommended to decompose
-      the model forecasts outside beforehand, as this reduces calculation times.
+      Either raw (NWP) model forecast data or decomposed (NWP) model forecast data.
+      If you supply decomposed data, it needs to be an array of shape
+      (n_models,timesteps+1) containing, per timestep (t=0 to lead time here) and
+      per (NWP) model or model ensemble member, a dictionary with a list of cascades
+      obtained by calling a method implemented in :py:mod:`pysteps.cascade.decomposition`.
+      If you supply the original (NWP) model forecast data, it needs to be an array of shape
+      (n_models,timestep+1,m,n) containing rainfall fields, which will
+      then be decomposed in this function.
+
+      Depending on your use case it can be advantageous to decompose the model
+      forecasts outside beforehand, as this slightly reduces calculation times.
       This is possible with :py:func:`pysteps.blending.utils.decompose_NWP`,
       :py:func:`pysteps.blending.utils.compute_store_nwp_motion`, and
-      :py:func:`pysteps.blending.utils.load_NWP`.
+      :py:func:`pysteps.blending.utils.load_NWP`. However, if you have a lot of (NWP) model
+      members (e.g. 1 model member per nowcast member), this can lead to excessive memory
+      usage.
+
+      To further reduce memory usage, both this array and the ``velocity_models`` array
+      can be given as float32, they will then be converted to float64 before computations
+      to minimize loss in precision
+
+      In case of one (deterministic) model as input, add an extra dimension to make sure
+      precip_models is five dimensional prior to calling this function.
     velocity: array-like
       Array of shape (2,m,n) containing the x- and y-components of the advection
       field. The velocities are assumed to represent one time step between the
       inputs. All values are required to be finite.
+
+      To reduce memory usage, both this array and the ``precip_models`` array
+      can be given as float32, they will then be converted to float64 before computations
+      to minimize loss in precision
     velocity_models: array-like
       Array of shape (n_models,timestep,2,m,n) containing the x- and y-components
       of the advection field for the (NWP) model field per forecast lead time.
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -51,8 +51,7 @@ from multiprocessing.pool import ThreadPool
 
 import numpy as np
 from scipy.linalg import inv
-from scipy.ndimage import (binary_dilation, generate_binary_structure,
-                           iterate_structure)
+from scipy.ndimage import binary_dilation, generate_binary_structure, iterate_structure
 
 from pysteps import blending, cascade, extrapolation, noise, utils
 from pysteps.nowcasts import utils as nowcast_utils
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -644,8 +644,30 @@ def forecast(
     else:
         # 2.3.3 If zero_precip_radar, make sure that precip_cascade does not contain
         # only nans or infs. If so, fill it with the zero value.
-        if zero_precip_radar:
-            precip_cascade[~np.isfinite(precip_cascade)] = np.nanmin(precip)
+        if np.any(precip_cascade[np.isfinite(precip_cascade)]):
+            precip_cascade[~np.isfinite(precip_cascade)] = np.nanmin(precip_cascade)
+        else:
+            done = False
+            for t in timesteps:
+                if done:
+                    break
+                for j in range(precip_models.shape[0]):
+                    if not blending.utils.check_norain(
+                        precip_models, precip_thr, norain_thr
+                    ):
+                        precip_models_cascade_temp = decompositor(
+                            precip_models[j, t, :, :],
+                            bp_filter=bp_filter,
+                            fft_method=fft,
+                            output_domain=domain,
+                            normalize=True,
+                            compute_stats=True,
+                            compact_output=True,
+                        )["cascade_levels"]
+                        precip_cascade[~np.isfinite(precip_cascade)] = np.nanmin(
+                            precip_models_cascade_temp
+                        )
+                        break
 
         # 2.3.5 If zero_precip_radar is True, only use the velocity field of the NWP
         # forecast. I.e., velocity (radar) equals velocity_model at the first time
@@ -2426,7 +2448,7 @@ def _init_noise_cascade(
 
 def _fill_nans_infs_nwp_cascade(
     precip_models_cascade,
-    precip_models_pm,
+    precip_models,
     precip_cascade,
     precip,
     mu_models,
@@ -2438,34 +2460,30 @@ def _fill_nans_infs_nwp_cascade(
     min_cascade = np.nanmin(precip_cascade)
     min_precip = np.nanmin(precip)
     precip_models_cascade[~np.isfinite(precip_models_cascade)] = min_cascade
-    precip_models_pm[~np.isfinite(precip_models_pm)] = min_precip
+    precip_models[~np.isfinite(precip_models)] = min_precip
     # Also set any nans or infs in the mean and sigma of the cascade to
     # respectively 0.0 and 1.0
     mu_models[~np.isfinite(mu_models)] = 0.0
     sigma_models[~np.isfinite(sigma_models)] = 0.0
 
-    return precip_models_cascade, precip_models_pm, mu_models, sigma_models
+    return precip_models_cascade, precip_models, mu_models, sigma_models
 
 
-def _determine_max_nr_rainy_cells_nwp(
-    precip_models_pm, precip_thr, n_models, timesteps
-):
+def _determine_max_nr_rainy_cells_nwp(precip_models, precip_thr, n_models, timesteps):
     """Initialize noise based on the NWP field time step where the fraction of rainy cells is highest"""
     if precip_thr is None:
-        precip_thr = np.nanmin(precip_models_pm)
+        precip_thr = np.nanmin(precip_models)
 
     max_rain_pixels = -1
     max_rain_pixels_j = -1
     max_rain_pixels_t = -1
     for j in range(n_models):
         for t in timesteps:
-            rain_pixels = precip_models_pm[j][t][
-                precip_models_pm[j][t] > precip_thr
-            ].size
+            rain_pixels = precip_models[j][t][precip_models[j][t] > precip_thr].size
             if rain_pixels > max_rain_pixels:
                 max_rain_pixels = rain_pixels
                 max_rain_pixels_j = j
                 max_rain_pixels_t = t
-    precip_noise_input = precip_models_pm[max_rain_pixels_j][max_rain_pixels_t]
+    precip_noise_input = precip_models[max_rain_pixels_j][max_rain_pixels_t]
 
     return precip_noise_input.astype(np.float64, copy=False)
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -645,12 +645,12 @@ def forecast(
         # 2.3.3 If zero_precip_radar, make sure that precip_cascade does not contain
         # only nans or infs. If so, fill it with the zero value.
         if zero_precip_radar:
+            # Look for a timestep and member with rain so that we have a sensible decomposition
             done = False
             for t in timesteps:
                 if done:
                     break
                 for j in range(precip_models.shape[0]):
-                    # Look for a timestep and member with rain so that we have a sensible decomposition
                     if not blending.utils.check_norain(
                         precip_models[j, t], precip_thr, norain_thr
                     ):
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -51,8 +51,7 @@ from multiprocessing import Pool
 
 import numpy as np
 from scipy.linalg import inv
-from scipy.ndimage import (binary_dilation, generate_binary_structure,
-                           iterate_structure)
+from scipy.ndimage import binary_dilation, generate_binary_structure, iterate_structure
 
 from pysteps import blending, cascade, extrapolation, noise, utils
 from pysteps.nowcasts import utils as nowcast_utils
@@ -787,7 +786,7 @@ def forecast(
                 starttime = time.time()
 
             if precip_models_cascade is not None:
-                decomp_precip_models = list(precip_models_cascade[:,t])
+                decomp_precip_models = list(precip_models_cascade[:, t])
 
             if precip_models.shape[0] == 1:
                 decomp_precip_models = [
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -644,35 +644,34 @@ def forecast(
     else:
         # 2.3.3 If zero_precip_radar, make sure that precip_cascade does not contain
         # only nans or infs. If so, fill it with the zero value.
-        if np.any(precip_cascade[np.isfinite(precip_cascade)]):
-            precip_cascade[~np.isfinite(precip_cascade)] = np.nanmin(precip_cascade)
-        elif precip_models_cascade is not None:
-            precip_cascade[~np.isfinite(precip_cascade)] = np.nanmin(
-                precip_models_cascade
-            )
-        else:
-            done = False
-            for t in timesteps:
-                if done:
-                    break
-                for j in range(precip_models.shape[0]):
-                    if not blending.utils.check_norain(
-                        precip_models, precip_thr, norain_thr
-                    ):
-                        precip_models_cascade_temp = decompositor(
-                            precip_models[j, t, :, :],
-                            bp_filter=bp_filter,
-                            fft_method=fft,
-                            output_domain=domain,
-                            normalize=True,
-                            compute_stats=True,
-                            compact_output=True,
-                        )["cascade_levels"]
-                        precip_cascade[~np.isfinite(precip_cascade)] = np.nanmin(
-                            precip_models_cascade_temp
-                        )
-                        done = True
+        if zero_precip_radar:
+            if precip_models_cascade is not None:
+                precip_cascade[~np.isfinite(precip_cascade)] = np.nanmin(
+                    precip_models_cascade
+                )
+            else:
+                done = False
+                for t in timesteps:
+                    if done:
                         break
+                    for j in range(precip_models.shape[0]):
+                        if not blending.utils.check_norain(
+                            precip_models, precip_thr, norain_thr
+                        ):
+                            precip_models_cascade_temp = decompositor(
+                                precip_models[j, t, :, :],
+                                bp_filter=bp_filter,
+                                fft_method=fft,
+                                output_domain=domain,
+                                normalize=True,
+                                compute_stats=True,
+                                compact_output=True,
+                            )["cascade_levels"]
+                            precip_cascade[~np.isfinite(precip_cascade)] = np.nanmin(
+                                precip_models_cascade_temp
+                            )
+                            done = True
+                            break
 
         # 2.3.5 If zero_precip_radar is True, only use the velocity field of the NWP
         # forecast. I.e., velocity (radar) equals velocity_model at the first time
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -646,6 +646,10 @@ def forecast(
         # only nans or infs. If so, fill it with the zero value.
         if np.any(precip_cascade[np.isfinite(precip_cascade)]):
             precip_cascade[~np.isfinite(precip_cascade)] = np.nanmin(precip_cascade)
+        elif precip_models_cascade is not None:
+            precip_cascade[~np.isfinite(precip_cascade)] = np.nanmin(
+                precip_models_cascade
+            )
         else:
             done = False
             for t in timesteps:
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -645,33 +645,33 @@ def forecast(
         # 2.3.3 If zero_precip_radar, make sure that precip_cascade does not contain
         # only nans or infs. If so, fill it with the zero value.
         if zero_precip_radar:
-            if precip_models_cascade is not None:
-                precip_cascade[~np.isfinite(precip_cascade)] = np.nanmin(
-                    precip_models_cascade
-                )
-            else:
-                done = False
-                for t in timesteps:
-                    if done:
-                        break
-                    for j in range(precip_models.shape[0]):
-                        if not blending.utils.check_norain(
-                            precip_models, precip_thr, norain_thr
-                        ):
-                            precip_models_cascade_temp = decompositor(
-                                precip_models[j, t, :, :],
-                                bp_filter=bp_filter,
-                                fft_method=fft,
-                                output_domain=domain,
-                                normalize=True,
-                                compute_stats=True,
-                                compact_output=True,
-                            )["cascade_levels"]
+            done = False
+            for t in timesteps:
+                if done:
+                    break
+                for j in range(precip_models.shape[0]):
+                    if not blending.utils.check_norain(
+                        precip_models, precip_thr, norain_thr
+                    ):
+                        if precip_models_cascade is not None:
                             precip_cascade[~np.isfinite(precip_cascade)] = np.nanmin(
-                                precip_models_cascade_temp
+                                precip_models_cascade[j, t]
                             )
-                            done = True
-                            break
+                            continue
+                        precip_models_cascade_temp = decompositor(
+                            precip_models[j, t, :, :],
+                            bp_filter=bp_filter,
+                            fft_method=fft,
+                            output_domain=domain,
+                            normalize=True,
+                            compute_stats=True,
+                            compact_output=True,
+                        )["cascade_levels"]
+                        precip_cascade[~np.isfinite(precip_cascade)] = np.nanmin(
+                            precip_models_cascade_temp
+                        )
+                        done = True
+                        break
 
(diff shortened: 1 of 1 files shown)


mats-knmi:
@sidekock I realized yesterday that, with these changes, we can run pysteps a lot cheaper at KNMI. I have tested these changes and it all still seems to run (and fast enough). Maybe you can take these changes along when you start on the refactoring.

RubenImhoff:
Hi @mats-knmi, this looks great, many thanks for the great improvement! Do I understand it well that the code now prefers to work with the NWP rainfall data instead of the decomposed data (whereas this used to be the other way around)? With other words, do we still want to support decomposed NWP data as input? It would be a breaking change, so this needs some discussion, but it would make the code cleaner if it is not necessary anymore.

RubenImhoff:
In addition, I think we should update the function arguments a bit (see below), where we then would switch around the array shape and description (favoring the four dimension without decomposition over the two dimensions with decomposition) and indicating that float32 is supported and even recommended for memory usage.
```
    precip_models: array-like
      Array of shape (n_models,timesteps+1) containing, per timestep (t=0 to
      lead time here) and per (NWP) model or model ensemble member, a
      dictionary with a list of cascades obtained by calling a method
      implemented in :py:mod:`pysteps.cascade.decomposition`. In case of one
      (deterministic) model as input, add an extra dimension to make sure
      precip_models is five dimensional prior to calling this function.
      It is also possible to supply the original (NWP) model forecasts containing
      rainfall fields as an array of shape (n_models,timestep+1,m,n), which will
      then be decomposed in this function. Note that for an operational application
      or for testing with multiple model runs, it is recommended to decompose
      the model forecasts outside beforehand, as this reduces calculation times.
      This is possible with :py:func:`pysteps.blending.utils.decompose_NWP`,
      :py:func:`pysteps.blending.utils.compute_store_nwp_motion`, and
      :py:func:`pysteps.blending.utils.load_NWP`.
```

mats-knmi:
Yes indeed, personally I would prefer to work with rainfall data in stead of decomposed data. However this might not be the case as much if you have less NWP model members (like RMI has), since then the decomposed data is not as big as it is for the KNMI usecase and the small extra performance you get by pre-decomposing it might be worth it. So I would be in favor of leaving both options in there, but I will see if I can try to word this nicely in the docstring.

mats-knmi:
@RubenImhoff I have updated the docstring, let me know if it is ok like this

pysteps/blending/steps.py
```diff
@@ -117,24 +119,37 @@ def forecast(
...
-      then be decomposed in this function. Note that for an operational application
-      or for testing with multiple model runs, it is recommended to decompose
-      the model forecasts outside beforehand, as this reduces calculation times.
+      Either raw (NWP) model forecast data or decomposed (NWP) model forecast data.
+      If you supply decomposed data, it needs to be an array of shape
+      (n_models,timesteps+1) containing, per timestep (t=0 to lead time here) and
+      per (NWP) model or model ensemble member, a dictionary with a list of cascades
+      obtained by calling a method implemented in :py:mod:`pysteps.cascade.decomposition`.
+      If you supply the original (NWP) model forecast data, it needs to be an array of shape
+      (n_models,timestep+1,m,n) containing rainfall fields, which will
```
RubenImhoff:
```suggestion
      (n_models,timestep+1,m,n) containing precipitation (or other) fields, which will
```

pysteps/blending/steps.py
```diff
@@ -117,24 +119,37 @@ def forecast(
...
+      :py:func:`pysteps.blending.utils.load_NWP`. However, if you have a lot of (NWP) model
+      members (e.g. 1 model member per nowcast member), this can lead to excessive memory
+      usage.
+
+      To further reduce memory usage, both this array and the ``velocity_models`` array
+      can be given as float32, they will then be converted to float64 before computations
+      to minimize loss in precision
+
+      In case of one (deterministic) model as input, add an extra dimension to make sure
+      precip_models is five dimensional prior to calling this function.
```
RubenImhoff:
```suggestion
      precip_models is four dimensional prior to calling this function.
```
Right?

pysteps/blending/steps.py
```diff
@@ -117,24 +119,37 @@ def forecast(
...
+      forecasts outside beforehand, as this slightly reduces calculation times.
       This is possible with :py:func:`pysteps.blending.utils.decompose_NWP`,
       :py:func:`pysteps.blending.utils.compute_store_nwp_motion`, and
-      :py:func:`pysteps.blending.utils.load_NWP`.
+      :py:func:`pysteps.blending.utils.load_NWP`. However, if you have a lot of (NWP) model
+      members (e.g. 1 model member per nowcast member), this can lead to excessive memory
+      usage.
+
+      To further reduce memory usage, both this array and the ``velocity_models`` array
+      can be given as float32, they will then be converted to float64 before computations
```
RubenImhoff:
```suggestion
      can be given as float32. They will then be converted to float64 before computations
```

pysteps/blending/steps.py
```diff
@@ -117,24 +119,37 @@ def forecast(
...
+      to minimize loss in precision
```
RubenImhoff:
```suggestion
      to minimize loss in precision.
```

pysteps/blending/steps.py
```diff
@@ -117,24 +119,37 @@ def forecast(
...
+
+      In case of one (deterministic) model as input, add an extra dimension to make sure
+      precip_models is five dimensional prior to calling this function.
     velocity: array-like
       Array of shape (2,m,n) containing the x- and y-components of the advection
       field. The velocities are assumed to represent one time step between the
       inputs. All values are required to be finite.
+
+      To reduce memory usage, both this array and the ``precip_models`` array
```
RubenImhoff:
```suggestion
      To reduce memory usage, this array
```
To keep the docstrings short, I've removed mentioning ``precip_models`` here (as ``velocity_models`` is also mentioned under the docstring of ``precip_models``.

By the way, I think this addition should be under ``velocity_models`` (it is under ``velocity`` now).

pysteps/blending/steps.py
```diff
@@ -117,24 +119,37 @@ def forecast(
...
+      can be given as float32, they will then be converted to float64 before computations
```
RubenImhoff:
```suggestion
      can be given as float32. They will then be converted to float64 before computations
```

pysteps/blending/steps.py
```diff
@@ -117,24 +119,37 @@ def forecast(
...
+      to minimize loss in precision
```
RubenImhoff:
```suggestion
      to minimize loss in precision.
```

pysteps/blending/steps.py
```diff
@@ -633,37 +645,14 @@ def forecast(
...
-            sigma_models,
-        ) = _fill_nans_infs_nwp_cascade(
-            precip_models_cascade,
-            precip_models_pm,
-            precip_cascade,
-            precip,
-            mu_models,
-            sigma_models,
-        )
+            precip_cascade[~np.isfinite(precip_cascade)] = np.nanmin(precip)
```
RubenImhoff:
```suggestion
            precip_cascade[~np.isfinite(precip_cascade)] = np.nanmin(precip_models)
```

However, I expect that the filler value should be the decomposed ``precip_models`` value as this is a normalized value which is used in the ``precip_cascade`` (also decomposed and normalized).

pysteps/blending/steps.py
```diff
@@ -2043,78 +2089,25 @@ def _compute_cascade_decomposition_radar(
...
-        # Recomposed cascade will have shape: [n_models, n_timesteps, m, n]
-        precip_models_pm = []
-        for i in range(precip_models_cascade.shape[0]):
-            precip_model_pm = []
-            for time_step in range(precip_models_cascade.shape[1]):
-                precip_model_pm.append(
-                    recompositor(precip_models_cascade[i, time_step])
-                )
-            precip_models_pm.append(precip_model_pm)
+        precip_model_pm = []
```
RubenImhoff: Is precip_model_pm still the most logical naming here, since you have changed it in the main loop to ``precip_models``?

RubenImhoff (changes requested):
Looks great so far! I've added a few comments; after that, I think we are good to go. :)

pysteps/blending/steps.py
mats-knmi: Yes I thought of this as well, I guess I can update it everywhere in the file where there is still precip_models_pm

pysteps/blending/steps.py
mats-knmi: Ah yes, this used to be `precip_models_cascade`, but that is not available anymore at this point in the code (if you supply just the regular model precipitation data). I am not quite sure what to do with this zero value business, as there is a zero value as well in the metadata, but that is rarely used as `np.nanmin(values)` is a good alternative.

In this case: what would be the downside of using `np.nanmin(precip_models)` here? Would it be easy to turn this into a normalized value?

mats-knmi:
@dnerini The unit tests github actions suddenly fail during the conda environment initialization, do you have any idea what is going on here? I didn't change anything that would impact this.

dnerini:
> @dnerini The unit tests github actions suddenly fail during the conda environment initialization, do you have any idea what is going on here? I didn't change anything that would impact this.

fixed it temporarily by pinning an older version of micromamba

pysteps/blending/steps.py
RubenImhoff: @mats-knmi, sorry for the slow reply here! I think the biggest difference is that ``precip_cascade`` is normalized to zero mean and standard deviation of one, so the min value may not be the same as in ``precip_models``. However, if I have read the code well, you could just use ``precip_cascade`` here, as it should always have a value somewhere in the grid, right? I'm not 100% sure about this statement, so good to double check, but if all values are infinites, we should be treating it as zero precipitation anyway.

mats-knmi: Hmmm yes, but the comment above this specifically says `make sure that precip_cascade does not contain only nans or infs`, which to me seems to indicate that it could only contain nans or infs... But actually I have been wondering for some time now why we would only fill nans or infs when there is no rain in the radar image. It would be sensible to fill nans and infs anyways right (if they are present)?

mats-knmi: I made a change  to check if `precip_cascade` is only nans/infs. If not we can use `nanmin(precip_cascade)`, but else we have to use `nanmin(precip_models_cascade)`, which means we have to look for a timestep (and model member) in the model field that has precipitation so that we can decompose that to get this value.

I am pretty sure this should always work and not hurt performance a lot. It may be a little bit overcomplicated though, so if we can do without that would still be nice.

RubenImhoff: Looks good - this could work, I think. The alternative would be to save a zero value up front and just use that in the else statement. What do you think?

mats-knmi: That would be neater, but I don't know what this zero value would be, because a zero value for a cascade is different than for a regular precip array right?

RubenImhoff: Yes, that's true indeed. Perhaps best to just leave it as you have proposed for now. :)

pysteps/blending/steps.py
```diff
@@ -644,8 +644,30 @@ def forecast(
     else:
         # 2.3.3 If zero_precip_radar, make sure that precip_cascade does not contain
         # only nans or infs. If so, fill it with the zero value.
-        if zero_precip_radar:
-            precip_cascade[~np.isfinite(precip_cascade)] = np.nanmin(precip)
+        if np.any(precip_cascade[np.isfinite(precip_cascade)]):
+            precip_cascade[~np.isfinite(precip_cascade)] = np.nanmin(precip_cascade)
+        else:
+            done = False
```
RubenImhoff, line 649: Is ``done`` set to ``True`` at some point in the code?

mats-knmi: Oops, forgot haha

RubenImhoff:
With the additions, we may have to add some tests for the coverage. But other than that, I think we're good to go. :) Nice work!

RubenImhoff (approved):

sidekock:
@mats-knmi @RubenImhoff I am planning to look through everything this afternoon.

sidekock:
@mats-knmi, I have some questions before I go more in-depth:
- I assume both options (providing raw NWP and decomposed) are still fully supported?
- Are the performance increases introduced now also applicable to the old method of decomposing first?
- Has there been any analysis as to the impact on final nowcast of the 32bit vs 64bit data? 
- Is there a decrease in performance (time for a forecast)? I would expect this because now the computations are done during the run...

pysteps/blending/steps.py
```diff
@@ -117,20 +119,29 @@ def forecast(
...
-      (deterministic) model as input, add an extra dimension to make sure
-      precip_models is five dimensional prior to calling this function.
-      It is also possible to supply the original (NWP) model forecasts containing
-      rainfall fields as an array of shape (n_models,timestep+1,m,n), which will
-      then be decomposed in this function. Note that for an operational application
-      or for testing with multiple model runs, it is recommended to decompose
-      the model forecasts outside beforehand, as this reduces calculation times.
+      Either raw (NWP) model forecast data or decomposed (NWP) model forecast data.
+      If you supply decomposed data, it needs to be an array of shape
+      (n_models,timesteps+1) containing, per timestep (t=0 to lead time here) and
```
sidekock, line 124: Has there been any change for people using the decomposed inputs compared to the previous version for the input? From the documentation this looks to be the case but I want to double check this with you

pysteps/blending/steps.py
```diff
@@ -117,20 +119,29 @@ def forecast(
...
+      per (NWP) model or model ensemble member, a dictionary with a list of cascades
+      obtained by calling a method implemented in :py:mod:`pysteps.cascade.decomposition`.
+      If you supply the original (NWP) model forecast data, it needs to be an array of shape
```
sidekock, line 127: Is this checked explicitly in the code? I am not sure this is possible, but maybe you have a good suggestion here.

pysteps/blending/steps.py
```diff
@@ -117,20 +119,29 @@ def forecast(
...
+      (n_models,timestep+1,m,n) containing precipitation (or other) fields, which will
+      then be decomposed in this function.
+
+      Depending on your use case it can be advantageous to decompose the model
+      forecasts outside beforehand, as this slightly reduces calculation times.
       This is possible with :py:func:`pysteps.blending.utils.decompose_NWP`,
```
sidekock, line 133: Has this been touched in any way? I would assume that the memory efficiency here could also be applied in the decompose_NWP function?

pysteps/blending/steps.py
```diff
@@ -557,22 +572,19 @@ def forecast(
...
-    # 2.2 If necessary, decompose (NWP) model forecasts and stack cascades
-    (
-        precip_models_cascade,
-        mu_models,
-        sigma_models,
-        precip_models_pm,
-    ) = _compute_cascade_decomposition_nwp(
-        precip_models, bp_filter, decompositor, recompositor, fft, domain
-    )
+    # 2.2 If necessary, recompose (NWP) model forecasts
```
sidekock, line 575: Does this mean that users who provided the decomposed NWP will then have it recomposed again? Why is this only done if the ndim =/= 4? If this check of ndims != 4 is to assume that you are in the decomposed case, I feel this is not very elegant. Would it not better be to have an extra variable for the user to finetune their call of the function where they have to determine this with the default being the old approach?

pysteps/blending/steps.py
```diff
@@ -557,22 +572,19 @@ def forecast(
...
+        precip_models_cascade = precip_models
+        precip_models = _compute_cascade_recomposition_nwp(precip_models, recompositor)
 
     # 2.3 Check for zero input fields in the radar and NWP data.
     zero_precip_radar = blending.utils.check_norain(precip, precip_thr, norain_thr)
     # The norain fraction threshold used for nwp is the default value of 0.0,
     # since nwp does not suffer from clutter.
     zero_model_fields = blending.utils.check_norain(
-        precip_models_pm, precip_thr, norain_thr
+        precip_models, precip_thr, norain_thr
```
sidekock, line 587: Why this name change? pm is only for the probability matching, or was the pm the only case of recombining the cascades, in which case this name is indeed better

pysteps/blending/steps.py
```diff
@@ -633,37 +645,41 @@ def forecast(
...
-            precip,
-            mu_models,
-            sigma_models,
-        )
+            done = False
+            for t in timesteps:
+                if done:
+                    break
+                for j in range(precip_models.shape[0]):
+                    # Look for a timestep and member with rain so that we have a sensible decomposition
```
sidekock: Would move this comment to the for t in timesteps

pysteps/blending/steps.py
```diff
@@ -633,37 +645,41 @@ def forecast(
...
+                            output_domain=domain,
+                            normalize=True,
+                            compute_stats=True,
+                            compact_output=True,
+                        )["cascade_levels"]
+                        precip_cascade[~np.isfinite(precip_cascade)] = np.nanmin(
+                            precip_models_cascade_temp
+                        )
+                        done = True
+                        break
```
sidekock, line 675: Am I correct that you are looping over all members for each time until you find a member with rain? Are you just doing this change the infs with the minimal value in this field? Is this minimal value in the found field necessarily the minimal value of precipitation in any field?

pysteps/blending/steps.py
```diff
@@ -809,23 +825,80 @@ def forecast(
             if measure_time:
                 starttime = time.time()
 
+            if precip_models_cascade is not None:
```
sidekock, line 828: I dont think precip_models_cascade is ever initialized, al I wrong in this?

pysteps/blending/steps.py
```diff
@@ -809,23 +825,80 @@ def forecast(
...
+                        decompositor(
+                            precip_models[0, t, :, :],
+                            bp_filter=bp_filter,
+                            fft_method=fft,
+                            output_domain=domain,
+                            normalize=True,
+                            compute_stats=True,
+                            compact_output=True,
+                        )
+                    ]
```
sidekock, line 842: Have you timed how long this takes approx?

mats-knmi:
@sidekock
- Yes both options should still work as intended. (The unit tests succeed for both options).
- The performance increases are not related to speed, just reduction of memory usage. In general I expect the memory usage to be lower using both options, but specifically when running with non-decomposed model data, as that allows you to input a lot less data (7x less data with 7 cascade levels).
- No I haven't done any analysis, but this should not impact a lot as all computations are still done with float64. It is just that you pass in the data as float32, meaning that it takes up less memory. Float32 has around 7 decimal digits of precision so it should be plenty accurate for storage.
- Yes running the blending this way has a slight performance decrease (around 5%, or 10 seconds) is what we experienced.

All of this, for us, is more than worth it. We run the blending with 12 hours lead time 5 minute time step and 20 members. This used to mean that we had to read 100+ GB of decomposed model data into memory, before even starting the blending. This is now reduced to around 20 GB.

pysteps/blending/steps.py
```diff
@@ -2043,78 +2116,24 @@ def _compute_cascade_decomposition_radar(
     return R_c, mu_extrapolation, sigma_extrapolation
 
 
-def _compute_cascade_decomposition_nwp(
```
sidekock, line 2046: Is this just no longer necessary because this was added in the code earlier?

pysteps/blending/steps.py
```diff
@@ -809,23 +825,80 @@ def forecast(
...
+            )
+            sigma_models_temp = np.array(
+                [decomp["stds"] for decomp in decomp_precip_models]
+            )
+
+            # 2.3.4 Check if the NWP fields contain nans or infinite numbers. If so,
+            # fill these with the minimum value present in precip (corresponding to
+            # zero rainfall in the radar observations)
+            (
+                precip_models_cascade_temp,
```
sidekock, line 872: are all these temp names the best here?

sidekock (commented):
@mats-knmi I did my best to go through all the changes. I added quite a few comments, most of them are just clarifications for me.

sidekock:
> @sidekock
> 
>     * Yes both options should still work as intended. (The unit tests succeed for both options).
> 
>     * The performance increases are not related to speed, just reduction of memory usage. In general I expect the memory usage to be lower using both options, but specifically when running with non-decomposed model data, as that allows you to input a lot less data (7x less data with 7 cascade levels).
> 
>     * No I haven't done any analysis, but this should not impact a lot as all computations are still done with float64. It is just that you pass in the data as float32, meaning that it takes up less memory. Float32 has around 7 decimal digits of precision so it should be plenty accurate for storage.
> 
>     * Yes running the blending this way has a slight performance decrease (around 5%, or 10 seconds) is what we experienced.
> 
> 
> All of this, for us, is more than worth it. We run the blending with 12 hours lead time 5 minute time step and 20 members. This used to mean that we had to read 100+ GB of decomposed model data into memory, before even starting the blending. This is now reduced to around 20 GB.

So you are doing this decomposition on the code now per time step and this is only impacting 10s for an entire forecast (for how long is this nowcast run? eg how many timesteps?) or 10s per step?

pysteps/blending/steps.py
mats-knmi: Well yes, but unfortunately this is how it was done before. Before these changes, both supplying decomposed and not decomposed data was already possible, and the check to see which of the 2 was supplied, was this ndim check.

An extra variable would have my preference as well, but that would be a breaking change with what it was before. Maybe you can take this up with your refactoring?

pysteps/blending/steps.py
mats-knmi: I don't know why it was called pm before, the only thing I know is that it (according to the comments in the code) would contain the recomposed model data. Recomposing the model data is the only way to get the input model data if you only supply the decomposed data, that is what I gathered from this code at least.

pysteps/blending/steps.py
sidekock: Good point, I did not know this was already possible! I will take this up as an issue for my refactoring

pysteps/blending/steps.py
sidekock: Okay, I guess this is fine with the name change you made. I will also take this into account for the refactoring :) pm should, in general no longer be a thing I guess

pysteps/blending/steps.py
sidekock: You already answered this in the previous comment (around 10s)

pysteps/blending/steps.py
mats-knmi: This was the most difficult part to get right (see also the extensive discussion Ruben and I had on this in this Pull Request). We need a zero value and apparently `np.nanmin(precip)` or `np.nanmin(precip_models)` are not correct, since the decomposed precip values are different than the not decomposed values. Therefore we need to decompose the model. To prevent just decomposing the entire model here (which would defeat the purpose of not pre-decomposing the model), we need to find a timestep and member that is representative for the entire model, decompose that and use the nanmin of that for the zero value. Or at least that is the only way I could see that we could realistically do this. I don't know how the decomposition works exactly, but I would expect that an image that has precipitation has a representative decomposition of which we could take the nanmin. This is based on the assumption that at some point somewhere in the domain there is no rain, but this assumption is made quite a bit in other parts of the pysteps code as well.

I would like to add that this code is supposed to solve the issue when `precip_cascade` is completely filled with nan values, but it doesn't explicitly check if this is the case, which seems weird to me. I also don't know if it is possible to get a `precip_cascade` value that is completely filled with nans, if you give a sensible value in `precip`.

pysteps/blending/steps.py
mats-knmi: On line 576 it is set to None and then on line 579 it is set to the input `precip_models` if the ndim is not equal to 4.

pysteps/blending/steps.py
mats-knmi: Yes, this gives different results on different machines, but locally I remember it took 10 seconds for 144 timesteps, for one model member. I added the multithreading option below to allow multiple members to be processed in parallel, so that you should be able to do this in 10 seconds regardless of your model size. As I said, operationally at KNMI these 10 seconds seem to hold true

pysteps/blending/steps.py
mats-knmi: I don't know, this is what it used to look like. Maybe you can find a better naming in the refactoring, probably something like `precip_models_cascade_timestep` would be more clear, since it is the part of precip models cascade that is valid for this timestep

pysteps/blending/steps.py
mats-knmi: This method is partially moved to within the nowcast loop, and partially replaced by the new method `_compute_cascade_recomposition_nwp`. This method used to decompose the nwp if not-decomposed data was provided and conversely recompose the nwp if decomposed data was provided. Now, not-decomposed data is decomposed "just in time" in the nowcast loop and decomposed data is recomposed using the new method, that is just a subset of what this method was.

Also I believe this method was another contributor to the high memory usage.

pysteps/blending/steps.py
mats-knmi: I did not touch these functions, but we don't use them as is currently in our code. We use an xarray version of these methods, that would introduce a breaking change, which I would like to implement in the pysteps v2 xarray version.

pysteps/blending/steps.py
mats-knmi: I am pretty sure the method `_check_inputs` checks this.

pysteps/blending/steps.py
mats-knmi: No as far as I can tell you should still be able to run this method the same as before and it should give the same end result. The only differences are that now you can supply these values as float32 without losing as much precision as the code casts it to float64 now before starting any computation and the memory consumption should be a lot lower.

pysteps/blending/steps.py
RubenImhoff: I can confirm that. :)

sidekock: Okay, thanks for this. Was not looking at the full code, only the diff

pysteps/blending/steps.py
sidekock: To me, this looks like an unfinished problem. I indeed see where this comes from, in an operational product, you should be ready for no radar data, but I feel like this is insufficient because, as far as I understand, this can give a lot of bias (a large precip field will give different min for the different cascades compared to a very intense convection event I would assume, please correct me if I am wrong here @RubenImhoff ). On the other hand, you do not expect the NWP to have huge variations in the 'weather type', so maybe it is fine. The only thing I worry about is that a very intense event with no other rain might give strange artefacts. I assume you used the same code as in "_fill_nans_infs_nwp_cascade" before?
I feel like we at least need to explore a full timestep before taking a min but that might be to computationally intensive (although we could also paralalize this)

pysteps/blending/steps.py
sidekock: So that means if dims is not 4 (the already decomposed case), it has been defined. If it is the other case (raw NWP), it is non, and thus, this decomposed. Is this correct?

pysteps/blending/steps.py
mats-knmi: No, this is actually the only part of the code that has some possible functional change. Since before at this point always the entire model field would have been decomposed, you could just take the nanmin over the entire decomposed forecast. But now this is not possible. Computationally it would indeed be less than ideal to decompose the entire forecast at this point in the code since the new changes were supposed to prevent this.

However, if this code is supposed to ONLY handle the case where the input (decomposed) precip field is completely filled with nans, I believe we should change the if statement here to check for that in stead of just checking for `zero_precip_radar`. If we do that, then I believe this case will not show up often (or at all) operationally, so it is not a big deal if it does take some extra time.

Also, this part of the code was never in `_fill_nans_infs_nwp_cascade` before. This was an additional check that was done separately to that method. [this](https://github.com/pySTEPS/pysteps/pull/435/files#diff-bb03b2c1b913f23015ec22a499edc0c635359f454aae3b9025296d80fd30b9e3L635-L642) is the part of the code this part is trying to replicate.

mats-knmi:
@sidekock 10 seconds longer over the entire forecast of 12 hours, with a 5 min timestep, so 144 timesteps. I believe this part might actually take like 20 seconds total, but the original `_compute_cascade_decomposition_nwp`, would also take 10 seconds in the old code, so bottom line is we lose around 10 seconds total.

pysteps/blending/steps.py
mats-knmi: Yes that is correct.

pysteps/blending/steps.py
sidekock: I think it is good for now but maybe it is a good thing to raise this as a separate issue? 
For now I see three possible solutions:
1) Leave as is, risking some possible issues due to early cutoff
2) Processing the whole NWP every run (very ram intensive and reduces computational efficiency
3) Pre-process NWP to get these no rain values once, and feed them into the model (can be used for multiple runs, depending on the frequency of the RUC).

I don't think we should fix this issue here but we should keep this noted somewhere that this can be optimised ...

sidekock:
> @sidekock 10 seconds longer over the entire forecast of 12 hours, with a 5 min timestep, so 144 timesteps. I believe this part might actually take like 20 seconds total, but the original `_compute_cascade_decomposition_nwp`, would also take 10 seconds in the old code, so bottom line is we lose around 10 seconds total.

Great, thanks! Good to have a benchmark, always easier to convince people that way :)
Thanks for answering all my questions, ill put my stamp on approval for this merge :) Then I can begin with the refactoring whenever it ends up in the master.

sidekock (approved):

pysteps/blending/steps.py
mats-knmi: Yes good point, do you want to open an issue or should I?

sidekock: Maybe we merge first and then create the issue on the master branch? Ill let you do it :)

mats-knmi: I cannot merge, but I can create an issue. Can you merge or should @dnerini do this?

sidekock: I can not merge either

dnerini: I can merge it

PR #436, opened 2024-10-09
Refactoring steps-nowcast

### LINKED ISSUE #395: Refactor the blending/steps.py nowcasting code
Start with low-hanging fruit:
- clearer variable names (compare to nowcasts/steps.py refactoring)
- potentially make the monolithic function more modular.

sidekock:
I started with the steps nowcast because that seemed more straightforward to tackle first. Since both methods should look alike, they must both be changed. I can't figure out how to run the test locally, so I will have to wait and see what the pull request test gives me as feedback.

This is a first suggestion how to split up the code but look forward to suggestions and improvments

### DIFF
--- a/pysteps/nowcasts/steps.py
+++ b/pysteps/nowcasts/steps.py
@@ -987,9 +987,20 @@ class StepsNowcaster:
             elapsed_time = time.time() - start_time
             print(f"{label} took {elapsed_time:.2f} seconds.")
 
-    def reset_states(self):
-        # TODO: Implement a method to reset the state of the nowcast object to make multiple iterations possible
-        pass
+    def reset_states_and_params(self):
+        """
+        Reset the internal state and parameters of the nowcaster to allow multiple forecasts.
+        This method resets the state and params to their initial conditions without reinitializing
+        the inputs like precip, velocity, time_steps, or config.
+        """
+        # Re-initialize the state and parameters
+        self.state = StepsNowcasterState()
+        self.params = StepsNowcasterParams()
+
+        # Reset time measurement variables
+        self.start_time_init = None
+        self.init_time = None
+        self.mainloop_time = None
 
 
 # Wrapper function to preserve backward compatibility
@@ -1256,6 +1267,6 @@ def forecast(
         precip, velocity, timesteps, steps_config=nowcaster_config
     )
     forecast_steps_nowcast = nowcaster.compute_forecast()
-    nowcaster.reset_states()
+    nowcaster.reset_states_and_params()
     # Call the appropriate methods within the class
     return forecast_steps_nowcast
--- a/pysteps/nowcasts/steps.py
+++ b/pysteps/nowcasts/steps.py
@@ -113,7 +113,7 @@ class StepsNowcaster:
         # Store inputs and optional parameters
         self.precip = precip
         self.velocity = velocity
-        self.tim_esteps = time_steps
+        self.time_steps = time_steps
 
         # Store the config data:
         self.config = steps_config
@@ -121,7 +121,7 @@ class StepsNowcaster:
         # Additional variables for internal state management
         self.fft = None
         self.bandpass_filter = None
-        self.extrapolator_method = None
+        self.extrapolation_method = None
         self.domain_mask = None
         self.precip_cascades = None
         self.gamma = None
@@ -140,14 +140,14 @@ class StepsNowcaster:
         self.fft_objs = None
         self.generate_noise = None
         self.decomposition_method = None
-        self.recomp_method = None
+        self.recomposition_method = None
         self.xy_coords = None
-
         self.precipitation_mean = None
-        # Initialize number of ensemble workers
-        self.num_ensemble_workers = min(
-            self.config.n_ens_members, self.config.num_workers
-        )
+        self.precip_mask = None
+        self.precip_mask_decomposed = None
+        self.war = None
+        self.struct = None
+        self.mask_rim = None
 
         # Store the state and params data:
         self.state = StepsNowcasterState()
@@ -222,17 +222,17 @@ class StepsNowcaster:
             precip,
             self.velocity,
             state,
-            self.tim_esteps,
+            self.time_steps,
             self.config.extrapolation_method,
             self._update_state,  # Reference to the update function
-            extrap_kwargs=self.extrapolation_kwargs,
+            extrap_kwargs=self.config.extrapolation_kwargs,
             velocity_pert_gen=self.velocity_perturbations,
             params=params,
             ensemble=True,
             num_ensemble_members=self.config.n_ens_members,
             callback=self.config.callback,
             return_output=self.config.return_output,
-            num_workers=self.num_ensemble_workers,
+            num_workers=self.params.num_ensemble_workers,
             measure_time=self.config.measure_time,
         )
 
@@ -256,8 +256,8 @@ class StepsNowcaster:
                 f"shape(precip)={self.precip.shape}, shape(velocity)={self.velocity.shape}"
             )
         if (
-            isinstance(self.tim_esteps, list)
-            and not sorted(self.tim_esteps) == self.tim_esteps
+            isinstance(self.time_steps, list)
+            and not sorted(self.time_steps) == self.time_steps
         ):
             raise ValueError("timesteps must be in ascending order")
         if np.any(~np.isfinite(self.velocity)):
@@ -301,15 +301,15 @@ class StepsNowcaster:
 
         # Handle None values for various kwargs
         if self.config.extrapolation_kwargs is None:
-            self.extrapolation_kwargs = {}
+            self.config.extrapolation_kwargs = {}
         if self.config.filter_kwargs is None:
-            self.filter_kwargs = {}
+            self.config.filter_kwargs = {}
         if self.config.noise_kwargs is None:
-            self.noise_kwargs = {}
+            self.config.noise_kwargs = {}
         if self.config.velocity_perturbation_kwargs is None:
-            self.velocity_perturbation_kwargs = {}
+            self.config.velocity_perturbation_kwargs = {}
         if self.config.mask_kwargs is None:
-            self.mask_kwargs = {}
+            self.config.mask_kwargs = {}
 
         print("Inputs validated and initialized successfully.")
 
@@ -355,21 +355,23 @@ class StepsNowcaster:
 
         print("Parameters")
         print("----------")
-        if isinstance(self.tim_esteps, int):
-            print(f"number of time steps:     {self.tim_esteps}")
+        if isinstance(self.time_steps, int):
+            print(f"number of time steps:     {self.time_steps}")
         else:
-            print(f"time steps:               {self.tim_esteps}")
+            print(f"time steps:               {self.time_steps}")
         print(f"ensemble size:            {self.config.n_ens_members}")
         print(f"parallel threads:         {self.config.num_workers}")
         print(f"number of cascade levels: {self.config.n_cascade_levels}")
         print(f"order of the AR(p) model: {self.config.ar_order}")
 
         if self.config.velocity_perturbation_method == "bps":
-            self.velocity_perturbation_parallel = self.velocity_perturbation_kwargs.get(
-                "p_par", noise.motion.get_default_params_bps_par()
+            self.velocity_perturbation_parallel = (
+                self.config.velocity_perturbation_kwargs.get(
+                    "p_par", noise.motion.get_default_params_bps_par()
+                )
             )
             self.velocity_perturbation_perpendicular = (
-                self.velocity_perturbation_kwargs.get(
+                self.config.velocity_perturbation_kwargs.get(
                     "p_perp", noise.motion.get_default_params_bps_perp()
                 )
             )
@@ -402,16 +404,16 @@ class StepsNowcaster:
         # Initialize the band-pass filter for the cascade decomposition
         filter_method = cascade.get_method(self.config.bandpass_filter_method)
         self.bandpass_filter = filter_method(
-            (M, N), self.config.n_cascade_levels, **(self.filter_kwargs or {})
+            (M, N), self.config.n_cascade_levels, **(self.config.filter_kwargs or {})
         )
 
         # Get the decomposition method (e.g., FFT)
-        self.decomposition_method, self.recomp_method = cascade.get_method(
+        self.decomposition_method, self.recomposition_method = cascade.get_method(
             self.config.decomposition_method
         )
 
         # Get the extrapolation method (e.g., semilagrangian)
-        self.extrapolator_method = extrapolation.get_method(
+        self.extrapolation_method = extrapolation.get_method(
             self.config.extrapolation_method
         )
 
@@ -442,7 +444,7 @@ class StepsNowcaster:
         else:
             self.mask_threshold = None
 
-        extrap_kwargs = self.extrapolation_kwargs.copy()
+        extrap_kwargs = self.config.extrapolation_kwargs.copy()
         extrap_kwargs["xy_coords"] = self.xy_coords
         extrap_kwargs["allow_nonfinite_values"] = (
             True if np.any(~np.isfinite(self.precip)) else False
@@ -452,7 +454,7 @@ class StepsNowcaster:
 
         def _extrapolate_single_field(precip, i):
             # Extrapolate a single precipitation field using the velocity field
-            return self.extrapolator_method(
+            return self.extrapolation_method(
                 precip[i, :, :],
                 self.velocity,
                 self.config.ar_order - i,
@@ -502,7 +504,7 @@ class StepsNowcaster:
 
             # Initialize the perturbation generator for the precipitation field
             self.perturbation_generator = init_noise(
-                self.precip, fft_method=self.fft, **self.noise_kwargs
+                self.precip, fft_method=self.fft, **self.config.noise_kwargs
             )
 
             # Handle noise standard deviation adjustments if necessary
@@ -627,12 +629,16 @@ class StepsNowcaster:
                 # Create random state for precipitation noise generator
                 rs = np.random.RandomState(self.config.seed)
                 self.random_generator_precip.append(rs)
-                self.seed = rs.randint(0, high=int(1e9))  # Update seed after generating
+                self.config.seed = rs.randint(
+                    0, high=int(1e9)
+                )  # Update seed after generating
 
                 # Create random state for motion perturbations generator
                 rs = np.random.RandomState(self.config.seed)
                 self.random_generator_motion.append(rs)
-                self.seed = rs.randint(0, high=int(1e9))  # Update seed after generating
+                self.config.seed = rs.randint(
+                    0, high=int(1e9)
+                )  # Update seed after generating
         else:
             self.random_generator_precip = None
             self.random_generator_motion = None
@@ -652,10 +658,10 @@ class StepsNowcaster:
             for j in range(self.config.n_ens_members):
                 kwargs = {
                     "randstate": self.random_generator_motion[j],
-                    "p_par": self.velocity_perturbation_kwargs.get(
+                    "p_par": self.config.velocity_perturbation_kwargs.get(
                         "p_par", self.velocity_perturbation_parallel
                     ),
-                    "p_perp": self.velocity_perturbation_kwargs.get(
+                    "p_perp": self.config.velocity_perturbation_kwargs.get(
                         "p_perp", self.velocity_perturbation_perpendicular
                     ),
                 }
@@ -687,12 +693,6 @@ class StepsNowcaster:
         else:
             self.precipitation_mean = None
 
-        self.precip_mask = None
-        self.precip_mask_decomposed = None
-        self.war = None
-        self.struct = None
-        self.mask_rim = None
-
         if self.config.mask_method is not None:
             self.mask_precip = self.precip[-1, :, :] >= self.config.precip_threshold
 
@@ -709,8 +709,8 @@ class StepsNowcaster:
 
             elif self.config.mask_method == "incremental":
                 # Get mask parameters
-                self.mask_rim = self.mask_kwargs.get("mask_rim", 10)
-                mask_f = self.mask_kwargs.get("mask_f", 1.0)
+                self.mask_rim = self.config.mask_kwargs.get("mask_rim", 10)
+                mask_f = self.config.mask_kwargs.get("mask_f", 1.0)
                 # Initialize the structuring element
                 self.struct = generate_binary_structure(2, 1)
                 # Expand the structuring element based on mask factor and timestep
@@ -783,7 +783,7 @@ class StepsNowcaster:
             "probmatching_method": self.config.probmatching_method,
             "precip": precip,
             "precip_thr": self.config.precip_threshold,
-            "recomp_method": self.recomp_method,
+            "recomp_method": self.recomposition_method,
             "struct": self.struct,
             "war": self.war,
         }
--- a/pysteps/nowcasts/steps.py
+++ b/pysteps/nowcasts/steps.py
@@ -25,8 +25,8 @@ from pysteps.postprocessing import probmatching
 from pysteps.timeseries import autoregression, correlation
 from pysteps.nowcasts.utils import compute_percentile_mask, nowcast_main_loop
 
-from dataclasses import field
-from typing import Optional, Dict, Any, Callable
+from dataclasses import dataclass, field
+from typing import Optional, Dict, Any, Callable, List
 
 try:
     import dask
@@ -36,6 +36,7 @@ except ImportError:
     DASK_IMPORTED = False
 
 
+@dataclass
 class StepsNowcasterConfig:
     n_ens_members: int = 24
     n_cascade_levels: int = 6
@@ -66,13 +67,51 @@ class StepsNowcasterConfig:
     return_output: bool = True
 
 
+@dataclass
+class StepsNowcasterParams:
+    fft: Any = None
+    bandpass_filter: Any = None
+    decomposition_method: Any = None
+    recomposition_method: Any = None
+    noise_generator: Optional[callable] = None
+    perturbation_generator: Optional[callable] = None
+    noise_std_coeffs: Optional[np.ndarray] = None
+    ar_model_coefficients: Optional[np.ndarray] = None
+    domain_mask: Optional[np.ndarray] = None
+    structuring_element: Optional[np.ndarray] = None
+    precipitation_mean: Optional[float] = None
+    wet_area_ratio: Optional[float] = None
+    num_workers: int = 1
+
+
+@dataclass
+class StepsNowcasterState:
+    precip_cascades: Optional[List[List[np.ndarray]]] = field(default_factory=list)
+    precip_decomposed: Optional[List[Dict[str, Any]]] = field(default_factory=list)
+    mask_precip: Optional[np.ndarray] = None
+    random_generator_precip: Optional[List[np.random.RandomState]] = field(
+        default_factory=list
+    )
+    random_generator_motion: Optional[List[np.random.RandomState]] = field(
+        default_factory=list
+    )
+    velocity_perturbations: Optional[List[callable]] = field(default_factory=list)
+    fft_objects: Optional[List[Any]] = field(default_factory=list)
+
+
 class StepsNowcaster:
-    def __init__(self, precip, velocity, timesteps, steps_config, **kwargs):
-        self.config = steps_config
+    def __init__(self, precip, velocity, timesteps, steps_config):
         # Store inputs and optional parameters
         self.precip = precip
         self.velocity = velocity
         self.timesteps = timesteps
+
+        # Store the config data:
+        self.config = steps_config
+
+        # Store the state and params data:
+        self.state = StepsNowcasterState()
+        self.params = StepsNowcasterParams()
 
         # Additional variables for internal state management
         self.fft = None
@@ -1201,7 +1240,7 @@ def forecast(
     nowcaster = StepsNowcaster(
         precip, velocity, timesteps, steps_config=nowcaster_config
     )
-    forecast = nowcaster.compute_forecast()
+    forecast_steps_nowcast = nowcaster.compute_forecast()
     nowcaster.reset_states()
     # Call the appropriate methods within the class
-    return forecast
+    return forecast_steps_nowcast
--- a/pysteps/nowcasts/steps.py
+++ b/pysteps/nowcasts/steps.py
@@ -25,6 +25,9 @@ from pysteps.postprocessing import probmatching
 from pysteps.timeseries import autoregression, correlation
 from pysteps.nowcasts.utils import compute_percentile_mask, nowcast_main_loop
 
+from dataclasses import dataclass, field
+from typing import Optional, Dict, Any, Callable
+
 try:
     import dask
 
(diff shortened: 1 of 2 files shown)


sidekock:
I have a question about the Codacy Static Code Analysis. It sais the method _check_inputs is to complex while this has barely been touched since the last version, I moved a few extra lines here because that just makes more sense. Should I split up that part of the code to remove the error or not? 
image

dnerini:
hi @sidekock , first of all, thanks for taking up this heroic task :-)

> can't figure out how to run the test locally

which part? to run the test locally you need to setup the test data correctly, see https://pysteps.readthedocs.io/en/stable/user_guide/example_data.html#example-data and https://pysteps.readthedocs.io/en/stable/developer_guide/test_pysteps.html

> I have a question about the Codacy Static Code Analysis. It sais the method _check_inputs is to complex while this has barely been touched since the last version, I moved a few extra lines here because that just makes more sense. Should I split up that part of the code to remove the error or not?

I agree with you, _check_inputs seems well in line with the complexity of the other methods, and feels like breaking it down into smaller pieces would somewhat overcrow your class. I silenced this issue on codacy.

pysteps/nowcasts/steps.py
```diff
@@ -33,6 +33,863 @@
...
+        - label: A description of the part of the process being measured.
+        - start_time: The timestamp when the process started (from time.time()).
+        """
+        if self.measure_time:
+            elapsed_time = time.time() - start_time
+            print(f"{label} took {elapsed_time:.2f} seconds.")
+
+
+# Wrapper function to preserve backward compatibility
 @deprecate_args({"R": "precip", "V": "velocity", "R_thr": "precip_thr"}, "1.8.0")
```
dnerini, line 1217: Looks like it's about time to deprecate these arguments. Would you do the honors?

sidekock: I will do this once I have also changed the blending/steps. This gives me some grasp of the old names to keep consistency. Once both are fully refactored, this will be the first thing on the list. 
I first wanted some feedback regarding the splitting of the steps function before I tackle the blending part because that code is quite a bit more complicated.

sidekock:
@mats-knmi  @RubenImhoff  Could you take a look at this and give feedback? That way I can go on to the blending version. I don't want to start working on that one if this one is not in a finalized state.

RubenImhoff:
@sidekock, I will try to do so asap.

pysteps/nowcasts/steps.py
```diff
@@ -33,6 +33,863 @@
...
+        self.precip_thr = kwargs.get("precip_thr", None)
+        self.kmperpixel = kwargs.get("kmperpixel", None)
+        self.timestep = kwargs.get("timestep", None)
+        self.extrap_method = kwargs.get("extrap_method", "semilagrangian")
+        self.decomp_method = kwargs.get("decomp_method", "fft")
+        self.bandpass_filter_method = kwargs.get("bandpass_filter_method", "gaussian")
+        self.noise_method = kwargs.get("noise_method", "nonparametric")
+        self.noise_stddev_adj = kwargs.get("noise_stddev_adj", None)
+        self.ar_order = kwargs.get("ar_order", 2)
+        self.vel_pert_method = kwargs.get("vel_pert_method", "bps")
```
RubenImhoff:
```suggestion
        self.velocity_perturbation_method = kwargs.get("velocity_perturbation_method", "bps")
```
If this works for you, too, then we should change it in the rest of the code as well.

pysteps/nowcasts/steps.py
```diff
@@ -33,6 +33,863 @@
...
+        self.conditional = kwargs.get("conditional", False)
+        self.probmatching_method = kwargs.get("probmatching_method", "cdf")
+        self.mask_method = kwargs.get("mask_method", "incremental")
+        self.seed = kwargs.get("seed", None)
+        self.num_workers = kwargs.get("num_workers", 1)
+        self.fft_method = kwargs.get("fft_method", "numpy")
+        self.domain = kwargs.get("domain", "spatial")
+        self.extrap_kwargs = kwargs.get("extrap_kwargs", None)
```
RubenImhoff:
```suggestion
        self.extrapolation_kwargs = kwargs.get("extrapolation_kwargs", None)
```
If this works for you, too, then we should change it in the rest of the code as well.

pysteps/nowcasts/steps.py
```diff
@@ -33,6 +33,863 @@
...
+        self.filter_kwargs = kwargs.get("filter_kwargs", None)
+        self.noise_kwargs = kwargs.get("noise_kwargs", None)
+        self.vel_pert_kwargs = kwargs.get("vel_pert_kwargs", None)
```
RubenImhoff:
```suggestion
        self.velocity_perturbation_kwargs = kwargs.get("velocity_perturbation_kwargs", None)
```
If this works for you, too, then we should change it in the rest of the code as well.

pysteps/nowcasts/steps.py
```diff
@@ -33,6 +33,863 @@
...
+        self.mask_kwargs = kwargs.get("mask_kwargs", None)
+        self.measure_time = kwargs.get("measure_time", False)
+        self.callback = kwargs.get("callback", None)
+        self.return_output = kwargs.get("return_output", True)
+
+        # Additional variables for internal state management
+        self.fft = None
+        self.bp_filter = None
```
RubenImhoff:
```suggestion
        self.bandpass_filter = None
```
It is bandpass, right? 
If this works for you, too, then we should change it in the rest of the code as well.

pysteps/nowcasts/steps.py
```diff
@@ -33,6 +33,863 @@
...
+        self.pert_gen = None
+        self.noise_std_coeffs = None
+        self.randgen_prec = None
+        self.randgen_motion = None
+        self.velocity_perturbations = None
+        self.precip_forecast = None
+        self.mask_prec = None
+        self.mask_thr = None
+        self.precip_decomp = None
+        self.vp_par = None
```
RubenImhoff:
```suggestion
        self.velocity_pertubation_parallel = None
```
If this works for you, too, then we should change it in the rest of the code as well.

pysteps/nowcasts/steps.py
```diff
@@ -33,6 +33,863 @@
...
+        self.vp_perp = None
```
RubenImhoff:
```suggestion
        self.velocity_perturbation_perpendicular = None
```
If this works for you, too, then we should change it in the rest of the code as well.

pysteps/nowcasts/steps.py
```diff
@@ -33,6 +33,863 @@
...
+        self.phi = None
+        self.pert_gen = None
+        self.noise_std_coeffs = None
+        self.randgen_prec = None
+        self.randgen_motion = None
+        self.velocity_perturbations = None
+        self.precip_forecast = None
+        self.mask_prec = None
+        self.mask_thr = None
+        self.precip_decomp = None
```
RubenImhoff:
```suggestion
        self.precip_decomposed = None
```
If this works for you, too, then we should change it in the rest of the code as well.

pysteps/nowcasts/steps.py
```diff
@@ -33,6 +33,863 @@
...
+        self.precip_cascades = None
+        self.gamma = None
+        self.phi = None
+        self.pert_gen = None
+        self.noise_std_coeffs = None
+        self.randgen_prec = None
+        self.randgen_motion = None
+        self.velocity_perturbations = None
+        self.precip_forecast = None
+        self.mask_prec = None
```
RubenImhoff:
```suggestion
        self.mask_precip = None
```
If this works for you, too, then we should change it in the rest of the code as well.

pysteps/nowcasts/steps.py
```diff
@@ -33,6 +33,863 @@
...
+        self.mask_thr = None
```
RubenImhoff:
```suggestion
        self.mask_threshold = None
```
If this works for you, too, then we should change it in the rest of the code as well.

pysteps/nowcasts/steps.py
```diff
@@ -33,6 +33,863 @@
...
+
+        # Additional variables for internal state management
+        self.fft = None
+        self.bp_filter = None
+        self.extrapolator_method = None
+        self.domain_mask = None
+        self.precip_cascades = None
+        self.gamma = None
+        self.phi = None
+        self.pert_gen = None
```
RubenImhoff:
```suggestion
        self.perturbation_generator = None
```
I'm not sure if this is the right name, what do you think?
If this works for you, too, then we should change it in the rest of the code as well.

pysteps/nowcasts/steps.py
```diff
@@ -33,6 +33,863 @@
...
+        self.noise_std_coeffs = None
+        self.randgen_prec = None
```
RubenImhoff: What should the full/long name be? Good to use longer names now that we are refactoring anyway.

pysteps/nowcasts/steps.py
```diff
@@ -33,6 +33,863 @@
...
+        self.randgen_motion = None
```
RubenImhoff: What should the full/long name be? Good to use longer names now that we are refactoring anyway.

pysteps/nowcasts/steps.py
```diff
@@ -33,6 +33,863 @@
...
+        # Store inputs and optional parameters
+        self.precip = precip
+        self.velocity = velocity
+        self.timesteps = timesteps
+        self.n_ens_members = kwargs.get("n_ens_members", 24)
+        self.n_cascade_levels = kwargs.get("n_cascade_levels", 6)
+        self.precip_thr = kwargs.get("precip_thr", None)
+        self.kmperpixel = kwargs.get("kmperpixel", None)
+        self.timestep = kwargs.get("timestep", None)
+        self.extrap_method = kwargs.get("extrap_method", "semilagrangian")
```
RubenImhoff:
```suggestion
        self.extrapolation_method = kwargs.get("extrapolation_method", "semilagrangian")
```
If this works for you, too, then we should change it in the rest of the code as well.

pysteps/nowcasts/steps.py
```diff
@@ -33,6 +33,863 @@
...
+        self.decomp_method = kwargs.get("decomp_method", "fft")
```
RubenImhoff:
```suggestion
        self.decomposition_method = kwargs.get("decomposition_method", "fft")
```
If this works for you, too, then we should change it in the rest of the code as well.

pysteps/nowcasts/steps.py
```diff
@@ -33,6 +33,863 @@
...
 
+class StepsNowcaster:
+    def __init__(self, precip, velocity, timesteps, **kwargs):
+        # Store inputs and optional parameters
+        self.precip = precip
+        self.velocity = velocity
+        self.timesteps = timesteps
+        self.n_ens_members = kwargs.get("n_ens_members", 24)
+        self.n_cascade_levels = kwargs.get("n_cascade_levels", 6)
+        self.precip_thr = kwargs.get("precip_thr", None)
```
RubenImhoff:
```suggestion
        self.precip_threshold = kwargs.get("precip_threshold", None)
```
If this works for you, too, then we should change it in the rest of the code as well.

pysteps/nowcasts/steps.py
```diff
@@ -33,6 +33,863 @@
...
+    def compute_forecast(self):
+        """
+        Main loop for nowcast ensemble generation. This handles extrapolation,
+        noise application, autoregressive modeling, and recomposition of cascades.
+        """
+        self._check_inputs()
+        self._print_forecast_info()
+        # Measure time for initialization
+        if self.measure_time:
+            self.start_time_init = time.time()
```
RubenImhoff: Should we add a test where measure_time is True (and False)? That would probably tick the box for the test coverage.

pysteps/nowcasts/steps.py
```diff
@@ -33,6 +33,863 @@
...
+        self._perform_extrapolation()
+        self._apply_noise_and_ar_model()
+        self._initialize_velocity_perturbations()
+        self._initialize_precipitation_mask()
+        self._initialize_fft_objects()
+        # Measure and print initialization time
+        if self.measure_time:
+            self._measure_time("Initialization", self.start_time_init)
+
+        # RUn the main nowcast loop
```
RubenImhoff:
```suggestion
        # Run the main nowcast loop
```

pysteps/nowcasts/steps.py
```diff
@@ -33,6 +33,863 @@
...
+        self.precip_forecast = [[] for _ in range(self.n_ens_members)]
+
+        if self.probmatching_method == "mean":
+            self.mu_0 = np.mean(
+                self.precip[-1, :, :][self.precip[-1, :, :] >= self.precip_thr]
+            )
+        else:
+            self.mu_0 = None
+
+        self.precip_m = None
```
RubenImhoff: Should this be:
```suggestion
        self.precip_mask = None
```
?

pysteps/nowcasts/steps.py
```diff
@@ -33,6 +33,863 @@
...
+        self.precip_m_d = None
```
RubenImhoff: Can we also rename this one?

RubenImhoff (changes requested):
Hi @sidekock, super nice work! Thanks a lot for picking up this tremendous task. I like that we use a class now, it is much cleaner this way. 

I think all looks good, but I've made some suggestions to change some variable names to a longer name.  

In addition, perhaps `utils.nowcast_main_loop` also can use a refactor based on the refactor here. In this `nowcast_main_loop`, the `subtimesteps` are also used. I remember that we were discussing about this during the hackathon. Perhaps this is the time to clean that up a bit, too? I might be the most difficult/unreadable part of the code.

pysteps/nowcasts/steps.py
sidekock: What about vp_prep? I would change that one too but dont know what the prep stands for

pysteps/nowcasts/steps.py
sidekock: We could to this but I don't know if it is worth it, its an internal method with almost no functionality...

pysteps/nowcasts/steps.py
sidekock: Is this a comment about all four of the shown names or only some?

sidekock: if it is about the last one: randoim_generator_precip ?

pysteps/nowcasts/steps.py
sidekock: randoim_generator_motion?

sidekock:
> Hi @sidekock, super nice work! Thanks a lot for picking up this tremendous task. I like that we use a class now, it is much cleaner this way.
> 
> I think all looks good, but I've made some suggestions to change some variable names to a longer name.
> 
> In addition, perhaps `utils.nowcast_main_loop` also can use a refactor based on the refactor here. In this `nowcast_main_loop`, the `subtimesteps` are also used. I remember that we were discussing about this during the hackathon. Perhaps this is the time to clean that up a bit, too? I might be the most difficult/unreadable part of the code.

I agree this might be worth looking at but the thing I am a bit worried about is the that this loop is used for all nowcasters so minor changes could affect the code in a lot of different places. I was thinking to tackle the loop in the blending first (since that only affect one place) and then transfer what I learned back. Its a bit of double work but it seems easier to debug that way what works and does not work. Any feedback on the way I split up the code?

pysteps/nowcasts/steps.py
RubenImhoff: That works!

pysteps/nowcasts/steps.py
```diff
@@ -33,6 +33,873 @@
...
+        if self.velocity_perturbation_method is not None:
+            init_vel_noise, generate_vel_noise = noise.get_method(
+                self.velocity_perturbation_method
+            )
+
+            self.velocity_perturbations = []
+            for j in range(self.n_ens_members):
+                kwargs = {
+                    "randstate": self.randgen_motion[j],
+                    "p_par": self.velocity_perturbation_kwargs.get(
```
RubenImhoff:
```suggestion
                    "parameter_parallel": self.velocity_perturbation_kwargs.get(
```

Last one, perhaps we should change these names too for consistency. Not sure if this suggestion is the best name, what do you think?

pysteps/nowcasts/steps.py
```diff
@@ -33,6 +33,873 @@
...
+                        "p_par", self.velocity_perturbation_parallel
+                    ),
+                    "p_perp": self.velocity_perturbation_kwargs.get(
```
RubenImhoff:
```suggestion
                    "parameter_perpendicular": self.velocity_perturbation_kwargs.get(
```
Not sure if this suggestion is the best name, what do you think?

pysteps/nowcasts/steps.py
sidekock: The problem is that the method "init_vel_noise()" expects these arguments to have a specific name. This method is used in multiple places, so another part of the code will crash if I change it. This shows the downside of working with all the dictionaries but I think at this point this is something we will have to accept... I could change all instances of this naming everywhere in the code, but I have no clue if this will break the code. If you feel we should do this, I will try. I can always revert to a previous commit

RubenImhoff: Good point. From that perspective, fine to leave this as is then.

RubenImhoff (approved):

dnerini:
@sidekock Should this be merged?

mats-knmi:
I am sorry for only commenting now. I have been busy with some other stuff. I will go over the changes and leave some input where I have it. If you don't want to hold off merging, that's fine since Ruben already reviewed this. Maybe you can then take my input along in the next steps of the refactoring

pysteps/nowcasts/steps.py
```diff
@@ -33,6 +33,873 @@
     DASK_IMPORTED = False
 
 
+class StepsNowcaster:
+    def __init__(self, precip, velocity, timesteps, **kwargs):
+        # Store inputs and optional parameters
+        self.precip = precip
+        self.velocity = velocity
+        self.timesteps = timesteps
+        self.n_ens_members = kwargs.get("n_ens_members", 24)
```
mats-knmi: In stead of using kwargs here, you could also make a `StepsNowcasterConfig`, which would be a dataclass. This way you can more explicitly specify to the user which additional arguments need to be provided and you could even specify type hints if you want.

pysteps/nowcasts/steps.py
```diff
@@ -33,6 +33,873 @@
...
+        self.precip_forecast = [[] for _ in range(self.n_ens_members)]
+
+        if self.probmatching_method == "mean":
+            self.mu_0 = np.mean(
+                self.precip[-1, :, :][self.precip[-1, :, :] >= self.precip_threshold]
+            )
+        else:
+            self.mu_0 = None
+
+        self.precip_mask = None
```
mats-knmi: I am not a huge fan of extending `self` with additional variables not defined in `__init__`. This can lead to confusing situations, where you can be unsure if the class has a property or not. Generally I would say that variables that are created as the nowcast runs, should just be passed along in stead of stored "globally". Ofcourse just passing everything along resulted in the 20 variables per functions we had before, which might be even less ideal. Maybe the variables that need to be passed along can be bundled in dataclasses as well, which again makes everything more explicit.

pysteps/nowcasts/steps.py
```diff
@@ -33,6 +33,873 @@
     DASK_IMPORTED = False
 
 
+class StepsNowcaster:
+    def __init__(self, precip, velocity, timesteps, **kwargs):
+        # Store inputs and optional parameters
+        self.precip = precip
```
mats-knmi: I am wondering if it would be possible to pass only the "scaffolding" here (what are now the kwargs) and then pass `precip`, `velocity` and `timesteps`, when you actually run the nowcast. This way you could create this class once and reuse it for every new iteration.

pysteps/nowcasts/steps.py
```diff
@@ -33,6 +33,873 @@
...
+            "pert_gen": self.perturbation_generator,
+            "probmatching_method": self.probmatching_method,
+            "precip": precip,
+            "precip_thr": self.precip_threshold,
+            "recomp_method": self.recomp_method,
+            "struct": self.struct,
+            "war": self.war,
+        }
+
+    def _update_state(self, state, params):
```
mats-knmi: The `state` could also be a dataclass. That way you could explicitly define all variables that are in the state.

pysteps/nowcasts/steps.py
```diff
@@ -33,6 +33,873 @@
...
+            "precip_m": self.precip_mask,
+            "precip_m_d": self.precip_mask_decomposed,
+            "randgen_prec": self.random_generator_precip,
+        }
+
+    def _initialize_params(self, precip):
+        """
+        Initialize the params dictionary used during the nowcast iteration.
+        """
+        return {
```
mats-knmi, line 978: The `params` could also be a dataclass

pysteps/nowcasts/steps.py
mats-knmi: Maybe, if I read the code correctly, you could create an empty `state` and `params` object (dataclass) in the `__init__` or at the start of `compute_forecast`. Then these objects can be passed along (or stored in self) to the methods that fill this state/params. This would group the currently existing variables a bit nicer together maybe?

pysteps/nowcasts/steps.py
```diff
@@ -33,6 +33,873 @@
...
+        self.probmatching_method = kwargs.get("probmatching_method", "cdf")
+        self.mask_method = kwargs.get("mask_method", "incremental")
+        self.seed = kwargs.get("seed", None)
+        self.num_workers = kwargs.get("num_workers", 1)
+        self.fft_method = kwargs.get("fft_method", "numpy")
+        self.domain = kwargs.get("domain", "spatial")
+        self.extrapolation_kwargs = kwargs.get("extrap_kwargs", None)
+        self.filter_kwargs = kwargs.get("filter_kwargs", None)
+        self.noise_kwargs = kwargs.get("noise_kwargs", None)
+        self.velocity_perturbation_kwargs = kwargs.get("vel_pert_kwargs", None)
```
mats-knmi: One small stylistic thing is that, as you did with the methods, you could also make all "private" variable names start with an underscore.

mats-knmi:
I have left my comments, let me know what you think. Again, sorry for taking so long to get back to you. Overall the code looks much more manageable now, compared to what it looked like before.

sidekock:
> @sidekock Should this be merged?

Let me use the feedback of Mats. I will contact you once it is ready to be merged.

pysteps/nowcasts/steps.py
sidekock: I understand the idea and that sounds very good indeed. If you need to run multiple, no need in setting up all these things. However, how would this work at runtime? I don't see how we can create a class once and then keep for multiple calls of the forecast function... Maybe its just to early in the morning :)

sidekock: So I see how we could do this but the old functionality of forecast() can not be changed. Only when people want to use the new framework more modularly would this be possible. In this case forecast would look something like this:
- init StepsNowcasterConfig
- init nowcaster (with precip, velocity, timesteps, and Config)
- nowcaster.compute_forecast()
This would enable people who run the forecast for multiple run times to define the StepsNowcasterConfig once and then loop over all times. This is the only solution I see right now but this would not be possible in the current forecast() method.

pysteps/nowcasts/steps.py
sidekock: I will do that but would like to see your feedback on this at is is the first time I am using these type hints. Also @dnerini would this be something you would like to see in this version of pysteps? I know @mats-knmi is doing this for the v2 version but want to check if this is still within the pysteps 'style'

pysteps/nowcasts/steps.py
sidekock: With private, do you mean everything needed under the hood and not explicitly used as an input or output? Just a check because of the quotations around private

pysteps/nowcasts/steps.py
sidekock: I agree, and this fits nicely with your previous suggestion. That way we have a dataclass which can be modified (because some of these are only calculated at run time) and can be retrieved with a simple get

pysteps/nowcasts/steps.py
mats-knmi: What I meant is that it is possible to do something like the snippet below when using this class. If `filter_kwargs` were renamed to `_filter_kwargs`, you would indicate to the user that this is not allowed.

The reason I put private in quotations is that just putting a single underscore before a variable does not enforce this. A user can choose to ignore the underscore (but a linter like pylint would warn them that this is not allowed).

It is also possible in python to start the variable name with 2 underscores (`__filter_kwargs`). In this case the python interpreter actually (somewhat) enforces the privateness of this variable, by doing something called "name mangling". This might even be preferred to the single underscores (also for the private methods). Personally I use double underscores as much as possible.

See also: https://stackoverflow.com/questions/1301346/what-is-the-meaning-of-single-and-double-underscore-before-an-object-name

```
nowcaster = StepsNowcaster(...)
nowcaster.filter_kwargs = new_filter_kwargs
```

pysteps/nowcasts/steps.py
mats-knmi: Yes the old `forecast()` method can indeed not be updated to this new way of running the code, but maybe we can deprecate the old `forecast()` method over time? Actually this could easily be done when the xarray v2 would be released.

As for how this would work, I was thinking something like this:
```
nowcaster = StepsNowcast(timesteps, StepsNowcastConfig(...))
nowcaster.compute_forecast(precip, velocity)
```
The `__init__` could actually already take care of some things that don't need `precip` and `velocity` like retrieving the oflow method to be used.

The only thing you need to be very careful with in this case is that you reset the state properly every time `compute_forecast` is called.

pysteps/nowcasts/steps.py
mats-knmi: Yeah I was wondering about type hints as well in the v2 version. Currently it does not really fit in the pysteps style, but it is possible to use them in some places and not others without breaking the code.

I was hoping we could slowly roll them out to more parts of the code and eventually make pysteps completely type hinted (also adding mypy to the github actions), but this would still be very far off.

For now I think it doesn't hurt sprinkling them in there where they make sense, but it does indeed mean that the style is somewhat different between different parts of the code. I am interested what @dnerini thinks of this.

pysteps/nowcasts/steps.py
sidekock: I am going step by step. This Config file is now added; please let me know if this is how you want me to continue or if I need to change the code. Currently, the config file is a class variable. All features from these variables are always accessed as follows: self.config.XXXX. That way, it is always clear what parameters are configuration parameters and which ones are parameters or states for the nowcaster. The next step is to add these two...

pysteps/nowcasts/steps.py
sidekock: This is done now! I might need to add typing and dataclass and typing as new optional dependicies if these are not in base python...

pysteps/nowcasts/steps.py
sidekock: I indeed missed this one... Currently I added it to the init. Next step is indeed to add the two suggested classes to remove the number of class variables

pysteps/nowcasts/steps.py
sidekock: I want to pick this up if all the other changes you suggested are implemented. We should also check this with @dnerini to see if this does not break the PySteps rules. It is in the official documentation, but I rarely see it in the actual code.

sidekock:
@mats-knmi could you look through the changes to give some feedback? I will look at the 'private' variables next week if I have time. In general, I think this approach is much cleaner compared to what we had before. We see what is changed and what type it is (state or param) and what is loaded (config, state or param) for each method. 

For the 'private' discussion, I have taken a small look at the StackOverflow page but do not yet see how I would use this at this moment. Most variables in the code are part of a dataclass so can I still make then private then?

pysteps/nowcasts/steps.py
```diff
@@ -33,6 +36,971 @@
     DASK_IMPORTED = False
 
 
+@dataclass
+class StepsNowcasterConfig:
```
mats-knmi, line 40: Since pysteps dropped support for python 3.9:
`Optional[float]` can be written as `float | None`
`Dict[str, Any]` can be written as `dict[str, Any]` and same goes for `List`

Doing this saves some imports from the typing module; you don't have to import `Optional`, `Dict` and `List`

pysteps/nowcasts/steps.py
```diff
@@ -33,6 +36,971 @@
...
+    )
+    random_generator_motion: Optional[List[np.random.RandomState]] = field(
+        default_factory=list
+    )
+    velocity_perturbations: Optional[List[callable]] = field(default_factory=list)
+    fft_objects: Optional[List[Any]] = field(default_factory=list)
+
+
+class StepsNowcaster:
+    def __init__(self, precip, velocity, time_steps, steps_config):
```
mats-knmi: The type hints in the dataclasses already help a lot. As an extra bonus these classes can then also be used as type hints for the methods that expect them.

Also, I believe this `__init__` method (or the class itself) should have a docstring indicating to the user what they are expected to provide as arguments, or is that done somewhere else currently?

```suggestion
    def __init__(self, precip, velocity, time_steps, steps_config: StepsNowcasterConfig):
```

mats-knmi:
Looks good, I added 2 small comments on the type hinting, but other than that I think this makes it a lot nicer indeed. Regarding the private discussion: 
Every method that currently starts with an underscore could start with 2 underscores and then they could still be called from within the class, but not as easily from outside. (e.g. `_nowcast_main` could be renamed to `__nowcast_main` and then be called normally as `self.__nowcast_main()`).
Every variable that is defined in `__init__` could be defined with 2 underscores and then still accessed normally within the class, but not as easily from outside. (e.g. `self.__state = StepsNowcasterState()` and then acccess it later like `self.__state.precip_forecast`)

sidekock:
> Looks good, I added 2 small comments on the type hinting, but other than that I think this makes it a lot nicer indeed. Regarding the private discussion: Every method that currently starts with an underscore could start with 2 underscores and then they could still be called from within the class, but not as easily from outside. (e.g. `_nowcast_main` could be renamed to `__nowcast_main` and then be called normally as `self.__nowcast_main()`). Every variable that is defined in `__init__` could be defined with 2 underscores and then still accessed normally within the class, but not as easily from outside. (e.g. `self.__state = StepsNowcasterState()` and then acccess it later like `self.__state.precip_forecast`)

Thanks for all the feedback @mats-knmi and @RubenImhoff. The new version of the code feels much more readable now! I made the last changes and will soon start on the real challenge: the blending code :)

sidekock:
@dnerini the Static code analysis is complaining again because I changed the name of the method "__check_inputs". Would it be possible to disable this once more? That way, all tests should pass now

pysteps/nowcasts/steps.py
mats-knmi: I think that the second part of this comment maybe still needs to be addressed? (part of) the docstring from the `forecast` method probably needs to be copied here so that when users use the class directly they know what to do with each parameter. Possibly the docstring can be cut into 2 pieces where the part that describes `precip`, `velocity` and `timesteps` can be moved here and the part that describes all other parameters could be a docstring for the `StepsNowcasterConfig` dataclass.

mats-knmi:
@sidekock I resolved all my comments that are fully resolved now, I left 2 open, since those are not fully resolved yet. 

I think the docstring comment is important to still address, but the one about passing `precip` and `velocity` and/or `timesteps` not at the `__init__` but at the `compute_forecast` is more of a stylistic preference.

pysteps/nowcasts/steps.py
sidekock: This should be done now

sidekock:
> @sidekock I resolved all my comments that are fully resolved now, I left 2 open, since those are not fully resolved yet.
> 
> I think the docstring comment is important to still address, but the one about passing `precip` and `velocity` and/or `timesteps` not at the `__init__` but at the `compute_forecast` is more of a stylistic preference.

The new naming is now applied and the documentation is provided.

sidekock:
@mats-knmi if all the requested changes are present, would you please approve the changes? Then I know I can start working on the blending part after the merge

mats-knmi:
Yes they are approved. ~~I wanted to approve them in github, but I can't find the button to do this. Not sure if I am just blind or if I am not authorized to approve stuff...~~ Nevermind I found the button. It is in a completely different place as to where I am used to in gitlab.

mats-knmi (approved):

sidekock:
@dnerini I think this is ready to be merged with the master branch

dnerini:
Fantastic work! 🙌

Issue #438, opened 2024-10-29
Make sure correct zerovalue is used for precip_models_cascade in blending code


mats-knmi:
During a recent performance improvement, there was some uncertainty on how to deal with computing the zerovalue for `precip_models_cascade`: https://github.com/pySTEPS/pysteps/pull/435#discussion_r1786151740. There is some code there that is supposed to get this zerovalue when `precip_cascade` is completely filled with nan's.

In the new code, it is not the case that the entire `precip_models_cascade` is stored in RAM at that time, so just taking the `nanmin` over that entire array is not possible anymore. I have solved it by looping over `precip_models` until we find a timestamp and member that has rain and decompose that and take the `nanmin` to get the zerovalue. This differs from the old code, but should not be functionally different.

When discussing this some other questions arose. Namely: since this part of the code is supposed to catch the case where `precip_cascade` is completely filled with nan's, why is this not checked? And is this the best way to get a zerovalue for `precip_cascade`?

This also ties in with 2 other issues: https://github.com/pySTEPS/pysteps/issues/396 and https://github.com/pySTEPS/pysteps/issues/384

PR #443, opened 2024-12-06
Refactor steps in blending code

### LINKED ISSUE #440: Refactor steps in blending code
Issue #395 should be done for the blending code aswel.

sidekock:
This is the refactored version of the blending code of steps, see issue #440


sidekock:
Hi reviewers,
The blending code is finally refactored. The following things still need to happen but I would like to do these during the review process:
- Documentation needs to be updated to the new structure (please let me know where more documentation is needed)
- @RubenImhoff  could you please take a look at the  # TODO's, I added some and some remain from the original version. I would like to remove most of these if possible. Some are still there for me to check I don't forget last checks.
Look forward to all the feedback!

sidekock:
Also, @ladc and I found a possible bug in the code that needs to be fixed. I added a test to the test files but commented it out for the moment as this would break the tests (also in the old version of the code)

RubenImhoff:
@sidekock, fantastic work! I'll try to give it a good review before the Christmas break.

dnerini:
> Also, @ladc and I found a possible bug in the code that needs to be fixed. I added a test to the test files but commented it out for the moment as this would break the tests (also in the old version of the code)

could you say a bit more about this bug @sidekock ? is it something that we should address in a separate bug fix?

sidekock:
> > Also, @ladc and I found a possible bug in the code that needs to be fixed. I added a test to the test files but commented it out for the moment as this would break the tests (also in the old version of the code)
> 
> could you say a bit more about this bug @sidekock ? is it something that we should address in a separate bug fix?

I should check it to be sure but don't have my computer with me now but is has to do with iterations over model numbers vs over duplicated model numbers. We just missed this issue in the tests because it is something that does not frequently happen. A possible reason this could fail is when you have a 2 NWP members but want to generate a blended nowcast with 4 members. This is a small bugfix I can do but I wanted to wait until this issue is tackled fully. It is one of the #TODO comments I added to the code so if you want you can take a look through the code.

The specific TODO is this one:
"TODO: check if j is the best accessor for this variable"
The type of test it failed (which is not yet present in the master branch) is this one:

n_models=2, timesteps=3, n_ens_members=4, n_cascade_levels=8, mask_method="incremental", probmatching_method="cdf", blend_nwp_members=True&False, weights_method="spn", decomposed_nwp=True, expected_n_ens_members=2, zero_radar=False, zero_nwp=False, smooth_radar_mask_range=0, resample_distribution=False),

.gitignore
```diff
@@ -91,3 +91,7 @@ venv.bak/
 
 # Mac OS Stuff
 .DS_Store
+
+# Running lcoal tests
```
mats-knmi:
```suggestion
# Running local tests
```

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2862 @@
 except ImportError:
     DASK_IMPORTED = False
 
+from dataclasses import dataclass, field
+from typing import Optional, List, Dict, Any, Callable, Union
```
mats-knmi: `Optional`, `List`, `Dict` and `Union` are not necessary to import for type hints since python 3.10.
`List` and `Dict` can be replaced with `list` and `dict`
and something like `Optional[Union[list, np.ndarray]]` could be written as `list | np.ndarray | None`

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2862 @@
...
+
+        self.__config = steps_blending_config
+
+        # Initialize Params and State
+        self.__params = StepsBlendingParams()
+        self.__state = StepsBlendingState()
+
+        # Additional variables for time measurement
+        self.__start_time_init = None
+        self.__zero_precip_time = None
```
mats-knmi, line 245: This variable appears to be unused

mats-knmi:
I have already noted some initial things I noticed, but other than that it looks really good, much more readable than before.
I will try to find some more time to look into this more in depth (it is a really big file to review haha).
Regarding docstrings: I think what we did for the steps code still makes sense here right? Adding a docstring to `StepsBlendingConfig` and one to `compute_forecast` to explain what all the variables do?

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2862 @@
...
-      :py:func:`pysteps.blending.utils.compute_store_nwp_motion`, and
-      :py:func:`pysteps.blending.utils.load_NWP`. However, if you have a lot of (NWP) model
-      members (e.g. 1 model member per nowcast member), this can lead to excessive memory
-      usage.
-
-      To further reduce memory usage, both this array and the ``velocity_models`` array
-      can be given as float32. They will then be converted to float64 before computations
-      to minimize loss in precision.
+# TODO: compare old and new version of the code, run a benchmark to compare the two
+# TODO: look at the documentation and try to improve it, lots of things are now combined together
```
RubenImhoff, line 71: Are you still planning to do that as part of this PR?

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2862 @@
...
+            final_blended_forecast_all_members_one_timestep = [
+                None for _ in range(self.__config.n_ens_members)
+            ]
 
-      lucaskanade/mch:
-      p_par  = [2.56338484, 0.3330941, -2.99714349]
-      p_perp = [1.31204508, 0.3578426, -1.02499891]
+            def worker(j):
+                # The state needs to be copied as a dataclass is not threadsafe in python
+                worker_state = deepcopy(self.__state)
```
RubenImhoff: So, does this mean the current solution is not thread safe?

pysteps/blending/steps.py
sidekock: Thanks Mats! I am clearly not that well versed in the typing in python and must have used an older resource when I was writing this. Ill update it later this week :)

pysteps/blending/steps.py
sidekock: Thanks, I should indeed use this one, currently that is a local variable :) I changed this in the new version

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2862 @@
...
 
-      Parameters fitted to the data (optical flow/domain):
+    def __blended_nowcast_main_loop(self):
+        """
+        Main nowcast loop that iterates through the ensemble members and time steps
+        to generate forecasts.
+        """
+        ###
+        # 8. Start the forecasting loop
+        ###
```
RubenImhoff, line 308: The comments above can be removed.

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2862 @@
...
+            def worker(j):
+                # The state needs to be copied as a dataclass is not threadsafe in python
+                worker_state = deepcopy(self.__state)
+                self.__determine_skill_for_next_timestep(t, j, worker_state)
+                self.__determine_weights_per_component(worker_state)
+                self.__regress_extrapolation_and_noise_cascades(j, worker_state)
+                self.__perturb_blend_and_advect_extrapolation_and_noise_to_current_timestep(
+                    t, j, worker_state
+                )
+                # 8.5 Blend the cascades
```
RubenImhoff, line 352:
```suggestion
                # Blend the cascades
```

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2862 @@
...
+                final_blended_forecast_single_member = []
+                for t_sub in self.__state.subtimesteps:
+                    # TODO: does it make sense to use sub time steps - check if it works?
```
RubenImhoff, line 355: Is this TODO for another PR? I think we should still make a decision on the usage of sub timesteps, but that is probably better for a separate issue and PR.

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2862 @@
...
-    filter_method = cascade.get_method(bandpass_filter_method)
-    bp_filter = filter_method(precip_shape, n_cascade_levels, **filter_kwargs)
+        print("Blended nowcast components initialized successfully.")
 
-    decompositor, recompositor = cascade.get_method(decomp_method)
+    def __prepare_radar_and_NWP_fields(self):
+        # determine the precipitation threshold mask
+        if self.__config.conditional:
+            # TODO: is this logical_and correct here? Now only those places where precip is in all images is saved?
+            self.__params.mask_threshold = np.logical_and.reduce(
```
RubenImhoff, line 700: I believe we do the same in `nowcast/steps.py`, right? I get where your question comes from, and I have to say that I'm not certain myself either. @dnerini, what do you think?

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2862 @@
...
 
-    extrapolator = extrapolation.get_method(extrap_method)
+        # we need to know the zerovalue of precip to replace the mask when decomposing after
+        # extrapolation
+        self.__params.precip_zerovalue = np.nanmin(self.__precip)
 
-    x_values, y_values = np.meshgrid(
-        np.arange(precip.shape[2]), np.arange(precip.shape[1])
-    )
+        # 1. Start with the radar rainfall fields. We want the fields in a
```
RubenImhoff, line 713:
```suggestion
        Start with the radar rainfall fields. We want the fields in a
```
My suggestion is to remove the numbering of the processing steps, as they are spread over multiple functions now. Instead, it would help the user to have good docstrings throughout the main functions. :)

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2862 @@
...
+            self.__precip[i, ~np.isfinite(self.__precip[i, :])] = np.nanmin(
+                self.__precip[i, :]
+            )
 
-    # 1. Start with the radar rainfall fields. We want the fields in a
-    # Lagrangian space
-    precip = _transform_to_lagrangian(
-        precip, velocity, ar_order, xy_coords, extrapolator, extrap_kwargs, num_workers
-    )
+        # 2. Perform the cascade decomposition for the input precip fields and,
```
RubenImhoff, line 758:
```suggestion
        # Perform the cascade decomposition for the input precip fields and,
```

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2862 @@
...
+        # if necessary, for the (NWP) model fields
+        # 2.1 Compute the cascade decompositions of the input precipitation fields
```
RubenImhoff, line 760:
```suggestion
```
Sentence was present twice.

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2862 @@
...
+            precip_forecast_decomp, self.__config.n_cascade_levels
+        )
 
-    # 2.2 If necessary, recompose (NWP) model forecasts
-    precip_models_cascade = None
+        precip_forecast_decomp = precip_forecast_decomp[-1]
+        self.__state.mean_extrapolation = np.array(precip_forecast_decomp["means"])
+        self.__state.std_extrapolation = np.array(precip_forecast_decomp["stds"])
+
+        # 2.2 If necessary, recompose (NWP) model forecasts
```
RubenImhoff, line 786:
```suggestion
        # If necessary, recompose (NWP) model forecasts
```

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2862 @@
...
+            self.__precip_models = np.stack(temp_precip_models)
 
-    # 2.3 Check for zero input fields in the radar and NWP data.
-    zero_precip_radar = blending.utils.check_norain(precip, precip_thr, norain_thr)
-    # The norain fraction threshold used for nwp is the default value of 0.0,
-    # since nwp does not suffer from clutter.
-    zero_model_fields = blending.utils.check_norain(
-        precip_models, precip_thr, norain_thr
-    )
+        # 2.3 Check for zero input fields in the radar and NWP data.
```
RubenImhoff, line 805:
```suggestion
        # Check for zero input fields in the radar and NWP data.
```

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2862 @@
...
+                )
             else:
-                return outarr
+                return precip_forecast_all_members_all_times
         else:
             return None
 
-    else:
+    def __prepare_nowcast_for_zero_radar(self):
         # 2.3.3 If zero_precip_radar, make sure that precip_cascade does not contain
```
RubenImhoff, line 871:
```suggestion
        # If zero_precip_radar, make sure that precip_cascade does not contain
```

pysteps/blending/steps.py
sidekock: Should be fixed in the new version

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2862 @@
...
-                            fft_method=fft,
-                            output_domain=domain,
-                            normalize=True,
-                            compute_stats=True,
-                            compact_output=True,
-                        )["cascade_levels"]
-                        precip_cascade[~np.isfinite(precip_cascade)] = np.nanmin(
-                            precip_models_cascade_temp
+
+        # Look for a timestep and member with rain so that we have a sensible decomposition
```
RubenImhoff, line 874: What happens if we don't find a suitable member?

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2862 @@
...
+                        compute_stats=True,
+                        compact_output=True,
+                    )["cascade_levels"]
+                    self.__state.precip_cascades[
+                        ~np.isfinite(self.__state.precip_cascades)
+                    ] = np.nanmin(precip_models_cascade_timestep)
+                    done = True
+                    break
 
         # 2.3.5 If zero_precip_radar is True, only use the velocity field of the NWP
```
RubenImhoff, line 907:
```suggestion
        # If zero_precip_radar is True, only use the velocity field of the NWP
```

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2862 @@
...
-            velocity = np.mean(velocity, axis=0)
+        # Use the velocity from velocity_models at time step 0
+        self.__velocity = self.__velocity_models[:, 0, :, :, :].astype(
+            np.float64, copy=False
+        )
+        # Take the average over the first axis, which corresponds to n_models
+        # (hence, the model average)
+        self.__velocity = np.mean(self.__velocity, axis=0)
 
         # 3. Initialize the noise method.
```
RubenImhoff, line 918:
```suggestion
```

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2862 @@
...
         # If zero_precip_radar is True, initialize noise based on the NWP field time
```
RubenImhoff, line 919:
```suggestion
        # Initialize noise based on the NWP field time
```

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2862 @@
...
-            num_workers,
-            seed,
-        )
-        precip_noise_input = None
+            self.__params.perturbation_generator = None
+            self.__params.noise_generator = None
+            self.__params.noise_std_coeffs = None
 
+    def __estimate_ar_parameters_radar(self):
         # 4. Estimate AR parameters for the radar rainfall field
```
RubenImhoff, line 1008:
```suggestion
```
Already mentioned in the docstring of the function.

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2862 @@
...
+                GAMMA[i, :] = correlation.temporal_autocorrelation(
+                    self.__state.precip_cascades[i], mask=self.__params.mask_threshold
+                )
+
+        # Else, use standard values for the auto-correlations
+        else:
+            # Get the climatological lag-1 and lag-2 auto-correlation values from Table 2
+            # in `BPS2004`.
+            # Hard coded, change to own (climatological) values when present.
+            # TODO: add user warning here so users can be aware of this without reading the code?
```
RubenImhoff, line 1024: Should we build in a functionality so that users can provide the climatological values for GAMMA, e.g. through the `kwargs`?

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2862 @@
...
+        self.__params.PHI = np.empty(
+            (self.__config.n_cascade_levels, self.__config.ar_order + 1)
         )
+        for i in range(self.__config.n_cascade_levels):
+            self.__params.PHI[i, :] = autoregression.estimate_ar_params_yw(GAMMA[i, :])
 
+        nowcast_utils.print_ar_params(self.__params.PHI)
+
+    def __multiply_precip_cascade_to_match_ensemble_members(self):
         # 5. Repeat precip_cascade for n ensemble members
```
RubenImhoff, line 1081:
```suggestion
        """
        Repeat self.__state.precip_cascades for n ensemble members for the AR(p) model by using the p-1 last cascades
        """
```
The two comments below can then be removed.

sidekock:
> I have already noted some initial things I noticed, but other than that it looks really good, much more readable than before. I will try to find some more time to look into this more in depth (it is a really big file to review haha). Regarding docstrings: I think what we did for the steps code still makes sense here right? Adding a docstring to `StepsBlendingConfig` and one to `compute_forecast` to explain what all the variables do?
Indeed a very big file :) Ill do the docstrings one of the next days and try to tackle the duplicate comments at that point too

sidekock:
> I have already noted some initial things I noticed, but other than that it looks really good, much more readable than before. I will try to find some more time to look into this more in depth (it is a really big file to review haha). Regarding docstrings: I think what we did for the steps code still makes sense here right? Adding a docstring to `StepsBlendingConfig` and one to `compute_forecast` to explain what all the variables do?

Indeed a very big file :) Ill do the docstrings one of the next days and try to tackle the duplicate comments at that point too

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2858 @@
...
+                [
+                    self.__state.precip_cascades[i][-self.__config.ar_order :].copy()
+                    for i in range(self.__config.n_cascade_levels)
+                ]
+            ]
+            * self.__config.n_ens_members
         )
 
+    def __initialize_random_generators(self):
         # 6. Initialize all the random generators and prepare for the forecast loop
```
RubenImhoff, line 1096:
```suggestion
```

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2858 @@
...
-        # iterate each time step
-        for t, subtimestep_idx in enumerate(timesteps):
-            if timestep_type == "list":
-                subtimesteps = [original_timesteps[t_] for t_ in subtimestep_idx]
+        else:
+            if self.__precip_models.shape[0] == 1:
+                decomp_precip_models = [
+                    self.__params.decomposition_method(
+                        self.__precip_models[0, t, :, :],
+                        bp_filter=self.__params.bandpass_filter,
```
RubenImhoff, line 1273:
```suggestion
                        bandpass_filter=self.__params.bandpass_filter,
```

Here and further, shall we use bandpass_filter instead of bp_filter (even though it stays within this function)?

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2858 @@
...
+            [decomp["stds"] for decomp in decomp_precip_models]
+        )
 
-            if is_nowcast_time_step:
-                print(
-                    f"Computing nowcast for time step {t}... ",
-                    end="",
-                    flush=True,
-                )
+        # 2.3.4 Check if the NWP fields contain nans or infinite numbers. If so,
```
RubenImhoff, line 1306:
```suggestion
        # Check if the NWP fields contain nans or infinite numbers. If so,
```

RubenImhoff: Should we also check for negative numbers by the way? It's rare, but it can happen in NWP.

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2858 @@
...
+        # respectively 0.0 and 1.0
+        self.__state.mean_models_timestep[
+            ~np.isfinite(self.__state.mean_models_timestep)
+        ] = 0.0
+        self.__state.std_models_timestep[
+            ~np.isfinite(self.__state.std_models_timestep)
+        ] = 0.0
+
+    def __find_nowcast_NWP_combination(self, t):
+        # 8.1.1 Before calling the worker for the forecast loop, determine which (NWP)
```
RubenImhoff, line 1333:
```suggestion
```
And also the docstrings below can be removed, as it is already mentioned in the function description (which is good!).

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2858 @@
...
-                                decompositor,
-                                bp_filter=bp_filter,
-                                fft_method=fft,
-                                output_domain=domain,
-                                normalize=True,
-                                compute_stats=True,
-                                compact_output=True,
-                            ),
-                            list(precip_models[:, t, :, :]),
+                    if n_model_members == n_ens_members_min:
```
RubenImhoff: This if-statement seems redundant as it is already checked in line 1375 (`elif n_model_members == n_ens_members_min:`)

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2858 @@
...
+                                axis=0,
+                            )
                         )
 
-            precip_models_cascade_temp = np.array(
-                [decomp["cascade_levels"] for decomp in decomp_precip_models]
-            )
-            mu_models_temp = np.array(
-                [decomp["means"] for decomp in decomp_precip_models]
+        # TODO: is this not duplicate from part 2.3.5? If so, is it still needed here?
```
RubenImhoff: Looks like it to me. Maybe good to do to a check with and without it to ensure it gives the same result, but I don't (yet) see a reason why it should stay here.

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2858 @@
...
+            blending.clim.save_skill(
+                current_skill=self.__params.rho_nwp_models,
+                validtime=self.__issuetime,
+                outdir_path=self.__config.outdir_path_skill,
+                **self.__config.climatology_kwargs,
             )
-            sigma_models_temp = np.array(
-                [decomp["stds"] for decomp in decomp_precip_models]
+        if t > 0:
+            # 8.1.3 Determine the skill of the components for lead time (t0 + t)
```
RubenImhoff, line 1478:
```suggestion
            # Determine the skill of the components for lead time (t0 + t)
```

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2858 @@
...
-                # 8.1.2 Calculate the initial skill of the (NWP) model forecasts at t=0
-                rho_nwp_models = _compute_initial_nwp_skill(
-                    precip_cascade,
-                    precip_models_cascade_temp,
-                    domain_mask,
-                    issuetime,
-                    outdir_path_skill,
-                    clim_kwargs,
+    def __determine_skill_for_next_timestep(self, t, j, worker_state):
+        # 8.1.2 Determine the skill of the nwp components for lead time (t0 + t)
```
RubenImhoff, line 1490:
```suggestion
        # Determine the skill of the nwp components for lead time (t0 + t)
```

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2858 @@
...
-            if t == 0:
-                # 8.1.2 Calculate the initial skill of the (NWP) model forecasts at t=0
-                rho_nwp_models = _compute_initial_nwp_skill(
-                    precip_cascade,
-                    precip_models_cascade_temp,
-                    domain_mask,
-                    issuetime,
-                    outdir_path_skill,
-                    clim_kwargs,
+    def __determine_skill_for_next_timestep(self, t, j, worker_state):
```
RubenImhoff: Would it make more sense to let this function be part of, or be used in,  `__determine_skill_for_current_timestep`?

RubenImhoff: I see now that this function is used individually in the parallel workers. If we leave it as is, maybe good to change the function name to something indicating that we are dealing with the model skill here (and not the extrapolation skill).

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2858 @@
...
-                # 8.1.3 Determine the skill of the components for lead time (t0 + t)
-                # First for the extrapolation component. Only calculate it when t > 0.
-                (
-                    rho_extr,
-                    rho_extr_prev,
-                ) = blending.skill_scores.lt_dependent_cor_extrapolation(
-                    PHI=PHI, correlations=rho_extr, correlations_prev=rho_extr_prev
-                )
+    def __determine_weights_per_component(self, worker_state):
+        # 8.2 Determine the weights per component
```
RubenImhoff, line 1524:
```suggestion
        """
        Determine the weights per component. Weights following the bps method are calculated. 
        These are needed for the velocity weights prior to the advection step. If weights method 
        spn is selected, weights will be overwritten with those weights prior to blending step.
        """
```

The remainder of the docstring below can then be removed.

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2858 @@
...
-                        else:
-                            R_f_ip = [
-                                forecast_prev[j][i][-1, :]
-                                for i in range(n_cascade_levels)
-                            ]
-                            Yn_ip = [
-                                noise_prev[j][i][-1, :] for i in range(n_cascade_levels)
-                            ]
+    def __regress_extrapolation_and_noise_cascades(self, j, worker_state):
+        # 8.3 Determine the noise cascade and regress this to the subsequent
```
RubenImhoff, line 1590:
```suggestion
        # Determine the noise cascade and regress this to the subsequent
```
Instead of the numbering here, maybe good to make this order clear in the docstrings of the main function(s).

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2858 @@
...
+        # time step + regress the extrapolation component to the subsequent
+        # time step
+
+        # 8.3.1 Determine the epsilon, a cascade of temporally independent
```
RubenImhoff, line 1594:
```suggestion
        # Determine the epsilon, a cascade of temporally independent
```

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2858 @@
...
+                input_domain=self.__config.domain,
+                output_domain=self.__config.domain,
+                compute_stats=True,
+                normalize=True,
+                compact_output=True,
+            )
+        else:
+            epsilon_decomposed = None
+
+        # 8.3.2 regress the extrapolation component to the subsequent time
```
RubenImhoff, line 1619:
```suggestion
        # Regress the extrapolation component to the subsequent time
```

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2858 @@
...
+            else:
+                # use the deterministic AR(p) model computed above if
+                # perturbations are disabled
+                worker_state.precip_cascades[j][i] = (
+                    worker_state.final_blended_forecast_non_perturbed[i]
+                )
 
-                        t_diff_prev = t_sub - t_prev[j]
-                        t_total[j] += t_diff_prev
+        # 8.3.3 regress the noise component to the subsequent time step
```
RubenImhoff, line 1642:
```suggestion
        # Regress the noise component to the subsequent time step
```

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2858 @@
...
-                        velocity_blended = blending.utils.blend_optical_flows(
-                            flows=V_stack,
-                            weights=weights[
-                                :-1, 1
-                            ],  # [(extr_field, n_model_fields), cascade_level=2]
-                        )
+    def __perturb_blend_and_advect_extrapolation_and_noise_to_current_timestep(
+        self, t, j, worker_state
+    ):
+        # 8.4 Perturb and blend the advection fields + advect the
```
RubenImhoff, line 1665:
```suggestion
        """
        Perturb and blend the advection fields + advect the
        extrapolation and noise cascade to the current time step
        (or subtimesteps if non-integer time steps are given)
```

And you can remove the subsequent docstrings now that they are in the function docstring.

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2858 @@
...
+        # extrapolation and noise cascade to the current time step
+        # (or subtimesteps if non-integer time steps are given)
+
+        # Settings and initialize the output
+        extrap_kwargs_ = self.__state.extrapolation_kwargs.copy()
+        extrap_kwargs_noise = self.__state.extrapolation_kwargs.copy()
+        extrap_kwargs_pb = self.__state.extrapolation_kwargs.copy()
+        velocity_perturbations_extrapolation = self.__velocity
+        # The following should be accesseble after this function
```
RubenImhoff:
```suggestion
        # The following should be accessible after this function
```

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2858 @@
...
+                self.__determine_skill_for_next_timestep(t, j, worker_state)
+                self.__determine_weights_per_component(worker_state)
+                self.__regress_extrapolation_and_noise_cascades(j, worker_state)
+                self.__perturb_blend_and_advect_extrapolation_and_noise_to_current_timestep(
+                    t, j, worker_state
+                )
+                # 8.5 Blend the cascades
+                final_blended_forecast_single_member = []
+                for t_sub in self.__state.subtimesteps:
+                    # TODO: does it make sense to use sub time steps - check if it works?
```
mats-knmi, line 355: Regarding the sub timesteps: I would love it if we could make it so that pysteps can just handle a variable timestep.

When you would go for example from a 5 to 15 minute timestep you could aggregate the previous 3 precipitation fields and multiply the last motion field by 3 and you would then probably just be able to continue in a 15 minute timestep from there right?

This would be much preferable to the sub timestep, which still has to do all the computations at the smaller timesteps. Also it should make the code easier to follow maybe

mats-knmi: I don't think this is part of the refactoring, but maybe something for after

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2858 @@
...
+        if self.__config.noise_method is not None:
+            self.__state.randgen_precip = []
+            self.__state.randgen_motion = []
+            for j in range(self.__config.n_ens_members):
+                rs = np.random.RandomState(self.__config.seed)
+                self.__state.randgen_precip.append(rs)
+                seed = rs.randint(0, high=1e9)
+                rs = np.random.RandomState(seed)
+                self.__state.randgen_motion.append(rs)
+                seed = rs.randint(0, high=1e9)
```
mats-knmi: I don't think this works `self.__config.seed` does not get updated, so every member will have the same seed, this would mean every member is the same right?

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2858 @@
...
+
+        if self.__config.velocity_perturbation_method is not None:
+            (
+                init_velocity_noise,
+                self.__params.generate_velocity_noise,
+            ) = noise.get_method(self.__config.velocity_perturbation_method)
+
+            # initialize the perturbation generators for the motion field
+            self.__state.velocity_perturbations = []
```
mats-knmi: According to my linter `velocity_perturbations` is not present in the `StepsBlendingState` definition above.

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2858 @@
...
+        the inputs like precip, velocity, time_steps, or config.
+        """
+        # Re-initialize the state and parameters
+        self.__state = StepsBlendingState()
+        self.__params = StepsBlendingParams()
+
+        # Reset time measurement variables
+        self.__start_time_init = None
+        self.__init_time = None
+        self.__mainloop_time = None
```
mats-knmi: Does this not only make sense if you specify `precip` and `velocity` (and `precip_models` and `velocity_models`) when you call `compute_forecast`, because otherwise you would only be able to rerun the exact same forecast with the same inputs again?

mats-knmi:
I just scrolled though the entire code. I did not read everything, but I mostly tried to follow along. I have added some additonal comments to the things I found, other than that it is good to go from my part (after @RubenImhoff also has had a look).

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2858 @@
...
+                    velocity_blended,
+                    [t_diff_prev_subtimestep],
+                    allow_nonfinite_values=True,
+                    **self.__state.extrapolation_kwargs,
+                )
+                precip_extrapolated_recomp_subtimestep = (
+                    precip_forecast_extrapolated_recomp_subtimestep_temp[0].copy()
+                )
+                temp_mask = ~np.isfinite(precip_extrapolated_recomp_subtimestep)
+                # TODO: WHERE DO CAN I FIND THIS -15.0
```
RubenImhoff, line 1798: I'm not sure if any decision have been made regarding the `zero_value`, as the same question is present in the `nowcasts` modules (see also issue #438).

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2858 @@
...
+                precip_forecast_extrapolated_recomp_subtimestep_temp = None
+                precip_extrapolated_recomp_subtimestep = None
+                precip_extrapolated_decomp = None
+                noise_cascade_subtimestep = None
+                noise_cascade_subtimestep_recomp = None
+                noise_extrapolated_recomp_temp = None
+                noise_extrapolated_recomp = None
+                noise_extrapolated_decomp = None
+
+                # Finally, also extrapolate the initial radar rainfall
```
RubenImhoff, line 1875:
```suggestion
                # C. Finally, also extrapolate the initial radar rainfall
```
For consistency with the docstrings above.

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2858 @@
...
-                        elif probmatching_method == "mean":
-                            # Use R_pm_blended as benchmark field and
-                            mu_0 = np.mean(R_pm_resampled[R_pm_resampled >= precip_thr])
-                            MASK = R_f_new >= precip_thr
-                            mu_fct = np.mean(R_f_new[MASK])
-                            R_f_new[MASK] = R_f_new[MASK] - mu_fct + mu_0
-                            R_pm_resampled = None
+        # First determine the blending weights if method is spn. The
+        # weights for method bps have already been determined.
+        # TODO: no other weight method is possible, should we not al least give a user warning if a different weight
```
RubenImhoff: Good point, maybe add a check that either "BPS" or "SPN" is provided? (or nothing, as that defaults to BPS)

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2858 @@
...
+            worker_state.final_blended_forecast_stds_mod_only,
+        ) = blend_means_sigmas(
+            means=means_stacked[1:, :],
+            sigmas=sigmas_stacked[1:, :],
+            weights=worker_state.weights_model_only,
+        )
 
-            res = None
+    def __recompose_cascade_to_rainfall_field(self, j, worker_state):
+        # 8.6 Recompose the cascade to a precipitation field
```
RubenImhoff, line 2140:
```suggestion
        """ 
        Recompose the cascade to a precipitation field
        """
```
The docstrings below can be removed.

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2858 @@
...
 
-            if is_nowcast_time_step:
-                if measure_time:
-                    print(f"{time.time() - starttime:.2f} seconds.")
-                else:
-                    print("done.")
+    def __post_process_output(
+        self, j, final_blended_forecast_single_member, worker_state
+    ):
+        # 8.7 Post-processing steps - use the mask and fill no data with
```
RubenImhoff, line 2173:
```suggestion
        """ 
        Post process output - use the mask and fill no data with
        the blended NWP forecast. Probability matching follows
        Lagrangian blended probability matching which uses the
        latest extrapolated radar rainfall field blended with the
        nwp model(s) rainfall forecast fields as 'benchmark'.
        """
```

The subsequent docstrings can be removed.

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2858 @@
...
+        # the blended NWP forecast. Probability matching following
+        # Lagrangian blended probability matching which uses the
+        # latest extrapolated radar rainfall field blended with the
+        # nwp model(s) rainfall forecast fields as 'benchmark'.
+
+        # 8.7.1 first blend the extrapolated rainfall field (the field
```
RubenImhoff, line 2179:
```suggestion
        # First blend the extrapolated rainfall field (the field
```
Perhaps give this the variable name, instead of extrapolated rainfall field, as it occurs in the class.

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2858 @@
...
+        nan_indices = np.isnan(worker_state.final_blended_forecast_recomposed)
+        worker_state.final_blended_forecast_recomposed[nan_indices] = np.nanmin(
+            worker_state.final_blended_forecast_recomposed
+        )
+        nan_indices = np.isnan(precip_forecast_probability_matching_blended)
+        precip_forecast_probability_matching_blended[nan_indices] = np.nanmin(
+            precip_forecast_probability_matching_blended
+        )
+
+        # 8.7.2. Apply the masking and prob. matching
```
RubenImhoff, line 2303:
```suggestion
        # Apply the masking and probability matching
```

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2858 @@
...
+    # correlations: [component, scale, ...]
+    square_corrs = np.square(correlations)
+    # Calculate the ratio of the explained variance to the unexplained
+    # variance of the nowcast and NWP model components
+    out = square_corrs / (1 - square_corrs)
+    # out: [component, scale, ...]
+    return out
+
+
+def calculate_weights_bps(correlations):
```
RubenImhoff: Now that we are cleaning up the code, would it make sense to move `calculate_weights_bps` and `calculate_weights_spn` to a separate model, called `weights.py` or so? However, at the same time, these weights are quite specific to the STEPS approach, so I'm not sure what is best. It would make the code here a little shorter and cleaner at least.

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2858 @@
...
-    return out
+      vet/fmi+mch:
+      p_par  = [0.29495222, 0.62429207, 8.6804131 ]
+      p_perp = [0.23127377, 0.59010281, 5.98180004]
 
+      fmi=Finland, mch=Switzerland, fmi+mch=both pooled into the same data set
 
-def calculate_weights_bps(correlations):
-    """Calculate BPS blending weights for STEPS blending from correlation.
+      The above parameters have been fitten by using run_vel_pert_analysis.py
```
RubenImhoff:
```suggestion
      The above parameters have been fitted by using run_vel_pert_analysis.py
```

pysteps/blending/steps.py
```diff
@@ -1918,577 +3053,3 @@ def blend_means_sigmas(means, sigmas, weights):
     # TODO: substract covariances to weigthed sigmas - still necessary?
```
RubenImhoff: I think we can remove this TO DO, as we already have this one open since the original implementation. I think we can say that the answer is no.

pysteps/blending/steps.py
sidekock: Indeed, but I think the refactoring makes the process a lot easier to to this.

pysteps/blending/steps.py
sidekock: I guess, i'll let @RubenImhoff  answer this one to be sure. Its an easy fix by itself. Am I correct in assuming the fix would be the following: 
- make seed a state variable (to adhere to the rule of config variables not being changed)
- assign self.__state.seed = rs.randint(0, high=1e9) on line 1085

pysteps/blending/steps.py
sidekock: Indeed this should be __params instead of state. Will be fixed in the next commit

pysteps/blending/steps.py
sidekock: I followed the same logic here as for the nowcaster but maybe that is not optimal for the nowcaster either. The idea is if people want to generate a bunch of forecasts, they can make one config and then loop over all forecast they want to do, each forecast makes a new blendednowcaster which then computed the forecasts. But indeed, in that scenario, we don't need the reset_states... What do you think would be the best way forward:
- Remove this reset and keep the rest as is
- Let people create a forecaster once and add the precip and velocity fields as parameters to the compute_forecast()
- Something else

pysteps/blending/steps.py
mats-knmi: I think you could also just do `seed = self.__config.seed` before the loop and then replace line 1080 with `rs = np.random.RandomState(seed)`

pysteps/blending/steps.py
```diff
@@ -65,1723 +64,2857 @@
...
-      All values are required to be finite.
 
-      To reduce memory usage, this array
-      can be given as float32. They will then be converted to float64 before computations
-      to minimize loss in precision.
-    timesteps: int or list of floats
-      Number of time steps to forecast or a list of time steps for which the
-      forecasts are computed (relative to the input time step). The elements of
-      the list are required to be in ascending order.
+@dataclass
```
mats-knmi: I just realized, based on your comment that the config variables should be immutable, that this should be a frozen dataclass to make that explicit. This will simply disallow you from mutating the config object after creation.
```suggestion
@dataclass(frozen=True)
```

pysteps/blending/steps.py
mats-knmi: Stylistically I like the idea of a reusable nowcast instance that you initialize once with a config and then can call multiple times with different precip and velocity fields (so option 2).

Practically though option 1 might be more robust. If you create a new nowcast instance every time that means that the old one (and all data in it) can be cleaned up by the garbage collector. If you keep the nowcast instance around for an extended time (in a production setting for example) you rely on the reset state method to clear everything, so that needs to be called every time after a nowcast has completed and the reset state method **has** to reset the state properly every time. So data running over from one run into the other or memory leaks happening are more likely with option 2.

Also the way that it is done now, where everything (including the inputs) are just stored in `self` and nothing is passed between methods using arguments fits better with option 1 I think.

RubenImhoff (commented):
@sidekock, super nice work! As @mats-knmi already mentioned, it is a lot of code, so I'm sure I've missed things. My main recommendation would be do to some comparison tests to ensure all functionality still operates the same. 

In addition, some overall points I had:

- Docstrings: already mentioned, but good to make this as clear as possible to in the end have a blending module that is more understandable than the old version was. :)
- Also, don't forget to sometimes add some short docstrings line-by-line in the individual functions. 
- Question: What internal functions in this piece of code do have overlap (or more or less overlap) with `nowcasts/steps.py`. Meaning, could we put some functions separate in a utils script or so?

pysteps/blending/steps.py
RubenImhoff: I agree, I think that approach would work, @mats-knmi :)

pysteps/blending/steps.py
sidekock: The problem with that fix is that during __check_inputs some variables are set to be a dict or something else if nothing is provided by the user. There are two suggestions:
- Do not used the frozen label
- Define the standard of the variables to be the assigned values in __check_inputs. This might not work for all scenarios as the assignments might be conditional.

pysteps/blending/steps.py
sidekock: Indeed that is the goal. Based on the feedback on the code I would first change the code and the, when it is in a final state. I will first compare master to this branch and then focus on documentation

pysteps/blending/steps.py
sidekock: I think the current solution is thread-safe, but because we now save everything in dataclasses to avoid the 20 return values of each function call, we have the downside of needing specific worker_states not to overwrite general information for info about a specific worker. Hope I made this clear :)

pysteps/blending/steps.py
sidekock: Agreed, maybe best to add this as a new issue to be fixed after this pull request is complete

pysteps/blending/steps.py
sidekock: I am not sure where to move it best. Two other functions are used a few times as well and I don't know where to move them to... Some, I would say, could maybe be added to utils

pysteps/blending/steps.py
sidekock: I made this change manually for the next commit

pysteps/blending/steps.py
mats-knmi: I think option 2 should work, I see that there is one case where the config default value is conditional. I think all the other cases could just be default values. For this one case I also think that we should be able to work around this pretty easilly, but this does raise the question to me if this derived value should be a part of the config class?
Is `n_models` not always equal to `self.__precip_models.shape[0]` and thus it should not be part of a config that you can also provide yourself?
We could for example move this value to the state in stead of the config?

```
            self.__config.climatology_kwargs = dict(
                {"n_models": self.__precip_models.shape[0]}
            )
```

mats-knmi: There is actually one more place where the config changes during the nowcasting with is this line. I also think we should be able to do without this line right? Because `xy_coordinates` is already stored in the params so it does not need to be copied to the config.
```
self.__config.extrapolation_kwargs["xy_coords"] = self.__params.xy_coordinates
```

mats-knmi: One more thing I would like to add is that while `frozen=True` makes the config object immutable any dict that is in it is still mutable, so the line from my previous comment would still be allowed. Ideally also that line would be disallowed. Apparently you can use `MappingProxyType` to make the dictionaries immutable as well: https://adamj.eu/tech/2022/01/05/how-to-make-immutable-dict-in-python/

mats-knmi: Ok well immutable dicts opens its own can of worms so maybe that is too much for now, but `frozen=True` is definitely attainable. I tried this out myself: https://github.com/pySTEPS/pysteps/pull/447/files and the tests still seem to succeed. I did remove the `{"n_models": self.__precip_models.shape[0]}` part of the code, but as far as I can tell it is used nowhere in this file, do you know if this is used anywhere?

pysteps/blending/steps.py
sidekock: I guess that would mean, no rain at all in the NWP and then we would be in the 
"if self.__params.zero_precip_radar and self.__params.zero_precip_model_fields" case so then we this piece of code is not used. @mats-knmi wrote this so he can explain this better :)

pysteps/blending/steps.py
sidekock: I think that might be a nice addition... This could be another issue we can open.

pysteps/blending/steps.py
sidekock: The problem is that this would mean we have to rewrite the decompositor importer code, and then we have breaking changes, so that is why I kept the left-hand side of these function arguments with the old names...

pysteps/blending/steps.py
sidekock: We could do that. Should I add it to this pr?

pysteps/blending/steps.py
sidekock: Indeed, this is done now

pysteps/blending/steps.py
sidekock: I ran both and no issues in either one, so I removed it

pysteps/blending/steps.py
sidekock: Changed the name to __determine_NWP_skill_for_next_timestep

pysteps/blending/steps.py
sidekock: I agree but two questions with this TODO:
- Mabye we can give it a better description?
- Is it something inherent to the nowcasting/blending or something pysteps specific?

pysteps/blending/steps.py
sidekock: But the BPS does nothing special then? Just to check how I would implement this in detail. Would it be something like:
if self.__config.weights_method == "spn" or self.__config.weights_method == "bps":
     if self.__config.weights_method == "spn":
            previous code
      elif self.__config.weights_method == "bps":
          pass # because nothing special happens for this one
I just added the first test since you asked for it but this is already checked in the __determine_weights_per_component() method.

pysteps/blending/steps.py
RubenImhoff: Okay, that should be good!

pysteps/blending/steps.py
RubenImhoff: Good idea!

pysteps/blending/steps.py
RubenImhoff: Fair point, let's leave it like this then for now. This could be a TO DO for version 2, as breaking changes are okay then.

pysteps/blending/steps.py
RubenImhoff: Yes, thanks!

pysteps/blending/steps.py
RubenImhoff: Great, thanks! :)

pysteps/blending/steps.py
RubenImhoff: I think it is (/can be) used in a many of the transformers and converters. But I'm not entirely sure.

pysteps/blending/steps.py
RubenImhoff: Ah sorry, I mean for the warning as mentioned in your TODO. I think good to check somewhere in the code that the right method is provided (or is that already done?) and that a warning or error is returned if another method is provided. If that is already there, you can remove this TODO.

pysteps/blending/steps.py
RubenImhoff: I'll think about it. Maybe I'll put them in a separate `weights.py` module, but otherwise they're good where they are now.

pysteps/blending/steps.py
sidekock: the "n_models" is used but in one of the helper functions of the sklill_scores: clim_regr_values(). So just removing it is not an option. However, if I make the standard a dict and a dict is still mutable, I don't think we have an issue, right?

sidekock: I just checked this and my tests seem to fail: The reason is, the original forecast function defines a lot of these variables that are assigned to be a dict to be None. Thus, my "field(default_factory=dict)" assigmnet as the default value does not apply as something is provided by the user. I could however change all the mutated variables over to the params class and then we would get rid of that issue... 
What do you think?

pysteps/blending/steps.py
sidekock: Okay, I will remove the reset funtion and add some comments as to how the user is expected to deal with the class.

pysteps/blending/steps.py
sidekock: I think the movement to the __params is not a bad idea with running multiple forecasts in mind. Lets say the __config would be mutable and the first forecast ran has x members while a later forecast has y members. That would currently not be dealt with correctly by the code. Is we move the dict assignments to the __params, that issue would go away.

sidekock:
> @sidekock, super nice work! As @mats-knmi already mentioned, it is a lot of code, so I'm sure I've missed things. My main recommendation would be do to some comparison tests to ensure all functionality still operates the same.
> 
> In addition, some overall points I had:
> 
>     * Docstrings: already mentioned, but good to make this as clear as possible to in the end have a blending module that is more understandable than the old version was. :)
> 
>     * Also, don't forget to sometimes add some short docstrings line-by-line in the individual functions.
> 
>     * Question: What internal functions in this piece of code do have overlap (or more or less overlap) with `nowcasts/steps.py`. Meaning, could we put some functions separate in a utils script or so?

Hi ruben, 
Docstrings are the thing to look at if the code has a finalized form but that does not seem to far away now :). Regarding your last question: conceptually there is quite a bit of similarity between nowcasting and blending steps. However there are very few pieces of code that are duplicate, this is because we have additional NWP fields to take into account. Stylisticly, I agree that we should keep the duplicated code to a minimum but readability wise and for understanding, I also see the benefit of keeping all the steps of the workflow in one file. If this range would be really needed, we could also make a new issue out of this, because i think it is probably something we do best on its own.

pysteps/blending/steps.py
mats-knmi: I just saw your commit where you made the config object immutable. Looks good!

pysteps/blending/steps.py
sidekock: Should I mark this as an issue or write a TODO line. It seems a bit to big of a tasks to put this as a one line TODO in one python file

pysteps/blending/steps.py
sidekock: Should these negative numbers be set to NaN or to 0?

pysteps/blending/steps.py
sidekock: So it is a more general issue? Then maybe this TODO is not the best place to add this problem. Maybe a new issue?

mats-knmi (approved):

mats-knmi:
@sidekock, since I had this regression test set up for our code I thought it would be useful to give your refactored code a go to see if it produced the same output as the original code.

I found that there was one issue with the refactored code and that is the statement below. This statement makes it so that any changes made to (for example) `precip_cascades` and `precip_noise_cascades`, do not carry over into the next iteration. In your comment you note that this was done because of thread safety, but I believe that the current pysteps implementation is not inherently thread safe either right? Correct me if I am wrong, but I believe that numpy arrays are also not thread safe, but the pysteps code is just written in such a way that each thread is only mutating the part behind the `j`th index.

```
worker_state = deepcopy(self.__state)
```

If you remove this part, then your refactored code passes my regression test. This means that for the implementation at KNMI the refactored code works identical to the old code. I have only tested our implementation, so it is not a complete test, but at least a first good result.

sidekock:
> @sidekock, since I had this regression test set up for our code I thought it would be useful to give your refactored code a go to see if it produced the same output as the original code.
> 
> I found that there was one issue with the refactored code and that is the statement below. This statement makes it so that any changes made to (for example) `precip_cascades` and `precip_noise_cascades`, do not carry over into the next iteration. In your comment you note that this was done because of thread safety, but I believe that the current pysteps implementation is not inherently thread safe either right? Correct me if I am wrong, but I believe that numpy arrays are also not thread safe, but the pysteps code is just written in such a way that each thread is only mutating the part behind the `j`th index.
> 
> ```
> worker_state = deepcopy(self.__state)
> ```
> 
> If you remove this part, then your refactored code passes my regression test. This means that for the implementation at KNMI the refactored code works identical to the old code. I have only tested our implementation, so it is not a complete test, but at least a first good result.

Hi Mats,
Thank you for already running some real-life tests on the code! I indeed got a similar problem when refactoring the steps nowcasting now that you mention it. So I just remove the whole line? In that case, I am wondering how the code would work since the worker state has not been initialised. Or do I just assign the worker_state to be self.__state?
I think for now, it is already good to have one type of test passing. I will try to run some RMI setup tests next week or the week after once the exams are past me.

mats-knmi:
> > @sidekock, since I had this regression test set up for our code I thought it would be useful to give your refactored code a go to see if it produced the same output as the original code.
> > I found that there was one issue with the refactored code and that is the statement below. This statement makes it so that any changes made to (for example) `precip_cascades` and `precip_noise_cascades`, do not carry over into the next iteration. In your comment you note that this was done because of thread safety, but I believe that the current pysteps implementation is not inherently thread safe either right? Correct me if I am wrong, but I believe that numpy arrays are also not thread safe, but the pysteps code is just written in such a way that each thread is only mutating the part behind the `j`th index.
> > ```
> > worker_state = deepcopy(self.__state)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If you remove this part, then your refactored code passes my regression test. This means that for the implementation at KNMI the refactored code works identical to the old code. I have only tested our implementation, so it is not a complete test, but at least a first good result.
> 
> Hi Mats, Thank you for already running some real-life tests on the code! I indeed got a similar problem when refactoring the steps nowcasting now that you mention it. So I just remove the whole line? In that case, I am wondering how the code would work since the worker state has not been initialised. Or do I just assign the worker_state to be self.__state? I think for now, it is already good to have one type of test passing. I will try to run some RMI setup tests next week or the week after once the exams are past me.

I indeed just replaced that line with `worker_state = self.__state`, but if you do that you could also just remove the `worker_state` all together and just reference `self.__state` directly as that would result in exactly the same behaviour.

pysteps/postprocessing/probmatching.py
```diff
@@ -274,7 +274,7 @@ def _get_error(scale):
     return shift, scale, R.reshape(shape)
 
 
-def resample_distributions(first_array, second_array, probability_first_array):
+def resample_distributions(first_array, second_array, probability_first_array, randgen):
```
RubenImhoff: Good to also add in the docstrings below what randgen does.

RubenImhoff (approved):

pysteps/postprocessing/probmatching.py
mats-knmi: Yes I updated this in the PR to master, maybe you can take over those changes here as well @sidekock https://github.com/pySTEPS/pysteps/pull/450/commits/c4ae29c432de6746063e249e4e19ce081388fec6

mats-knmi:
Now that this refactoring is mostly done, I am planning to start work on the xarray implementation again sometime next week I will merge these changes into that branch and start work on the blending code, because I expect no major changes to happen anymore here right @sidekock?

sidekock:
> Now that this refactoring is mostly done, I am planning to start work on the xarray implementation again sometime next week I will merge these changes into that branch and start work on the blending code, because I expect no major changes to happen anymore here right @sidekock?

I just need some last tests and some documentation needs to be moved. I can do this int the middle of next week as soon as possible. So if the tests are working, there should be no more changes, with your first tests, I think it is fine to already start adapting this to xarray.

pysteps/postprocessing/probmatching.py
sidekock: Done now

sidekock:
@RubenImhoff I just ran some test on Belgian data and I seem to be having a problem. I though you might be able to point me in the correct direction as you probably know the code the best:
image
The top row of the image is the output from the master, the bottom row from the refactored version (the 5 plots are 5 members). There are some differences outside the radar mask. My question is, where in the code should I be looking for this? I this in probablility_matching or somewhere else? 
@mats-knmi you said you already ran some test with both the old and new config, did you see a difference between the two? Did the code run correctly? I had to make a few changes to the code to make it run (there was an issue with some parts being initialised as np.array and others where normal arrays.

RubenImhoff:
Hi @sidekock, Hmm, looks like you almost entirely filter out the NWP? Could you check if that data pipeline goes well? If all goes well on that side, I expect that it may have to do with the probability matching (but not sure yet).

sidekock:
> Hi @sidekock, Hmm, looks like you almost entirely filter out the NWP? Could you check if that data pipeline goes well? If all goes well on that side, I expect that it may have to do with the probability matching (but not sure yet).

So there seem to be multiple problems. I will list them below but I do not seem to find a solution:
1) Indeed NWP does not seem to be properly blended
2) All the members of the new branch seem te be the same or of 2 batches (not 5 invdivudal members)
3) Currently, sometimes the code crashes for some weird reason, I will try to explain this below as best I can but I don't fully understand it:
Sometime there is a crash because the of the following line: worker_state.precip_extrapolated_probability_matching.append(
                    precip_forecast_extrapolated_probability_matching_temp[0]
                )
The reason this happens in the trace seems to be because numpy arrays do not support append (which is true, however as far as I can see in the code, this worker_state.precip_extrapolated_probability_matching is always a list until the function just before it, where apparently for some reason it changes? I checked this by printing the type because I debug mode is hard when working with multiple workers... I now added some np.array to list conversions just before these appends and that does seem to work. However, now the code is crashing in a different place, when stacking the code (in the else loop):             precip_forecast_probability_matching_final = np.concatenate(
                (
                    worker_state.precip_extrapolated_probability_matching[
                        None, worker_state.subtimestep_index
                    ],
                    worker_state.precip_models_timestep[None, j],
                ),
                axis=0,
            )
I get the following error: TypeError: list indices must be integers or slices, not tuple. I cant seem to solve this problem in any way. Could anyone else try and take a look? I am kind of breaking my head over this one. 

The best way forward appears to be to solve problem 1 by treating the problem with only 1 member and thus one thread, that way we can already evaluate the single thread output. Once that is done I think we should solve problem 3 because only after we can compare the performance of a multi member nowcast. 

Sorry, this project is taking slightly longer as expected :(

sidekock:
@RubenImhoff @mats-knmi Here is a minor update on the problem. The code runs well on a single core (no crashes means 'well). I still need to look at the outputted data to compare the master versus the new branch. If these do not match up, I will try to fix that issue first. Afterwards, I will look at the multithreaded problem, but I might need some help there as I do not see a solution or the cause of the crash here.
@dnerini Is it possible to write tests to multithread the code? I think so far, this issue has avoided all the tests that are present, and I am afraid that if someone happens in the future, we will have to revert the master for code that was not properly tested.

sidekock:
@RubenImhoff I don't know how much the NWP needs to take over the extrapolation at around 2 hours but I clearly see NWP blending in outside the radar domain. If this assumption is right, I think the following plots show that the issue is with running the code with multiple working threads at the same time:
Plot info: Top is master, bottom is this branch. Left to right are the 5 members for in this case timestep 24. The member togle is to switch between the two NWPs we have at RMI.
image

sidekock:
> @sidekock, since I had this regression test set up for our code I thought it would be useful to give your refactored code a go to see if it produced the same output as the original code.
> 
> I found that there was one issue with the refactored code and that is the statement below. This statement makes it so that any changes made to (for example) `precip_cascades` and `precip_noise_cascades`, do not carry over into the next iteration. In your comment you note that this was done because of thread safety, but I believe that the current pysteps implementation is not inherently thread safe either right? Correct me if I am wrong, but I believe that numpy arrays are also not thread safe, but the pysteps code is just written in such a way that each thread is only mutating the part behind the `j`th index.
> 
> ```
> worker_state = deepcopy(self.__state)
> ```
> 
> If you remove this part, then your refactored code passes my regression test. This means that for the implementation at KNMI the refactored code works identical to the old code. I have only tested our implementation, so it is not a complete test, but at least a first good result.

Hi Mats,
With all the issues I faced before, I thought that it might be a good idea to switch back to before this "fix". Now the code seems to both compile and produce different members (see picture). I do agree however that in the final version, we need to go back to a global state but this shows me that, the current implementation after this change is not theadsafe and that certain threads are accessing and modifying data that should not be toughed by them:
image
I did make one change compared to the previous file, the rang_gen is still done from the state as else I guess I will have 5 members which look the same (still need to check if this assumption holds.
EDIT: this assumption was false but I do see why, most of the random parts of the code are already generated by this point. I see a few possibles way forward which would solve your problem and mine:
1) We try to find the place in the code where there is some non-thread-safe things going on
2) We create a worker_state for each possible worker which would be stored throughout the whole computation, I think that would solve your error related to the precip_cascades and precip_noise_cascades. The only thing we need to make sure then is that each of these workers is not generating the same random numbers. For this we could go 2 ways: 
a) we look how many random numbers are calculated per timestep (not clean and prone to errors in possible extensions)
b) we just provide new seeds to each of the members. This will result in non-compairable matches between the master and this version but would be a lot quicker to implement as option 1 :p

mats-knmi:
Hmm OK I think I understand what is going on...
First of all, yes I did run this code without any issues on a full test run and the output was exactly the same as for the master branch. **But** in the KNMI implementation we do not make use of pysteps multithreading, since we just call pysteps 20 times in 20 different processes each with a different seed. This multiprocessing works better for our implementation and doesn't have these annoying thread unsafety issues.

The issues that you are having almost certainly have to do with the thread safety of the worker state indeed, earlier I missed the extend that the worker state has and therefore missed this issue. The problem is that currently the worker state contains variables that are supposed to be accessed by all threads and variables that are not supposed to be accessed this way. Basically the variables that used to be defined outside the worker method are required to be accessed by all workers at once (because that is how it used to be). However all variables that used to be defined inside the worker method are not supposed to be accessed by all variables at once, so these variables should **not** go into the shared state.

Basically I believe that your initial idea of defining a `worker_state` is the way to go, but it should **not** contain the variables from `self.__state`, instead it should contain the variables that are defined inside the worker and have to be passed around the many sub methods in there. Conversely the `self.__state` should **not** contain the variables that do go into the `worker_state`.

Does this make sense?

mats-knmi:
<del>I believe that this worker state should only contain all those variables that are directly assigned at any point in the worker. By doing a regex search I find these variables that should not be in the `self.__state`, but should go into a separate `worker_state`:</del>
<del>- rho_final_blended_forecast</del>
<del>- precip_extrapolated_decomp</del>
<del>- noise_extrapolated_decomp</del>
<del>- precip_extrapolated_probability_matching</del>
<del>- subtimestep_index</del>
<del>- weights</del>
<del>- final_blended_forecast_cascades</del>
<del>- final_blended_forecast_cascades_mod_only</del>
<del>- final_blended_forecast_recomposed</del>
<del>- final_blended_forecast_recomposed_mod_only</del>
EDIT: I think the proposed solution below is more promissing

sidekock:
> I believe that this worker state should only contain all those variables that are directly assigned at any point in the worker. By doing a regex search I find these variables that should not be in the `self.__state`, but should go into a separate `worker_state`:
> 
>     * rho_final_blended_forecast
> 
>     * precip_extrapolated_decomp
> 
>     * noise_extrapolated_decomp
> 
>     * precip_extrapolated_probability_matching
> 
>     * subtimestep_index
> 
>     * weights
> 
>     * final_blended_forecast_cascades
> 
>     * final_blended_forecast_cascades_mod_only
> 
>     * final_blended_forecast_recomposed
> 
>     * final_blended_forecast_recomposed_mod_only

Thanks, this saves me a lot of work. I should be able to finish this issue by tonight, then we hopefully can merge

mats-knmi:
I think I missed some variables when doing the regex search as some assignments go over multiple lines. But I had another idea that might be even better and easier. In stead of making a deep copy of the worker state, you could make a shallow copy (by using `copy.copy` in stead of `copy.deepcopy`). I am 99% sure that this should be exactly what we need and solve all problems at once.

The reason this works is that any variable that gets reassigned completely will not overwrite the existing variable in the `self.__state`, but things like `worker_state.precip_cascades[j][i] = ...` will still just write to the global state, which is exactly what we want.

Basically this should revert the behavior to exactly how it was before the refactoring, because in the old code a call like `precip_extrapolated_decomp = ...` would create a new variable in the worker that shadows the existing variable from the outer scope, but a call like `precip_cascades[j][i] = ...` would overwrite data in the existing variable from the outer scope. A shallow copy of the `self.__state` would behave the exact same way.

sidekock:
Thanks to @mats-knmi it seems we have a working version that reproduces the master exactly:
image
I will do some additional checks to make sure this works in all scenarios (single worker vs multiple workers) and for longer lead times with more members to make absolutely sure but I am hopeful :)

mats-knmi:
@sidekock Did you only do this validation visually or also on dataset level? Because in my tests I did the following and that succeeded also.
```
import xarray as xr
assert xr.load_dataset("old_output.nc").identical(xr.load_dataset("new_output.nc"))
```

sidekock:
> @sidekock Did you only do this validation visually or also on dataset level? Because in my tests I did the following and that succeeded also.
> 
> ```
> import xarray as xr
> assert xr.load_dataset("old_output.nc").identical(xr.load_dataset("new_output.nc"))
> ```

This was indeed the last check I wanted to do :) The visual comparison is just a bit easier while debugging because most of the times you can see what goes wrong

sidekock:
@mats-knmi I have a very weird issue, I don't see the 10 tests appearing like they used to, do you have any idea why this is? Is this because my role changed or did something else go wrong?

mats-knmi:
I think this is because this PR has merge conflicts with master. If you run `git merge origin/master` you should be able to resolve these and then the actions should start again. If that doesn't work you could try closing and re-opening this PR.

sidekock:
Issue with tests, closing pull request and opening a new one

PR #449, opened 2025-01-02
Use the seed in all rng in blending code

mats-knmi:
In some places in the blending code the rng was done without using the provided seed. This meant that a run with a specified seed did not have a completely deterministic outcome and this makes testing harder. This PR fixes this.

NOTE: This PR is based on the refactor branch from @sidekock, so that branch needs to be merged first before this can be merged.
The only changes I added are in this commit: https://github.com/pySTEPS/pysteps/commit/07fbeb08ec297dc7da25f2d74a71645fe0a12fa6.

### DIFF
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -1096,20 +1096,16 @@ class StepsBlendingNowcaster:
     def __initialize_random_generators(self):
         # 6. Initialize all the random generators and prepare for the forecast loop
         """Initialize all the random generators."""
+        seed = self.__config.seed
         if self.__config.noise_method is not None:
             self.__state.randgen_precip = []
-            seed = self.__config.seed
             for j in range(self.__config.n_ens_members):
                 rs = np.random.RandomState(seed)
                 self.__state.randgen_precip.append(rs)
                 seed = rs.randint(0, high=1e9)
-                rs = np.random.RandomState(seed)
-                self.__state.randgen_motion.append(rs)
-                seed = rs.randint(0, high=1e9)
 
         if self.__config.probmatching_method is not None:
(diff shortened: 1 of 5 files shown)


sidekock:
@mats-knmi Did you make any other changes but this issue? It is on my TO-DO list to run both the old and new versions of the code to compare the outputs to make sure that they provide the same output (which will be difficult from what I understand...).

mats-knmi:
In this PR I only made the changes I described (only this commit: https://github.com/pySTEPS/pysteps/commit/07fbeb08ec297dc7da25f2d74a71645fe0a12fa6). Indeed the current blending code is non-deterministic even if you provide a seed.

We could apply this fix to the master branch separately and then also to your refactor branch. Then it should be the case that outputs from 2 runs are identical if you did not make any functional changes.

mats-knmi:
I didn't want to open a PR for any changes to the blending code before you got around to merging your PR, because I expect it would be hard to merge any new changes back into your refactored code. But if it helps you out in validating we could do that.

sidekock:
I think it might be best to add your changes to my PR and then make the changes to the master so I can compare them properly.

mats-knmi:
Ok I will open a new PR for master then. You can merge this PR into your branch when you're ready. I updated the target branch to yours so if you click merge it will go into your branch.

mats-knmi:
Oh wait, I noticed that the tests have issues with the new code. I will update them in this PR still

mats-knmi:
I commited the last fixes so it should be good to go to be merged into your branch @sidekock

mats-knmi:
And I opened a new PR for the master branch: https://github.com/pySTEPS/pysteps/pull/450

Issue #457, opened 2025-02-22
Module distutils is deprecated


RubenImhoff:
The module distutils is deprecated since python v3.12, see also https://docs.python.org/3.10/library/distutils.html. I believe it is only used in `datasets.py` and can be replaced with `shutils.copytree`.

PR #458, opened 2025-02-24
Replace use of distutils by shutils.copytree in datasets.py

RubenImhoff:
As mentioned in issue #457, the module `distutils` is deprecated since python v3.12. I believe it is only used in `datasets.py` and can be replaced with `shutils.copytree`.

This PR attempts to fix the issue by replacing `distutils.dir_util.copy_tree` by `shutils.copytree`.


dnerini (approved):
Very good, thanks @RubenImhoff!

PR #460, opened 2025-02-24
fix noise tapering and error handling

### LINKED ISSUE #451: Initializing noise field with default arguments returns array of zeros. Is this expected behavior?
I am running a probabilistic nowcast for a rectangular precipitation field (199x113) using the steps function with the following settings: 
```
domain = 'spatial'
extrap_method = 'semilagrangian'
decomp_method = 'fft'
fft_method = 'numpy'
noise_method = 'nonparametric'
noise_stddev_adj = 'auto'
```
 and run into the following error before the noise adjustment coefficients are computed: 
```ValueError: field contains non-finite values```. 

This points me to the following statement in the cascade [decomposition](https://github.com/pySTEPS/pysteps/blob/e3325854dae006b1e78eb43a6d2203f2b0b71560/pysteps/cascade/decomposition.py#L193C1-L194C61):
```
if np.any(~np.isfinite(field)):
    raise ValueError("field contains non-finite values") 
```

which in turn points me to ``` decomp_N = decomp_method(N, F, mask=MASK_)``` in the following function: https://github.com/pySTEPS/pysteps/blob/e3325854dae006b1e78eb43a6d2203f2b0b71560/pysteps/noise/utils.py#L24 

This stems from the fact that the noise field F returned in the function:
https://github.com/pySTEPS/pysteps/blob/e3325854dae006b1e78eb43a6d2203f2b0b71560/pysteps/noise/fftgenerators.py#L214

is an array of zeroes, and thus N in the following lines in [noise generation](https://github.com/pySTEPS/pysteps/blob/e3325854dae006b1e78eb43a6d2203f2b0b71560/pysteps/noise/fftgenerators.py#L400C1-L418C75) returns a ```nan```. 

It appears the issue lies in the following lines in the function https://github.com/pySTEPS/pysteps/blob/e3325854dae006b1e78eb43a6d2203f2b0b71560/pysteps/noise/fftgenerators.py#L214
```
F = np.zeros(fft_shape, dtype=complex)
 for i in range(nr_fields):
     if use_full_fft:
         F += fft.fft2(field[i, :, :] * tapering)
     else:
         F += fft.rfft2(field[i, :, :] * tapering)
 F /= nr_fields
```

Using the default arguments, with window_function="tukey" and a field size of 199x113, the tapering returns an oval shape (black for values >0):  
download

This essentially means that if we have a precipitation field, with precipitation only at the edges, and use the default steps settings, this would return a noise field of zeros and eventually lead to an error in the cascade decomposition. 

Though the function appears to numerically behave as expected, is this an error or is this desired behavior? Does this mean it's not possible to compute a nowcast if we have precipitation only around the edges of our domain?

### PR DESCRIPTION
**Pull request by mats-knmi**
Fixes #451

This applies the 2 solutions proposed in the issue. For the error handling I choose a more pragmatic approach than the one I proposed in the issue.

I define the no_precip booleans based on the zero mask of the tapering method provided. This should be exactly what we want, because these booleans are intended to represent whether we can use the specified field for the noise generation right?

@RubenImhoff: can you verify that this will not have unintented side effects?

### DIFF
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -56,6 +56,7 @@ from pysteps import blending, cascade, extrapolation, noise, utils
 from pysteps.nowcasts import utils as nowcast_utils
 from pysteps.postprocessing import probmatching
 from pysteps.timeseries import autoregression, correlation
+from pysteps.utils.check_norain import check_norain
 
 try:
     import dask
@@ -1092,33 +1093,20 @@ class StepsBlendingNowcaster:
 
             self.__precip_models = np.stack(temp_precip_models)
 
-        if self.__params.noise_kwargs["win_fun"] is not None:
-            tapering = utils.tapering.compute_window_function(
-                self.__precip.shape[1],
-                self.__precip.shape[2],
-                self.__params.noise_kwargs["win_fun"],
-            )
-        else:
-            tapering = np.ones((self.__precip.shape[1], self.__precip.shape[2]))
-
-        tapering_mask = tapering == 0.0
-        masked_precip = self.__precip.copy()
-        masked_precip[:, tapering_mask] = np.nanmin(self.__precip)
-        masked_precip_models = self.__precip_models.copy()
-        masked_precip_models[:, :, tapering_mask] = np.nanmin(self.__precip_models)
-
         # Check for zero input fields in the radar and NWP data.
-        self.__params.zero_precip_radar = blending.utils.check_norain(
-            masked_precip,
+        self.__params.zero_precip_radar = check_norain(
+            self.__precip,
             self.__config.precip_threshold,
             self.__config.norain_threshold,
+            self.__params.noise_kwargs["win_fun"],
         )
         # The norain fraction threshold used for nwp is the default value of 0.0,
         # since nwp does not suffer from clutter.
-        self.__params.zero_precip_model_fields = blending.utils.check_norain(
-            masked_precip_models,
+        self.__params.zero_precip_model_fields = check_norain(
+            self.__precip_models,
             self.__config.precip_threshold,
             self.__config.norain_threshold,
+            self.__params.noise_kwargs["win_fun"],
         )
 
     def __zero_precipitation_forecast(self):
@@ -1192,10 +1180,11 @@ class StepsBlendingNowcaster:
             if done:
                 break
             for j in range(self.__precip_models.shape[0]):
-                if not blending.utils.check_norain(
+                if not check_norain(
                     self.__precip_models[j, t],
                     self.__config.precip_threshold,
                     self.__config.norain_threshold,
+                    self.__params.noise_kwargs["win_fun"],
                 ):
                     if self.__state.precip_models_cascades is not None:
                         self.__state.precip_cascades[
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -777,7 +777,7 @@ class StepsBlendingNowcaster:
             self.__params.filter_kwargs = deepcopy(self.__config.filter_kwargs)
 
         if self.__config.noise_kwargs is None:
-            self.__params.noise_kwargs = dict()
+            self.__params.noise_kwargs = {"win_fun": "tukey"}
         else:
             self.__params.noise_kwargs = deepcopy(self.__config.noise_kwargs)
 
@@ -1092,16 +1092,31 @@ class StepsBlendingNowcaster:
 
             self.__precip_models = np.stack(temp_precip_models)
 
+        if self.__params.noise_kwargs["win_fun"] is not None:
+            tapering = utils.tapering.compute_window_function(
+                self.__precip.shape[1],
+                self.__precip.shape[2],
+                self.__params.noise_kwargs["win_fun"],
+            )
+        else:
+            tapering = np.ones((self.__precip.shape[1], self.__precip.shape[2]))
+
+        tapering_mask = tapering == 0.0
+        masked_precip = self.__precip.copy()
+        masked_precip[:, tapering_mask] = np.nanmin(self.__precip)
+        masked_precip_models = self.__precip_models.copy()
+        masked_precip_models[:, :, tapering_mask] = np.nanmin(self.__precip_models)
+
         # Check for zero input fields in the radar and NWP data.
         self.__params.zero_precip_radar = blending.utils.check_norain(
-            self.__precip,
+            masked_precip,
             self.__config.precip_threshold,
             self.__config.norain_threshold,
         )
         # The norain fraction threshold used for nwp is the default value of 0.0,
         # since nwp does not suffer from clutter.
         self.__params.zero_precip_model_fields = blending.utils.check_norain(
-            self.__precip_models,
+            masked_precip_models,
             self.__config.precip_threshold,
             self.__config.norain_threshold,
         )
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -78,7 +78,7 @@ class StepsBlendingConfig:
     precip_threshold: float, optional
       Specifies the threshold value for minimum observable precipitation
       intensity. Required if mask_method is not None or conditional is True.
-    norain_threshold: float, optional
+    norain_threshold: float
       Specifies the threshold value for the fraction of rainy (see above) pixels
       in the radar rainfall field below which we consider there to be no rain.
       Depends on the amount of clutter typically present.
@@ -436,7 +436,6 @@ class StepsBlendingNowcaster:
 
         # Additional variables for time measurement
         self.__start_time_init = None
-        self.__zero_precip_time = None
         self.__init_time = None
         self.__mainloop_time = None
 
@@ -1144,7 +1143,7 @@ class StepsBlendingNowcaster:
                 precip_forecast_workers = None
 
         if self.__config.measure_time:
-            self.__zero_precip_time = time.time() - self.__start_time_init
+            zero_precip_time = time.time() - self.__start_time_init
 
         if self.__config.return_output:
             precip_forecast_all_members_all_times = np.stack(
@@ -1157,8 +1156,8 @@ class StepsBlendingNowcaster:
             if self.__config.measure_time:
                 return (
                     precip_forecast_all_members_all_times,
-                    self.__zero_precip_time,
-                    self.__zero_precip_time,
+                    zero_precip_time,
+                    zero_precip_time,
                 )
             else:
                 return precip_forecast_all_members_all_times
@@ -2929,7 +2928,7 @@ def forecast(
     precip_thr: float, optional
       Specifies the threshold value for minimum observable precipitation
       intensity. Required if mask_method is not None or conditional is True.
-    norain_thr: float, optional
+    norain_thr: float
       Specifies the threshold value for the fraction of rainy (see above) pixels
       in the radar rainfall field below which we consider there to be no rain.
       Depends on the amount of clutter typically present.
--- a/pysteps/blending/utils.py
+++ b/pysteps/blending/utils.py
@@ -553,7 +553,7 @@ def check_norain(precip_arr, precip_thr=None, norain_thr=0.0):
 
     """
     warnings.warn(
-        "pysteps.blending.utils.check_norain has been deprecated, use pysteps.utils.check_norain.check_norain in stead"
+        "pysteps.blending.utils.check_norain has been deprecated, use pysteps.utils.check_norain.check_norain instead"
     )
     return new_check_norain(precip_arr, precip_thr, norain_thr, None)
 
--- a/pysteps/blending/utils.py
+++ b/pysteps/blending/utils.py
@@ -15,11 +15,11 @@ Module with common utilities used by the blending methods.
     decompose_NWP
     compute_store_nwp_motion
     load_NWP
-    check_norain
     compute_smooth_dilated_mask
 """
 
 import datetime
+import warnings
 from pathlib import Path
 
 import numpy as np
@@ -28,6 +28,7 @@ from pysteps.cascade import get_method as cascade_get_method
 from pysteps.cascade.bandpass_filters import filter_gaussian
 from pysteps.exceptions import MissingOptionalDependency
 from pysteps.utils import get_method as utils_get_method
+from pysteps.utils.check_norain import check_norain as new_check_norain
 
 try:
     import netCDF4
@@ -534,7 +535,7 @@ def load_NWP(input_nc_path_decomp, input_path_velocities, start_time, n_timestep
 
 def check_norain(precip_arr, precip_thr=None, norain_thr=0.0):
     """
-
+    DEPRECATED use :py:mod:`pysteps.utils.check_norain.check_norain` in stead
     Parameters
     ----------
     precip_arr:  array-like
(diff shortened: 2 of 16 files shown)

### CODE REVIEW

File: pysteps/blending/steps.py [hunk 1]
```diff
@@ -1092,16 +1092,31 @@ def transform_to_lagrangian(precip, i):
 
             self.__precip_models = np.stack(temp_precip_models)
 
+        if self.__params.noise_kwargs["win_fun"] is not None:
+            tapering = utils.tapering.compute_window_function(
+                self.__precip.shape[1],
+                self.__precip.shape[2],
+                self.__params.noise_kwargs["win_fun"],
+            )
+        else:
```
RubenImhoff: Can the noise_kwargs be None, given the following earlier in the code: (?)

```
        if self.__config.noise_kwargs is None:
            self.__params.noise_kwargs = {"win_fun": "tukey"}
```

RubenImhoff: The idea is to just fill up the tapering with ones (instead of a function) if None is given?

RubenImhoff (commented):
Hi @mats-knmi, fantastic, thanks a lot for picking this up! This seems to work, I think. Could you maybe also plot what the tapering looks like now? @bwalraven, would you like to have a look at it as well?

pysteps/blending/steps.py
mats-knmi: `noise_kwargs` cannot be `None`, but `noise_kwargs["win_fun"]` can be set to `None` by the user.

Yes filling the tapering with ones if the `win_fun` is `None` is what is done in the fftgenerators as well: https://github.com/pySTEPS/pysteps/blob/006fb49bd01385e6e18a3e29027c502c76fb38c9/pysteps/noise/fftgenerators.py#L288-L293
In this case it doesn't really matter what we set it to, we just want to have an emtpy mask so that we don't remove any values from the original precip field.

mats-knmi:
> Hi @mats-knmi, fantastic, thanks a lot for picking this up! This seems to work, I think. Could you maybe also plot what the tapering looks like now? @bwalraven, would you like to have a look at it as well?

This is what the tukey tapering looks like now for our domain (765x700):
Figure_4
And this is what it would look like for @bwalraven 's domain (199x113):
Figure_3
I tried to stretch the images so that x and y are the same scale.

RubenImhoff:
Good and when `win_fun` is set to None, all are set to ones? (Is that recommended in this case?)

RubenImhoff:
Besides the former questions, good to go from my side, @mats-knmi, great work!

RubenImhoff (approved):

mats-knmi:
Just to be sure @RubenImhoff, did you check that the other locations in the blending code that use `zero_precip_radar` are not harmed by this change. I am not sure what these parts of the code do and if they would be negatively impacted:
https://github.com/pySTEPS/pysteps/blob/76539bb00110d5d4643f12b148d45c85041c8efe/pysteps/blending/steps.py#L1333-L1340
https://github.com/pySTEPS/pysteps/blob/76539bb00110d5d4643f12b148d45c85041c8efe/pysteps/blending/steps.py#L2118-L2126
https://github.com/pySTEPS/pysteps/blob/76539bb00110d5d4643f12b148d45c85041c8efe/pysteps/blending/steps.py#L2160-L2168

bwalraven:
Thanks @mats-knmi! This would indeed solve 'my' issue in #451! 

Perhaps a 'nice-to-have', with regards to @dnerini's comment about [handling zero fields more graciously instead of raising an error](https://github.com/pySTEPS/pysteps/issues/451#issuecomment-2590905710) I am wondering whether it would be useful to also generate a warning or more informative error message in the case that `win_fun` is not `None`, but `precip_field * tapering` does return a field of zeros? 

This might not happen very often now that the adapted `tukey` function also covers the edges of a more rectangular grid, but in the case that it does happen, it would point the user more directly to the problem without having to dig through the code extensively. Alternatively this could also be done by a "no-rain check" like @mats-knmi proposed in [point 2: error handling](https://github.com/pySTEPS/pysteps/issues/451#issuecomment-2677933774).

In any case, thanks for proposing a solution that solves the issue!

RubenImhoff:
> Just to be sure @RubenImhoff, did you check that the other locations in the blending code that use `zero_precip_radar` are not harmed by this change. I am not sure what these parts of the code do and if they would be negatively impacted:
> 
> https://github.com/pySTEPS/pysteps/blob/76539bb00110d5d4643f12b148d45c85041c8efe/pysteps/blending/steps.py#L1333-L1340
> 
> 
> https://github.com/pySTEPS/pysteps/blob/76539bb00110d5d4643f12b148d45c85041c8efe/pysteps/blending/steps.py#L2118-L2126
> 
> 
> https://github.com/pySTEPS/pysteps/blob/76539bb00110d5d4643f12b148d45c85041c8efe/pysteps/blending/steps.py#L2160-L2168

I have not tested it on real data yet (I would need to have a couple of good rain-no rain cases for that), but from what I can see, it will mainly change when we have or don't have `no_rain`, which should still give a stable result (and probably goes to a no rain alternative sooner, which is not bad in these cases, I think). Perhaps @bwalraven, could test it for a couple of his failing cases?

mats-knmi:
> Thanks @mats-knmi! This would indeed solve 'my' issue in #451!
> 
> Perhaps a 'nice-to-have', with regards to @dnerini's comment about [handling zero fields more graciously instead of raising an error](https://github.com/pySTEPS/pysteps/issues/451#issuecomment-2590905710) I am wondering whether it would be useful to also generate a warning or more informative error message in the case that `win_fun` is not `None`, but `precip_field * tapering` does return a field of zeros?
> 
> This might not happen very often now that the adapted `tukey` function also covers the edges of a more rectangular grid, but in the case that it does happen, it would point the user more directly to the problem without having to dig through the code extensively. Alternatively this could also be done by a "no-rain check" like @mats-knmi proposed in [point 2: error handling](https://github.com/pySTEPS/pysteps/issues/451#issuecomment-2677933774).
> 
> In any case, thanks for proposing a solution that solves the issue!

The solution I proposed makes it so that if `precip_field * tapering` would be a field of zeros, then `zero_precip_radar` will be `True`. This means that the code that throws the error will not be executed. This should be enough for this error to not come back.

bwalraven:
> I have not tested it on real data yet (I would need to have a couple of good rain-no rain cases for that), but from what I can see, it will mainly change when we have or don't have `no_rain`, which should still give a stable result (and probably goes to a no rain alternative sooner, which is not bad in these cases, I think). Perhaps @bwalraven, could test it for a couple of his failing cases?

I am not familiar enough with the blending steps to say something about that, but I can have a look and test it for the simple nowcasting cases I have. But seeing @mats-knmi's resulting image of the adapted `tukey` function, and especially if the default tapering function becomes `None` i.e. `np.ones()` rather than `tukey` this should solve the issue I encountered when running the [steps function](https://github.com/pySTEPS/pysteps/blob/master/pysteps/nowcasts/steps.py).

mats-knmi:
The default will still be `tukey`, you can set it to `None` yourself if you want, but that is not recommended as @dnerini said.

pysteps/utils/check_norain.py
```diff
@@ -0,0 +1,51 @@
+import numpy as np
+
+from pysteps import utils
+
+
+def check_norain(precip_arr, precip_thr=None, norain_thr=0.0, win_fun="tukey"):
```
dnerini: I'd suggest to leave `win_fun=None` as default, so to maintain the same behaviour. The default can then be changed in the call in the blending method

pysteps/blending/utils.py
```diff
@@ -28,6 +28,7 @@
 from pysteps.cascade.bandpass_filters import filter_gaussian
 from pysteps.exceptions import MissingOptionalDependency
 from pysteps.utils import get_method as utils_get_method
+from pysteps.utils.check_norain import check_norain as new_check_norain
```
dnerini, line 31: Is this really necessary? If the default behaviour of `check_norain` method doesn't change with respect to the original method, it'd be enough to import it here, remove the older one, without need for a deprecation warning. Do you agree?

dnerini (approved):
hi @mats-knmi, thanks a lot for taking this up! 

The changes to the tapering windows make a lot of sense to me and definitely improve over the current version, very nice!

For the approach with zero precip, I agree that the best way to handle this is to account for the effect of the tapering window on the norain check. I only left a very small comment concerning the deprecation warning, which I think it could be safely removed.

dnerini:
@mats-knmi commit [ee9a787](https://github.com/pySTEPS/pysteps/pull/460/commits/ee9a78732448737f8b5dc5b9ba7d8994aacad0f0) adds a test for the steps nowcast method with all zero inputs (it will fail), as a reminder that something needs to be still done for the use case, as discussed in #451.

pysteps/utils/check_norain.py
mats-knmi: Hmmm, yes but in the fftgenerators the default method is `tukey`. So if a user uses an ensemble nowcast with default settings and then also the `check_norain` method with default settings, you would expect it to work. Which it won't if the default is different here as in the fftgenerators. It would be ideal if they could somehow be synced, but I don't have a good idea for that right now.

pysteps/blending/utils.py
mats-knmi: This is a good point. This does make it slightly more explicit to users that they should no longer import it from here though. But I would be ok with simplifying this into just importing `check_norain` here.

mats-knmi:
@dnerini Do you want me to add the `check_norain` to the beginning of the steps nowcast? This way it should not go wrong anymore for new users?

And what should I do with the other 2 nowcast methods that use noise generators? I see linda only uses pbs, but sseps could use fft generated noise as well right?

pysteps/utils/check_norain.py
dnerini: agreed, but in principle `check_norain` is indipendent of the settings in the fftgenerators. So it'll be important to use the right win_fun argument in the nowcasting methods, but in itself I would still set the default to None here. Or am I missing on something?

pysteps/blending/utils.py
dnerini: you also have a good point! Let's leave it!

dnerini:
> @dnerini Do you want me to add the `check_norain` to the beginning of the steps nowcast? This way it should not go wrong anymore for new users?

yes, I think this could be a good approach. What do you do with the blending? return an empty, all zero output?

> And what should I do with the other 2 nowcast methods that use noise generators? I see linda only uses pbs, but sseps could use fft generated noise as well right?

in principle yes, but maybe a better approach is to include a dedicated test for these methods too, then we'll need to include the check only for those methods that fail?

pysteps/utils/check_norain.py
mats-knmi: The reason I set the tapering method to `tukey` by default is because this `check_norain` method currently only exists for the purpose of running the fftgenerators. Since for the fftgenerators the `tukey` method is the default, I thought it would make sense to have it as the default here as well.

Ofcourse the `check_norain` method is a quite generic method that could be used for other purposes as well, in which case a default of `None` would make more sense.

I think that if we call `check_norain` from the nowcast methods, so that the user doesn't have to do that anymore, it should be fine to set the default to `None`. But then we would have to update the error message in the fftgenerators as well to indicate to a user like @SKB-7 (who uses the fftgenerators directly) that in order to successfully check for no rain in the context of the fftgenerators, you will have to call `check_norain` with the correct tapering method (which is `tukey` by default for the fftgenerators).

mats-knmi: I think the best approach is to do what I typed in the last paragraph of my previous message, so I will implement that today.

mats-knmi:
@dnerini I saw your new tests, thanks. I will implement the no rain check in all the nowcast methods that fail these tests and that should fix them. I will return only zeros when the norain check fails indeed.

dnerini:
Note that currently the failing `test_plugins_support.py` is outside the scope of this PR. I believe this is caused by changes in https://github.com/pySTEPS/cookiecutter-pysteps-plugin that will be addressed in https://github.com/pySTEPS/pysteps/pull/418. Maybe @ladc can quickly confirm that this is indeed the case?

mats-knmi:
@dnerini I think I have fixed everything now, can you review the final version? If you are on board I would like to merge this and release this as a new version, since we want to start using this operationally. How do I do this? Do I just bump the version in the setup.py file and then merge this pull request?

pysteps/blending/utils.py
```diff
@@ -551,15 +552,10 @@ def check_norain(precip_arr, precip_thr=None, norain_thr=0.0):
...
     """
-
-    if precip_thr is None:
-        precip_thr = np.nanmin(precip_arr)
-    rain_pixels = precip_arr[precip_arr > precip_thr]
-    norain = rain_pixels.size / precip_arr.size <= norain_thr
-    print(
-        f"Rain fraction is: {str(rain_pixels.size / precip_arr.size)}, while minimum fraction is {str(norain_thr)}"
+    warnings.warn(
+        "pysteps.blending.utils.check_norain has been deprecated, use pysteps.utils.check_norain.check_norain in stead"
```
dnerini:
```suggestion
        "pysteps.blending.utils.check_norain has been deprecated, use pysteps.utils.check_norain.check_norain instead"
```

pysteps/nowcasts/sseps.py
```diff
@@ -188,13 +187,7 @@ def forecast(
 
     See also
     --------
-    pysteps.extrapolation.interface, pysteps.cascade.interface,
-    pysteps.noise.interface, pysteps.noise.utils.compute_noise_stddev_adjs
-
-    Notes
-    -----
-    Please be aware that this represents a (very) experimental implementation.
```
dnerini: why remove this note? this is still a very experimental method ...

dnerini (approved):
Looks very good to me! Left only two super minor comments, then I believe this is ready to be merged

pysteps/nowcasts/sseps.py
mats-knmi: I don't know how this happened. I did not mean to remove this part also I did not mean to add this line
```
pysteps.extrapolation.interface, pystepsemilagrangian
```
I am pretty confused, but I re-added this note.

mats-knmi:
@dnerini I will merge this then, how do I turn this into a release? Or is there something I need to wait on before doing that?

dnerini:
> I would like to merge this and release this as a new version, since we want to start using this operationally. How do I do this? Do I just bump the version in the setup.py file and then merge this pull request?

So what I typically do:

1. merge PR
2. bump version directly on master (see e.g https://github.com/pySTEPS/pysteps/commit/ee60fa6e73dbf6157d494dec6443aa8ac37537b6)
3. release new version (use automatic release notes). In this case, we probably want to release a new minor version, ie `v1.15.0`
4. check that the release github actions succesfully completes, eg https://github.com/pySTEPS/pysteps/actions/runs/13141012118

do you rely on the conda-forge installation? this typically takes few extra days (if everything goes well)

mats-knmi:
Thanks, I'll try that!
> do you rely on the conda-forge installation? this typically takes few extra days (if everything goes well)

No we install pysteps through pip

mats-knmi:
@dnerini I cannot push directly to master (which is a good thing imo), but this means I cannot bump the version directly. So either you would have to do this for me or you could approve this PR: https://github.com/pySTEPS/pysteps/pull/462.

So maybe next time I bump the version directly in the feature PR or how do we do this?

RubenImhoff:
Great work!!

ladc:
> Note that currently the failing `test_plugins_support.py` is outside the scope of this PR. I believe this is caused by changes in https://github.com/pySTEPS/cookiecutter-pysteps-plugin that will be addressed in #418. Maybe @ladc can quickly confirm that this is indeed the case?

Yes, the merge request fixing this is #405.

PR #461, opened 2025-02-25
Use max rain pixels for timestep and member to decompose for precip cascades in case of zero radar

mats-knmi:
This is something I noticed while working on the other PR. But I think this makes more sense than what I did during the memory reduction PR.

I used to look for a timestep and member that has any rain so that that can be used to decompose the model precip. But if we just use the timestep and member that has the most rainy pixels, this should be even better right? Since this was already computed a bit further down the line, this should be a simplification of the code without any significant side effects.

Let me know what you think.

### DIFF
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -324,6 +324,7 @@ class StepsBlendingParams:
     precip_models_provided_is_cascade: bool = False
     xy_coordinates: np.ndarray | None = None
     precip_zerovalue: float | None = None
+    precip_threshold: float | None = None
     mask_threshold: np.ndarray | None = None
     zero_precip_radar: bool = False
     zero_precip_model_fields: bool = False
@@ -803,6 +804,8 @@ class StepsBlendingNowcaster:
         else:
             self.__params.mask_kwargs = deepcopy(self.__config.mask_kwargs)
 
+        self.__params.precip_threshold = self.__config.precip_threshold
+
         if np.any(~np.isfinite(self.__velocity)):
             raise ValueError("velocity contains non-finite values")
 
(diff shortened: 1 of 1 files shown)


dnerini:
> I used to look for a timestep and member that has any rain so that that can be used to decompose the model precip. But if we just use the timestep and member that has the most rainy pixels, this should be even better right? 

Makes sense to me! Is there any equivalent logic in the radar-only nowcast method that could benefit from this too?

sidekock:
I agree, this is much cleaner. 
Could you also change the following while you are changing the code @mats-knmi :

`        # Initialize noise based on the NWP field time step where the fraction of rainy cells is highest
        if self.__config.precip_threshold is None:
            self.__config.precip_threshold = np.nanmin(self.__precip_models)
`

Here we overwrite the config in the same function you are touching. However, StepsBlendingConfig is frozen so should not be written to. A new variable should be made in StepsBlendingParams and we should look where in the code after this point self.__config.precip_threshold is used, it should be changed for self.__params.precip_threshold

mats-knmi:
@dnerini As far as I know this is a blending only functionality as this uses the model data to create noise when the radar data has no precipitation. This is ofcourse not possible in a radar only nowcast.

@sidekock I added a commit. Is this what you meant?

sidekock:
@mats-knmi indeed, that is what I was aiming for. Weird the tests are faling on ".tests\test_plugins_support.py::test_importers_plugins - AssertionError: assert False"...

mats-knmi:
@sidekock This has nothing to do with this PR, apparently something with the cookiecutter plugin (see https://github.com/pySTEPS/pysteps/pull/460#issuecomment-2694378126).

ladc:
Yes, our bad, sorry! Working on it.

dnerini:
@mats-knmi can you please check that I solved the merge conflicts correctly?

mats-knmi:
> @mats-knmi can you please check that I solved the merge conflicts correctly?

Yes this looks good. This should be ready to merge now.

dnerini (approved):

PR #466, opened 2025-04-17
Fixes to linear and saliency based blending

### LINKED ISSUE #467: Saliency-based blending gives direct jump from nowcast to NWP
See the figure below from our Gallery, where it becomes clear that the saliency-based blending at +80 min becomes directly the same as NWP, while that is not warranted, as this is the time where the saliency-based blending should have just started. This seems to be caused by nans following from the _get_ranked_salience function.

Image

RubenImhoff:
Fixes #467 

PR to ensure that linear blending method can also handle multiple timesteps in the `precip` input variable, and to ensure that saliency-based blending works well when there are nans in the forecast.

- [x] Ensure linear blending can also handle multiple input timesteps.
- [x] Ensure that saliency-based blending works well when there are nans in the forecast. See the figure below from our Gallery, where it becomes clear that the saliency-based blending at +80 min becomes directly the same as NWP, while that is not warranted, as this is the time where the saliency-based blending should have just started. This seems to be caused by nans following from the `_get_ranked_salience` function.
- [x] Add tests (if needed).

image

### DIFF
--- a/pysteps/blending/linear_blending.py
+++ b/pysteps/blending/linear_blending.py
@@ -177,6 +177,17 @@ def forecast(
             precip_nwp.shape, precip_nowcast.shape
         )
 
+        # Ensure we are not working with nans in the bleding.
+        # Check if the NWP data contains any nans. If so, fill them with 0.0.
+        precip_nwp = np.nan_to_num(precip_nwp, nan=0.0)
+
+        # Fill nans in precip_nowcast
+        nan_indices = np.isnan(precip_nowcast)
+        if fill_nwp:
+            precip_nowcast[nan_indices] = precip_nwp[nan_indices]
+        else:
+            precip_nowcast[nan_indices] = 0.0
+
         # Initialise output
         precip_blended = np.zeros_like(precip_nowcast)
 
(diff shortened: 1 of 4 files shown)


pysteps/blending/linear_blending.py
```diff
@@ -173,6 +177,17 @@ def forecast(
...
+        # Ensure we are not working with nans in the bleding.
+        # Check if the NWP data contains any nans. If so, fill them with 0.0.
+        precip_nwp = np.nan_to_num(precip_nwp, nan=0.0)
+
+        # Fill nans in precip_nowcast
+        nan_indices = np.isnan(precip_nowcast)
+        if fill_nwp:
+            precip_nowcast[nan_indices] = precip_nwp[nan_indices]
+        else:
+            precip_nowcast[nan_indices] = 0.0
```
mats-knmi:
```suggestion
        nan_mask = np.isnan(precip_nowcast)
        if fill_nwp:
            precip_nowcast[nan_mask] = precip_nwp[nan_mask]
        else:
            precip_nowcast[nan_mask] = 0.0
```
Nitpicky, sorry. But `np.isnan` returns a mask not a list of indices.

mats-knmi: Is filling with 0.0 not an issue if you are nowcasting in dBZ?

RubenImhoff: Good suggestion, thanks!

RubenImhoff: This should be fine, as in the linear blending procedure, we always calculate in mm/h (the conversion is done earlier on in the code).

mats-knmi (approved):

PR #470, opened 2025-05-20
fix: speed up linear blending through more efficient nowcasting step

### LINKED ISSUE #469: Speed up linear blending through more efficient nowcast calculation
In the current linear blending setup, the initial nowcast is calculated for the full forecast horizon, even though the nowcast is only needed until `end_blending` time after which the forecast will entirely consist of the NWP forecast. This works, but is not very efficient. With this issue, we can try to implement the linear blending setup in such a way that only a nowcast up to `end_blending` is required.

RubenImhoff:
Fixes #469.

By only calculating a nowcast up to the `end_blending` time, we can make the linear blending code more efficient, particularly when longer forecast horizons are requested. 

This fix required a few changes in the linear blending code. Good to give it a thorough review.

### DIFF
--- a/pysteps/blending/linear_blending.py
+++ b/pysteps/blending/linear_blending.py
@@ -109,20 +109,26 @@ def forecast(
     if len(precip.shape) == 3:
         precip = precip[-1, :, :]
 
-    # Calculate the nowcasts
-    nowcast_method_func = nowcasts.get_method(nowcast_method)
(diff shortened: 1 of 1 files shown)


dnerini (approved):
Looking good! thanks for the contribution :)

NathalieRombeek (approved):
Looks good, thanks for the work Ruben!

Issue #471, opened 2025-05-22
Occasional crashes in LINDA nowcasts


pulkkins:
LINDA nowcasts are occasionally crashing with the following errors:
```
Traceback (most recent call last):
  File "/home/users/spulkkin/tmp/pysteps_data_ruben/test.py", line 11, in <module>
    nowcast = linda.forecast(precip, velocity, 5, add_perturbations=True, kmperpixel=1, timestep=5, feature_method="domain")
  File "/home/users/spulkkin/ohjelmistokehitys/pySTEPS/pysteps/build/lib.linux-x86_64-cpython-313/pysteps/nowcasts/linda.py", line 331, in forecast
    pert_gen = _linda_perturbation_init(
        precip,
    ...<13 lines>...
        measure_time,
    )
  File "/home/users/spulkkin/ohjelmistokehitys/pySTEPS/pysteps/build/lib.linux-x86_64-cpython-313/pysteps/nowcasts/linda.py", line 1348, in _linda_perturbation_init
    pert_gen = _estimate_perturbation_params(
        err,
    ...<6 lines>...
        use_multiprocessing,
    )
  File "/home/users/spulkkin/ohjelmistokehitys/pySTEPS/pysteps/build/lib.linux-x86_64-cpython-313/pysteps/nowcasts/linda.py", line 829, in _estimate_perturbation_params
    r = worker(i)
  File "/home/users/spulkkin/ohjelmistokehitys/pySTEPS/pysteps/build/lib.linux-x86_64-cpython-313/pysteps/nowcasts/linda.py", line 811, in worker
    return distpar, std, np.sqrt(np.abs(np.fft.rfft2(acf)))
                                        ~~~~~~~~~~~~^^^^^
  File "/home/users/spulkkin/miniconda3/envs/pysteps/lib/python3.13/site-packages/numpy/fft/_pocketfft.py", line 1464, in rfft2
    return rfftn(a, s, axes, norm, out=out)
  File "/home/users/spulkkin/miniconda3/envs/pysteps/lib/python3.13/site-packages/numpy/fft/_pocketfft.py", line 1380, in rfftn
    s, axes = _cook_nd_args(a, s, axes)
              ~~~~~~~~~~~~~^^^^^^^^^^^^
  File "/home/users/spulkkin/miniconda3/envs/pysteps/lib/python3.13/site-packages/numpy/fft/_pocketfft.py", line 704, in _cook_nd_args
    s = take(a.shape, axes)
[...]

PR #473, opened 2025-06-11
Fix crash in LINDA when perturbation parameters cannot be estimated

pulkkins:
Pull request to fix issue https://github.com/pySTEPS/pysteps/issues/471.


RubenImhoff (approved):
Thanks, Seppo! I've already locally tested the fixes and they solve the mentioned issues for me. Good to go, I think!

Issue #464, opened 2025-03-28
Implement Reduced-Space Ensemble Kalman Filter Approach for Blending


RubenImhoff:
Next to the already implemented STEPS blending approach (Bowler et al., 2006; Seed et al., 2013 and our paper Imhoff et al., 2023), we thought it was getting time to introduce another blending method based on [Daniele's Monthly Weather Review paper from 2019](https://journals.ametsoc.org/view/journals/mwre/147/3/mwr-d-18-0258.1.xml). This blending approach uses a Bayesian blending approach, by using a reduced-space ensemble Kalman filter that is flow-dependent to combine extrapolation-based nowcasts and NWP. 

@m-rempel (and colleagues) has continued with this approach in his research to make it ready for operations. This seems like an excellent time to include this approach in pysteps, I've discussed with @m-rempel to get started with this in the coming months. @m-rempel will make a new branch in which we can start working. We can use this issue (among others) for discussion about the approach.

ladc:
Hi Ruben, 
Yes indeed, and it is also a good moment to start thinking about issues to tackle or finish during the pysteps Hackathon 😊 @sidekock  suggested to create a project board already, which I think is a good idea.

PR #483, opened 2025-06-25
Add importer for DWD radar data products and reprojection onto a regular grid as part of #464

m-rempel:
Implementation of an importer for DWD radar data products and a reprojection onto a regular grid. These are necessary to test the implementation of the reduced-space EnKF combination approach (Nerini et al., 2019) and to compare the results with the already existing code base that uses DWD data.

### DIFF
--- a/pysteps/io/importers.py
+++ b/pysteps/io/importers.py
@@ -1656,6 +1656,13 @@ def import_dwd_hdf5(filename, qty="RATE", **kwargs):
             "but it is not installed"
         )
 
+    if not PYPROJ_IMPORTED:
+        raise MissingOptionalDependency(
+            "pyproj package is required to import "
+            "DWD's radar reflectivity composite "
+            "but it is not installed"
+        )
+
     if qty not in ["ACRR", "DBZH", "RATE"]:
         raise ValueError(
             "unknown quantity %s: the available options are 'ACRR', 'DBZH' and 'RATE'"
@@ -1740,11 +1747,13 @@ def import_dwd_hdf5(filename, qty="RATE", **kwargs):
         transform = None
 
     startdate = datetime.datetime.strptime(
-        f"{file_content['dataset1']['what']['startdate']}{file_content['dataset1']['what']['starttime']}",
+        file_content["dataset1"]["what"]["startdate"]
+        + file_content["dataset1"]["what"]["starttime"],
         "%Y%m%d%H%M%S",
     )
     enddate = datetime.datetime.strptime(
-        f"{file_content['dataset1']['what']['enddate']}{file_content['dataset1']['what']['endtime']}",
+        file_content["dataset1"]["what"]["enddate"]
+        + file_content["dataset1"]["what"]["endtime"],
         "%Y%m%d%H%M%S",
     )
     accutime = (enddate - startdate).total_seconds() / 60.0
@@ -1762,7 +1771,7 @@ def import_dwd_hdf5(filename, qty="RATE", **kwargs):
         "xpixelsize": xpixelsize,
         "ypixelsize": ypixelsize,
         "cartesian_unit": "m",
-        "yorigin": "lower",
+        "yorigin": "upper",
         "institution": file_content["what"]["source"],
         "accutime": accutime,
         "unit": unit,
@@ -1962,7 +1971,7 @@ def _import_dwd_geodata(product, dims):
     geodata["xpixelsize"] = 1000.0
     geodata["ypixelsize"] = 1000.0
     geodata["cartesian_unit"] = "m"
-    geodata["yorigin"] = "lower"
+    geodata["yorigin"] = "upper"
 
     prod_cat1 = ["RX", "RY", "RW"]  # 900x900
     prod_cat2 = ["WN"]  # 1200x1100
--- a/pysteps/io/importers.py
+++ b/pysteps/io/importers.py
@@ -82,11 +82,14 @@ Available Importers
     import_odim_hdf5
     import_opera_hdf5
     import_saf_crri
+    import_dwd_hdf5
+    import_dwd_radolan
 """
 
 import gzip
 import os
 import array
+import datetime
 from functools import partial
 
 import numpy as np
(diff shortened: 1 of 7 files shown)


RubenImhoff:
Many thanks for this contribution, @m-rempel! 

Although we have discussed within the pysteps community to limit the number of new importers, I can imagine that this importer is a useful addition to allow for the ensemble Kalman filter blending setup, testing and possibly some gallery examples. I will have a quick look to see if there is any overlap with other HDF5 importers to see if we can shorten the code a bit. @dnerini, what do you think?

Regarding the addition to the reprojection, the code looks good, but at this moment pysteps does not yet support xarray (we will from version 2). Is there a way to rewrite this approach without xarray? If not, we could consider these improvements as part of the pysteps v2 release. @dnerini, what is your take on this?

pysteps/io/importers.py
```diff
@@ -1617,3 +1621,384 @@ def _import_saf_crri_geodata(filename):
...
+
+                d[key] = np.array(value)
+
+            except TypeError:
+
+                d[key] = np.array(value.astype(float))
+
+            except:
+
+                d[key] = value.value
```
RubenImhoff: What exception are you expecting here?

pysteps/io/importers.py
```diff
@@ -1617,3 +1621,384 @@ def _import_saf_crri_geodata(filename):
...
+    metadata.update(kwargs)
+
+    f.close()
+
+    return precip, quality, metadata
+
+
+def _read_hdf5_cont(f, d):
+    """
+    Recursively read nested dictionaries from a HDF5 file.
```
RubenImhoff, line 1875: Good to add some extra docstrings here about what the input variables are and what you expect to return.

pysteps/io/importers.py
```diff
@@ -1617,3 +1621,384 @@ def _import_saf_crri_geodata(filename):
...
+    data[no_data_mask] = np.nan
+
+    geodata = _import_dwd_geodata(product, dims)
+    metadata = geodata
+
+    return data, None, metadata
+
+
+def _identify_info_bits(data):
+
```
RubenImhoff, line 2064: Could you also add some docstrings here?

pysteps/io/importers.py
```diff
@@ -1617,3 +1621,384 @@ def _import_saf_crri_geodata(filename):
...
+@postprocess_import()
+def import_dwd_radolan(filename, product):
+    """
+    Import a RADOLAN precipitation product from a binary file.
+
+    Parameters
+    ----------
+    filename: str
+        Name of the file to import.
+    product: {'WX','RX','EX','RY','RW','AY','RS','YW','WN'}
```
RubenImhoff: Would it make sense to call this variable `product_name`? Then the `prod` variable can get a longer, more descriptive name as well.

pysteps/io/importers.py
```diff
@@ -1617,3 +1621,384 @@ def _import_saf_crri_geodata(filename):
...
+
+    return data, no_data_mask
+
+
+def _import_dwd_geodata(product, dims):
+    """
+    Geodata for RADOLAN products. Hard-coded here, cause there is only basic
+    information about the projection in the header of the binary RADOLAN
+    files.
+    """
```
RubenImhoff, line 2098: Add some more docstrings here, too (input parameters and output). I have already added much more docstrings, but feel free to edit them from here.

RubenImhoff:
Almost forgot to mention it, but we will have to add some tests if we continue with this approach! :)

RubenImhoff:
Many thanks, @m-rempel! @dnerini, what is your take on adding one more importer to pysteps for the sake of testing and implementing the reduced-space Ensemble Kalman filter based blending?

pysteps/io/importers.py
m-rempel: I have removed this exception, since I had implemented this in my code for testing purposes. However, to read the HDF content np.array should be sufficient.

pysteps/io/importers.py
m-rempel: I have added here and in other functions the docstrings and have extended the descriptions.

pysteps/io/importers.py
m-rempel: I followed your suggestion and changed it in product_name.

m-rempel:
> Almost forgot to mention it, but we will have to add some tests if we continue with this approach! :)

Sure. This is the first version of the implementation. Also to see if you are happy with it.

RubenImhoff:
> > Almost forgot to mention it, but we will have to add some tests if we continue with this approach! :)
> 
> Sure. This is the first version of the implementation. Also to see if you are happy with it.

Perfect! I think the importer code looks good now. Could you add a data sample of the DWD data to https://github.com/pySTEPS/pysteps-data? Once it is there, we have a dataset we can use for setting up some simple tests. 

Other than that, and if @dnerini also agrees with this setup, we are good to go. :)

dnerini:
hi @RubenImhoff and @m-rempel, apologies for the slow response...

It's true that in principle we prefer to use plugins for importers instead of adding them into pysteps itself. This said, I'm personally OK with a few exceptions like in this case, especially if it can help implementing a new cool feature ;) but the new importers should come with some basic testing at least. Can we add some tests for the `import_dwd_hdf5` and `unstructured2regular` methods before merging?

m-rempel:
Hi @dnerini !
In our last meeting, @RubenImhoff and I discussed the issue about the missing tests. I'll implement these by no later than next week.

pysteps/tests/test_utils_reprojection.py
```diff
@@ -109,4 +112,67 @@ def test_utils_reproject_grids(
...
+_, _, metadata_dst = pysteps.io.import_dwd_hdf5(filename_radar, qty="RATE")
+
+filename_nwp = os.path.join(
+    root_path_nwp, "2025", "06", "04", "20250604_1600_PR_GSP_060_120.grib2"
+)
+kwargs = {
+    "varname": "PR_GSP",
+    "grid_file_path": "./aux/grid_files/dwd/icon/R19B07/icon_grid_0047_R19B07_L.nc",
+}
+array_src, _, metadata_src = pysteps_nwp_importers.importer_dwd_nwp.import_dwd_nwp(
```
RubenImhoff: @m-rempel, `pysteps` won't find `pysteps_nwp_importers` as it is an optional additional package to `pysteps`. Hence, this test would only work with dummy data, or if you can make a very simple reproduction of the data import here in the tests. As you can see in the reprojection test above for the RMI data, some dummy data is created an put on the grid and projection that this data has. Would something like that be feasible? Other than that, this PR is good to go!

m-rempel: I thought about this also yesterday. One option could be to have `unstructured2regular()` within `pysteps_nwp_importers`, since it is needed solely for the ICON-D2 data. On the other side, `pysteps_nwp_importers` would then no longer be independent, since some regular grid is necessary for that test.
What do you think, @RubenImhoff and @dnerini?

RubenImhoff: I see that I have totally forgotten to respond to this question, sorry! Since the `unstructured2regular()` function is only used for ICON-D2 (and it indeed is something that likely will only be used for ICON), I am fine with moving it to `pysteps_nwp_importers`. To keep `pysteps_nwp_importers` independent, would it work to make a test with dummy data there on a regular grid (the grid size, extent and projection can of course still be identical to the DWD data).

m-rempel: Thanks @RubenImhoff! I have now moved `unstructured2regular()` to `pysteps_nwp_importers` and adjusted the tests in pysteps accordingly.

RubenImhoff (approved):

RubenImhoff:
I think we are ready to go - great work @m-rempel!

PR #486, opened 2025-07-06
Drop python 3.10, support python 3.13

dnerini:
Following https://scientific-python.org/specs/spec-0000/


RubenImhoff (approved):
Looks good to me, nice work, @dnerini!

PR #489, opened 2025-07-09
Update installation docs for gcc version consistency

kafitzgerald:
Minor updates to the install docs so they refer to a consistent version of gcc.


dnerini (approved):
Thanks for the contribution!

PR #491, opened 2025-09-02
Add first version of the reduced-space ensemble Kalman filter method as part of #464

m-rempel:
Hi @RubenImhoff and @dnerini!

Here's my first version of the implementation of the reduced-space ensemble Kalman filter method. By calling `blending_method = blending.get_method("pca_enkf")`

One can now compute a combined forecast by pass arrays of observed precipitation fields, the NWP ensemble model forecast, the motion vector field based on observations as well as the timestamps corresponding to the observations and NWP forecast, respectively.

It's not necessary that the temporal resolution of the NWP forecast is equal to the observation. However, one has to adjust the background inflation factor to >1.0 in this case.

There is also a comprehensive set of additional combination keyword arguments like:
```
combination_kwargs = dict(n_tapering=0,
                              non_precip_mask = True,
                              n_ens_prec=1,
                              lien_criterion=True,
                              n_lien=10,
                              post_prob_matching = False,
                              iterative_prob_matching = True,
                              inflation_factor_bg=1.0,
                              inflation_factor_obs=1.0,
                              offset_bg=0.0,
                              offset_obs=0.0,
                              nwp_hres_eff=14.0,
                              sampling_prob_source="ensemble",)
```

Currently, only an AR(1) process within the forecast step is implemented and the respective tests for the method are not implemented.


RubenImhoff:
Fantastic, Martin! I will have a look at it tomorrow (already gave it a go earlier, of course, but I'll make it a more thorough review now).

pysteps/blending/ens_kalman_filter_methods.py
```diff
@@ -0,0 +1,546 @@
...
+n_ens_prec: int, (1)
+    Minimum number of ensemble members that forecast precipitation for the above
+    mentioned mask.
+lien_criterion: bool, (True)
+    Flag to specify whether Lien criterion (Lien et al., 2013) should be applied for
+    the computation of the update step within the ensemble Kalman filter.
+n_lien: int, (n_ens_members/2)
+    Minimum number of ensemble members that forecast precipitation for the Lien
+    criterion. As standard, half of the ensemble members should forecast precipitation.
+post_prob_matching: bool, (False)
```
RubenImhoff: This one is currently not used from what I can see. Should we delete it?

pysteps/blending/ens_kalman_filter_methods.py
```diff
@@ -0,0 +1,546 @@
...
+    Flag to specify whether a probability matching should be applied at the end of the
+    forecast computation.
+iterative_prob_matching: bool, (True)
+    Flag to specify whether a probability matching should be applied at each correction
+    step.
+inflation_factor_bg: float, (1.0)
```
RubenImhoff, line 33: Could you add a few lines here describing what the inflation factor is used for?

pysteps/blending/ens_kalman_filter_methods.py
```diff
@@ -0,0 +1,546 @@
...
+    Inflation factor of the background (NWC) covariance matrix.
+inflation_factor_obs: float, (1.0)
+    Inflation factor of the observation (NWP) covariance matrix.
+offset_bg: float, (0.0)
```
RubenImhoff, line 41: Could you add a few lines here describing what the offset is used for?

pysteps/blending/ens_kalman_filter_methods.py
```diff
@@ -0,0 +1,546 @@
...
+    Offset of the background (NWC) covariance matrix.
+offset_obs: float, (0.0)
+    Offset of the observation (NWP) covariance matrix.
+nwp_hres_eff: float
```
RubenImhoff, line 49: This one is currently not used from what I can see. Should we delete it?

pysteps/blending/ens_kalman_filter_methods.py
```diff
@@ -0,0 +1,538 @@
...
+        # Estimate the Kalman gain
+        self.K = np.dot(P, np.linalg.inv(P + R))
+
+        # Update the background ensemble
+        X_ana = X_bg.T + np.dot(self.K, (Y_obs - X_bg).T)
+
+        return X_ana
+
+    def get_covariance_matrix(self, M: np.ndarray, inflation_factor: float, offset: float):
+        """
```
RubenImhoff, line 199: In this function, can you refer to the equation(s) from Nerini et al. (2019) that are used?

pysteps/blending/ens_kalman_filter_methods.py
m-rempel: I've merged the argumentes `post_prob_matching` and `iter_prob_matching` into a new one named `prob_matching`.

pysteps/blending/ens_kalman_filter_methods.py
m-rempel: Added.

pysteps/blending/ens_kalman_filter_methods.py
m-rempel: Done.

pysteps/blending/ens_kalman_filter_methods.py
m-rempel: `nwp_hres_eff` is now used in `ForecastModel` to simulate the standard deviation of the smaller scales.

pysteps/blending/ens_kalman_filter_methods.py
m-rempel: Done.

RubenImhoff:
To do:

- [x] Refactor code.
- [x] Ensure that less energy/power is lost at the smallest scales.
- [x] Fix no-rain instances.
- [x] Ensure that we end with full NWP weight.
- [ ] Add tests.
- [ ] Test with real-world data.
- [ ] Create gallery example.

pysteps/blending/pca_ens_kalman_filter.py
```diff
@@ -0,0 +1,1584 @@
...
+        #. Extrapolate the recomposed rainfall field to the current timestep.
+
+        Correction Step
+        ===============
+        #. Get grid boxes where rainfall is forecast.
+        #. Reduce Nowcast and NWP ensemble onto these grid boxes and apply principal
+           component analysis to further reduce the dimensionality.
+        #. Apply update step of ensemble Kalman filter.
+        #. (optional) Apply probability matching.
+
```
RubenImhoff, line 37: Some things in the order changed here (advection/extrapolation happens later, no?), so good to change that here, too. Maybe we should explicitly mention here that this is where this code is different from Nerini et al. (2019) and why we chose to do so.

m-rempel: Done.

pysteps/blending/pca_ens_kalman_filter.py
```diff
@@ -0,0 +1,1628 @@
...
+                            num_workers=self.__params.num_ensemble_workers,
+                        )
+                    else:
+                        for j in range(self.__config.n_ens_members):
+                            worker(j)
+
+                    dask_worker_collection = None
+
+            # Apply back transformation
+            [FC_Model.backtransform() for FC_Model in self.FC_Models.values()]
```
RubenImhoff: @m-rempel, is this line used at all?

RubenImhoff:
@m-rempel, the test still fail because the `sklearn` package is not (yet) part of pysteps. We may have to add it to the package

pysteps/blending/pca_ens_kalman_filter.py
m-rempel: Yep, it is. Currently, it sets the no data values back into the output field.

RubenImhoff:
@m-rempel, the PR still complains about two merge conflicts. Can you merge the master branch from pysteps into your branch? The merge conflict originating from `examples/blended_forecast.py` originates from our change to a different naming for this file (because we will add a new blended forecast example in the gallery, too). So, you can just neglect it and delete `examples/blended_forecast.py`. The file `doc/source/references.bib` has been updated with a new reference in the last commit and so has it in this PR, so it is probably a matter of adding both (github may have difficulties with that in .bib files).

pysteps/utils/pca.py
```diff
@@ -0,0 +1,166 @@
...
+
+.. autosummary::
+    :toctree: ../generated/
+
+    pca_transform
+    pca_backtransform
+"""
+
+import numpy as np
+from sklearn import decomposition
```
dnerini: I think this should be handled as an an optional dependency (and hence documented as such), see other examples in the code

m-rempel: Thanks @dnerini, I've implemented it in this manner.

m-rempel: Hi again @dnerini,
the tests are still failing, since the PCA from sklearn is necessary for the correction step. I've talked with @RubenImhoff about this issue and he has also no idea how to handle this.
One option could be to compute a pure extrapolation nowcast instead of a combination when there's no sklearn installed. However, that makes no really sense for me.

dnerini: We can install sklearn for the tests, but it shouldn't be added to the base dependencies. This means that users that want to use your method need to install
Sklearn too, otherwise they get an error when they try to use the method

dnerini: See example here https://github.com/pySTEPS/pysteps/blob/master/pysteps/feature/tstorm.py#L117

which is documented in the installation page https://pysteps.readthedocs.io/en/stable/user_guide/install_pysteps.html

dnerini: Test dependencies are specified here

https://github.com/pySTEPS/pysteps/blob/master/ci/ci_test_env.yml

so you can add sklearn there

m-rempel:
Even after adding the test of the PCA, codacy issues that pca is import in pysteps/utils/__init__.py but never used. Do you have any ideas what the reason could be for this, @dnerini and @RubenImhoff?

RubenImhoff:
> Even after adding the test of the PCA, codacy issues that pca is import in pysteps/utils/**init**.py but never used. Do you have any ideas what the reason could be for this, @dnerini and @RubenImhoff?

It gives the same issue for the `reprojection` module. I'm also not entirely sure why. @dnerini, do you know?

dnerini:
We can safely ignore those, they shouldn't appear in the next iterations

m-rempel:
Ok, thank you @dnerini,
I've fixed another bug in the PCA test and now it should be working again.

RubenImhoff:
I adjusted the gallery example. Due to the very short NWP forecast horizon in `pysteps_data`, I have adjusted the time step artificially to 15 minutes, so that we can make a blend up to 120 minutes ahead. It works, but is maybe not the prettiest example this way.

RubenImhoff:
@m-rempel, looks like we still have some failing pca tests :(

RubenImhoff:
Not the prettiest plot, nor the best example of the reduced-space ensemble Kalman filter blending, but at least this is a sign that it also work on the KNMI ensemble data (30 members). On my laptop, I could only test it with hourly NWP data, so that explains the mismatches so far. I'll try it out with 5-min data next week. :) 
Anyway, for now a sign that the model runs and finishes successfully.

Test_KNMI_data_blending

m-rempel:
Thank you @RubenImhoff  for sharing these first results with other than DWD data! I've now fixed the bug in the PCA test for cases when n_components < n_ens_member.

RubenImhoff:
Test_KNMI_data_blending_new

A new figure with KNMI data, this time using 5-min data and that looks much much better! :) 

I have used an inflation factor of 1.3 now for the radar data and get after 140 minutes full NWP in this example. Looks good to me!

RubenImhoff (approved):

RubenImhoff:
I think we are ready to go - @dnerini, do you agree as well?

PR #517, opened 2025-09-11
removed failing plugin test. Cookiecutter does not work correctly with xarray

larsbonnefoy:
Cookie cutter makes two calls to importers, one which is not wrapped with postprocess_import resulting in the importer
returning 3 values on the first call. On the second call the importer goes through the postprocess_import resulting in one dataset being returned. Should fix this issue in the cookie cutter before fixing tests. Test has been therefore commented out. Fails on line 73 in decorators.py


sidekock (commented):
Good comments to explain the issue

PR #522, opened 2025-09-14
xarray in blending code

mats-knmi:
This PR implements xarray into the blending code and fixes the associated tests

### DIFF
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -42,6 +42,7 @@ consists of the following main steps:
     calculate_weights_spn
     blend_means_sigmas
 """
+from datetime import datetime
 import math
 import time
 from copy import copy, deepcopy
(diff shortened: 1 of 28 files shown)


larsbonnefoy (commented):
Looks fine for me, many changes at once make it a bit difficult to follow along. 
I've seen that you touched the "transform" values in the meta data.
Have you moved that information elsewhere ?

mats-knmi:
> Looks fine for me, many changes at once make it a bit difficult to follow along. I've seen that you touched the "transform" values in the meta data. Have you moved that information elsewhere ?

Netcdf doesn't allow you to write attributes with None values, so I removed the transform attribute when it was set to None. The checks for this attribute are now changed to `if "transfom" is in attrs` in stead of `if attrs["transform"] is None`. So functionally nothing changed.

sidekock:
Is this pull request still active or do we have a more updated one with the one I just commented on?

mats-knmi:
@sidekock The PR you just commented on contains all the changes in this PR and more. I tried to split up these changes in 2 PRs, but the other pull request extends on the changes in this one, so that contains all these changes plus the extra changes. Once we merge this PR, the other one will be a lot clearer since then that one will only contain the extra changes.

But since you reviewed the other one already, you don't have to also look at  this one.

PR #524, opened 2025-09-15
fix the remaining issues with the tests

mats-knmi:
I removed the importer decorator as that only caused confusion most of the time, while it didn't really add anything. Also some tests failed again because of changes I made in the blending branch. This branch is based on the blending branch so it should be reviewed after https://github.com/pySTEPS/pysteps/pull/522 is merged


pysteps/blending/linear_blending.py
```diff
@@ -105,31 +106,27 @@ def forecast(
     if nowcast_kwargs is None:
         nowcast_kwargs = dict()
 
-    # Ensure that only the most recent precip timestep is used
```
sidekock: Does this not also have to happen in the xarray version?

pysteps/cascade/decomposition.py
```diff
@@ -286,7 +286,7 @@ def recompose_fft(decomp, **kwargs):
     ):
         result = np.sum(levels, axis=0)
     else:
-        if decomp["compact_output"]:
+        if decomp["domain"] == "spectral" and decomp["compact_output"]:
```
sidekock, line 289: Why this new addition compared to the functionality that was there before?

pysteps/extrapolation/eulerian_persistence.py
```diff
@@ -0,0 +1,51 @@
+import numpy as np
```
sidekock, line 1: Was this part not in pysteps before?

pysteps/io/interface.py
```diff
@@ -14,7 +14,6 @@
 """
 from importlib.metadata import entry_points
 
-from pysteps.decorators import postprocess_import
```
sidekock, line 17: I don't realy see why this is no longer used?

pysteps/tests/test_blending_steps.py
```diff
@@ -159,13 +161,21 @@ def test_steps_blending(
 
     metadata = dict()
     metadata["unit"] = "mm"
-    metadata["transformation"] = "dB"
```
sidekock: Don't we need to keep track of this?

sidekock (commented):
Hi Mats,
I wrote some follow ups, no glaring issues but just some things I wanted to double check with you

pysteps/tests/test_blending_steps.py
mats-knmi: The transformation will be added to the attrs of the dataset once we perform the transformation (line 219 below), so it does not have to (should not) be in the metadata of the dataset before this transformation is applied.

pysteps/cascade/decomposition.py
mats-knmi: I am not sure why this was not in there before, but this is a bug. This is because above this code (on line 255) you can see that the weight_masks only gets set if the decomp domain is spectral and compact output is true. So if you don't check for both here you will get into the situation where it tries to read the weight masks while it is not set.

pysteps/extrapolation/eulerian_persistence.py
mats-knmi: I put this extrapolation method in its own file. It used to be a method in the interface.py, but I had some issues with circular imports and tried to solve it like this. I also think that it is more logical that each method has its own file.

pysteps/io/interface.py
mats-knmi: I removed the postprocess_import decorator, because this gave weird issues with the cookiecutter and also issues where I had to create a private method that I wrapped and then called from a public one. Also this decorator didn't really do too much, so I just applied what it did directly in the importers themselves.

pysteps/blending/linear_blending.py
mats-knmi: I moved this part to the extrapolation nowcaster itself. That nowcast code now just extracts the latest precip timestamp from the dataset it is given in stead of the user having to do this. This means that this nowcaster can now be called exactly the same as all the others and you don't have to extract the latest frame manually

sidekock:
Seems good to me. I think we can go ahead an merge with xarray/main

mats-knmi:
@larsbonnefoy I have now merged the other PR. So you can take a look at this one if you still want to. @sidekock has already reviewed this one.

PR #531, opened 2025-12-03
Add Precision-Recall curve in probscores (PR_curve)

DANIELV95:
Added Precision-Recall curve in probscores. Similar to ROC curve but better metric for imbalanced distribution.

### DIFF
--- a/pysteps/verification/probscores.py
+++ b/pysteps/verification/probscores.py
@@ -429,28 +429,29 @@ def ROC_curve_compute(ROC, compute_area=False):
 
 def PR_curve(P_f, X_o, X_min, n_prob_thrs=10, compute_area=False):
     """
-    Compute the Precision-Recall curve, its area, and optionally the prevalence baseline.
+    Compute the Precision–Recall (PR) curve and optionally its area.
 
     Parameters
     ----------
-    P_f: array_like
-      Forecasted probabilities for exceeding the threshold specified in the PR object.
-      Non-finite values are ignored.
-    X_o: array_like
-      Observed values. Non-finite values are ignored.
-    X_min: float
-      Precipitation intensity threshold for yes/no prediction.
-    n_prob_thrs: int
-      The number of probability thresholds to use.
-      The interval [0,1] is divided into n_prob_thrs evenly spaced values.
-    compute_area: bool, optional
-      If True, compute the area under the Precision-Recall curve.
-    
+    P_f : array_like
+        Forecasted probabilities for exceeding the threshold X_min.
+        Non-finite values are ignored.
+    X_o : array_like
+        Observed values. Non-finite values are ignored.
+    X_min : float
+        Precipitation intensity threshold for yes/no prediction.
+    n_prob_thrs : int, optional
+        Number of probability thresholds to evaluate.
+        The interval [0, 1] is divided into n_prob_thrs evenly spaced values.
+    compute_area : bool, optional
+        If True, compute the area under the PR curve using trapezoidal integration.
+
     Returns
     -------
-    out: tuple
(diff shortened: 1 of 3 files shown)


pysteps/verification/probscores.py
```diff
@@ -421,3 +425,142 @@ def ROC_curve_compute(ROC, compute_area=False):
...
+    for i, p in enumerate(PR["prob_thrs"]):
+        forecast_yes = P_f >= p
+        obs_yes = X_o >= PR["X_min"]
+        PR["hits"][i] += np.sum(np.logical_and(forecast_yes, obs_yes))
+        PR["misses"][i] += np.sum(np.logical_and(~forecast_yes, obs_yes))
+        PR["false_alarms"][i] += np.sum(np.logical_and(forecast_yes, ~obs_yes))
+        PR["corr_neg"][i] += np.sum(np.logical_and(~forecast_yes, ~obs_yes))
+
+
+def PR_curve_compute(PR, X_o, X_min, compute_area=False):
```
dnerini: `PR_curve_compute`does not use `X_o` or `X_min` anywhere, maybe a copy/paste error? please remove them

pysteps/verification/probscores.py
```diff
@@ -421,3 +425,142 @@ def ROC_curve_compute(ROC, compute_area=False):
...
+    precision_vals = []
+    recall_vals = []
+
+    for i in range(len(PR["prob_thrs"])):
+        hits = PR["hits"][i]
+        misses = PR["misses"][i]
+        false_alarms = PR["false_alarms"][i]
+
+        recall = hits / (hits + misses) if (hits + misses) > 0 else 0.0
+        precision = hits / (hits + false_alarms) if (hits + false_alarms) > 0 else 1.0
```
dnerini, line 562: the convention using `if (hits + false_alarms) > 0 else 1.0` should be documented in the docstrings as it can lead to some surprising results when thresholds yield zero predicted positives

pysteps/verification/probscores.py
```diff
@@ -421,3 +425,142 @@ def ROC_curve_compute(ROC, compute_area=False):
...
+    -----
+    Updates the counters (hits, misses, false alarms, correct negatives)
+    for each probability threshold in PR["prob_thrs"].
+    """
+    mask = np.logical_and(np.isfinite(P_f), np.isfinite(X_o))
+    P_f = P_f[mask]
+    X_o = X_o[mask]
+    for i, p in enumerate(PR["prob_thrs"]):
+        forecast_yes = P_f >= p
+        obs_yes = X_o >= PR["X_min"]
```
dnerini: this line computing `obs_yes` is constant and can be moved outside of the loop

dnerini (changes requested):
hi @DANIELV95 , thanks for this useful addition (PR curves are often more informative than ROC curves for rare-event verification). The structure (init/accum/compute) matches the ROC approach, which is good for API consistency.

This said, there are a few details worth correcting before merge:

- Missing visualization: I suggest that this PR be complemented with the relative plotting function `plot_PR` in https://github.com/pySTEPS/pysteps/blob/master/pysteps/verification/plots.py (similar to `plot_ROC`).
- Missing converage: we need basic units test along side the new method, see examples in https://github.com/pySTEPS/pysteps/blob/master/pysteps/tests/test_verification_probscores.py

pysteps/verification/probscores.py
dnerini: could you comment on how this is handled in scikit-learn for example?

DANIELV95:
Hello @dnerini, thanks for the comments, I think it should be better now, let me know if there is anything missing

PR #532, opened 2025-12-03
Motion field updated every timestep

DANIELV95:
Added function to update the motion field or to be able to input an array of "real" motion field for every timestep, so instead of having a constant motion field during the nowcast, it can be updated and time-varying.

I haven't tested or evaluated if perfomance of nowcast improve after updating motion field instead of a constant motion but I think that it is better to have the possibility to use a variable motion field as input.


dnerini:
hi @DANIELV95, thanks for proposing this new feature! Could you please elaborate on the use cases for this functionality? Do you have results showing its benefit from your work?

PR #533, opened 2025-12-11
fix: Advection of domain mask for smooth blending at domain edges in reduced-space EnKF

m-rempel:
Hi @RubenImhoff and @mats-knmi,

unfortunately, there were still two bugs regarding the advection of the domain mask as well as in the use of the correct timestep for NWP data when it comes to the full utilization of this data.

Now, the domain mask is advected only once per timestep and the correct timestep for the NWP ensemble is used. In addition, there is now also a switch to further return the background ensemble that is used in the correction step to enable further statistics regarding the EnKF.

### DIFF
--- a/pysteps/blending/pca_ens_kalman_filter.py
+++ b/pysteps/blending/pca_ens_kalman_filter.py
@@ -176,6 +176,8 @@ class EnKFCombinationConfig:
         :py:mod:`pysteps.blending.ens_kalman_filter_methods`.
     measure_time: bool
         If set to True, measure, print and return the computation time.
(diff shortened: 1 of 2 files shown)


RubenImhoff (approved):
Hi @m-rempel, sorry it took so long, but this looks good! Thanks a lot for finding and solving this bug. :)

dnerini:
@RubenImhoff do you want to merge this into main and make a patch release?

RubenImhoff:
Will do! The patch release will follow some time next week, I think, but I can merge this one right away. :)

PR #535, opened 2026-01-15
Return xsize and ysize metadata from the ODIMhdf5 reader

jorahu:
ODIMhdf5 reader has been missing xsize and ysize in the importing stage. I know it could be easily calculated from the x and y coordinates and the xpixelsize and ypixelsize, but why add the calculation when the variables are present already in the metadata.

### DIFF
--- a/pysteps/io/importers.py
+++ b/pysteps/io/importers.py
@@ -37,6 +37,10 @@ The metadata dictionary contains the following recommended key-value pairs:
 +------------------+----------------------------------------------------------+
 |   ypixelsize     | grid resolution in y-direction                           |
 +------------------+----------------------------------------------------------+
+|    xsize         | grid size in x-direction                                 |
++------------------+----------------------------------------------------------+
+|    ysize         | grid size in y-direction                                 |
++------------------+----------------------------------------------------------+
 |   cartesian_unit | the physical unit of the cartesian x- and y-coordinates: |
 |                  | e.g. 'm' or 'km'                                         |
 +------------------+----------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


pysteps/io/importers.py
```diff
@@ -37,6 +37,10 @@
 +------------------+----------------------------------------------------------+
 |   ypixelsize     | grid resolution in y-direction                           |
 +------------------+----------------------------------------------------------+
+|    xsize         | grid size in x-direction                                 |
++------------------+----------------------------------------------------------+
+|    ysize         | grid size in y-direction                                 |
++------------------+----------------------------------------------------------+
```
dnerini, line 43: Including these would imply providing these two metedata for all available readers

dnerini (commented):
hi @jorahu, many thanks. My main concern is about keeping the output consistent across all readers. If you think adding the xsize and ysize metadata would be valuable for the users, then why not doing it for all readers? As you say, these are easily computed on the fly from the other metadata...

pysteps/io/importers.py
jorahu: Hey. I now see your point more clearly. First, apologies for this incomplete idea, this is the first time I am using the pull request. This idea might have been better off as a suggestion under Issues probably. Sorry for that. As I am only using the ODIM importer, where this xsize and ysize is a mandatory variable, I thought it's easier to read that parameter from metadata (as it must be there anyway) than to calculate it. But sure, it might not be the case for all the other importer options that pysteps supports. I apparently only looked it from my narrow perspective. As I don't have good knowledge of all the other importers I am certainly not able to fix them all to proceed with this. Probably better to discard this idea at this point as the parameters can be found/calculated from the data itself.

PR #539, opened 2026-02-26
Fix pandas-compatible scalar assignment in tdating

dnerini:
This PR fixes `pysteps.tracking.tdating` DataFrame assignments to use scalar-safe `.at[...]` when storing array/list-like values (prevents pandas length-mismatch errors in `output_splits_merges=True` multistep tracking).

additionally:
- Fix Python 3.13 `SyntaxWarning: invalid escape sequence` warnings in docstrings/labels (`nowcasts.interface`, `timeseries.correlation`, `fit_vel_pert_params.py`).
- Reduce IO-related warning noise by:
  - explicitly setting GDAL exception mode (`DontUseExceptions`) in IO importer/exporter modules
  - avoiding NumPy 2.x deprecation warnings in SAF CRRI netCDF reads (`np.asarray(...[:])`).


RubenImhoff (approved):

RubenImhoff:
Fantastic, many thanks for this, @dnerini! Looks like the new pandas version gave some breaking changes.. Happy that this is resolved.

PR #544, opened 2026-03-27
Support for mixed precip/noprecip input and AR order 1

### LINKED ISSUE #543: Precip + no-precip input time steps cause blending failure AND in general ar_order=1 causes blending failure
We noticed that https://github.com/pySTEPS/pysteps/blob/master/pysteps/blending/steps.py fails if your radar input consists of time steps with precipitation and time steps without precip. That is because we cannot calculate the correlation coefficient with 0 variance.

While trying to fix this issues, we also noticed that the blending fails for ar_order=1. This is a more general issue.
There are no tests for ar_order=1, too.

ladc:
Fix several issues in case of intermittent rainfall in the radar images:
- [x] Fix mixed rain/norain input frames
- [x] Add tests
- [x] Make AR1 work for regular steps
- [x] Add tests
- [x] Make AR1 work for steps blending
- [x] Add tests

Works towards fixing issue #543.


FelixE91:
We consider these fixes important enough to merge them into the master branch. They can have a direct positive impact on the operation of pysteps, as they
(i) prevent some runs from crashing when mixed time steps with and without precipitation are present in the radar inputs
(ii) provide more readable and clearer error messages if the run does abort due to mixed time steps with and without precipitation in the radar inputs
The tests (pytest) are still generating numerous warnings. Most of them are unrelated to the code we added here.
The warnings from the pytest execution should be investigated as a separate issue #545

FelixE91:
The implemented solution prevents pysteps from crashing when the radar input data contains both precip and noprecip time steps.
If possible, the most recent time steps with precipitation are used, and the ar_order value is adjusted (i.e. reduced) as needed.
Otherwise, the radar input is set to an observation array without precipitation and a warning is printed. Pysteps with NWP blending will still run if rain is present in the NWP.

pysteps/blending/steps.py
```diff
@@ -2228,9 +2253,12 @@ def __regress_extrapolation_and_noise_cascades(self, j, worker_state, t):
...
                     # Renormalize the cascade
-                    worker_state.precip_cascades[j][i][1] /= np.std(
-                        worker_state.precip_cascades[j][i][1]
-                    )
+                    for nl, obs_cascade_level in enumerate(
+                        worker_state.precip_cascades[j][i]
+                    ):
+                        worker_state.precip_cascades[j][i][nl] /= np.std(
+                            obs_cascade_level
+                        )
```
mats-knmi, line 2261: Was this just an unrelated bug that has been present since longer? Seems weird to only renormalize the 1st timestep? Does anyone know why this was initially done? (@ladc @RubenImhoff?) Also does it need to be done for all timesteps here or only for the timestep `t` this method is currently running on?

pysteps/blending/steps.py
```diff
@@ -3686,6 +3719,81 @@ def forecast(
...
+      Adapted to match with precip.shape equal (ar_order+1,m,n).
+    """
+    # Quick check radar input - at least 2 time steps
+    if not precip.shape[0] >= 2:
+        raise ValueError(
+            "Wrong precip shape. The radar input must have at least 2 time steps."
+        )
+
+    # Check all time steps for zero-precip (constant field, minimum==maximum)
+    zero_precip = [np.nanmin(obs) == np.nanmax(obs) for obs in precip]
```
mats-knmi: Why do you not use the `check_norain` method here?

pysteps/nowcasts/steps.py
```diff
@@ -1534,3 +1547,77 @@ def forecast(
     nowcaster.reset_states_and_params()
     # Call the appropriate methods within the class
     return forecast_steps_nowcast
+
+
+# TODO: Where does this piece of code best fit: in utils or inside the class?
```
mats-knmi: If this is exactly the same method as the one in the steps blending, then I think it is best to move it to utils, to avoid code duplication.

mats-knmi:
We have been having these issues as well in our processing, so I hope this PR finally fixes it. To me it looks good, I just  had a couple of questions. If they can be answered I can approve it.

pysteps/blending/steps.py
FelixE91: The output differs. While `check_norain` returns a Boolean value for the entire array, the check here is performed for each individual timestep and returns a list of length `ar_order+1`. This is a simple technical check to verify whether the input array is constant for the respective timestep, which must be avoided in order to calculate the spatial correlation coefficient.
Note that `check_norain` is still used later in steps.py to apply the specified thresholds.

FelixE91: If you prefer, we could call this `const_precip` or so to avoid confusion?

pysteps/nowcasts/steps.py
FelixE91: I agree. I will move it to utils.

FelixE91 (commented):
I have tried to answer @mats-knmi's questions. Final approval for the updates of the blending and autoregression functions is still pending, most of it to enable AR-1 runs.

pysteps/blending/steps.py
FelixE91: I had to update this so that AR-1 would work. I don't quite understand the hard-coded index `[1]` either.
My take on this:
The code should renormalize the latest `precip_cascade` step. I think it might have been a typo to use `[1]` instead of `[-1]`. For example, the subsequent function `perturb_blend_and_advect_extrapolation_and_noise_to_current_timestep(...)` uses indexing as `precip_cascades[j][i][-1, :]`.
However, for me, it was more logical to loop through the timesteps and renormalize the entire array. In case further processing would cut the `precip_cascades` array, it would still be normalized.

Note that current AR code with Yule-Walker equations was basically hard coded to AR-2 model (`ar_order=2`). The array `precip_cascades` had exactly `ar_order` timesteps (i.e., 2) after running function `multiply_precip_cascade_to_match_ensemble_members(...)`. Hence, the hard-coded index `[1]` always meant the latest timestep and pysteps has been running correctly with AR-2.

RubenImhoff:
Thanks for picking this up, great work! I will go through the changes and leave my review as well. 

Regarding AR_order = 1, indeed this was not yet supported (also mentioned in the code by the way:  "The order of the autoregressive model to use. Must be >= 1."). I am glad to see that we are fixing this now, too. :)

pysteps/utils/check_norain.py
```diff
@@ -49,3 +49,77 @@ def check_norain(precip_arr, precip_thr=None, norain_thr=0.0, win_fun=None):
...
+      The order of the autoregressive model to use. Must be >= 1.
+      Adapted to match with precip.shape equal (ar_order+1,m,n).
+    """
+    if not precip.shape[0] >= 2:
+        raise ValueError(
+            "Wrong precip shape. The radar input must have at least 2 time steps."
+        )
+
+    # Check all time steps for zero-precip/constant field (constant field, minimum==maximum)
+    const_precip = [np.nanmin(obs) == np.nanmax(obs) for obs in precip]
```
RubenImhoff: A lot of 'no_rain' cases consist of a couple of cells containing clutter. Our experience is that the autoregressive model and/or determining the correlation coefficients then also fail. So, ideally we check for that, too. 

In the `check_norain` function, we have built in a `norain_thr`, which is the minimum fraction of rainy pixels, and a threshold (`precip_thr`). I can imagine that it would be good, and consistent, to use that here, too. If so, the function becomes more similar to the `check_norain` function, so we could even see if an integration of the two functions makes more sense at that point (no preference on my side).

pysteps/blending/steps.py
RubenImhoff: Hmm, I am a bit confused here, too. The renormalization is a more recent addition, but I think your interpretation is right, @FelixE91. This looks good to me now.

RubenImhoff (changes requested):

pysteps/utils/check_norain.py
FelixE91: Ah, gotcha! That makes sense, now I understand why the `check_norain` method is for flexible than just checking the min and max.
As mentioned, it was a technical test here to avoid const arrays as they cause errors in the correlation coefficients.
I will modify the function to use the thresholds and check for a constant field.

RubenImhoff: Great, thanks, @FelixE91!

mats-knmi: Also in `check_norain` we now also have support for a window function. I am not sure how it works in this case, but we have had instances where there was precipitation outside the window function, so it passed a no_rain check, but after the window function was applied it was still a constant field so we got an error.

I am not sure if that is the case here, but imo just using the `check_norain` method here is slightly more robust and it should also cover the case of a constant field that you are checking for.

Do you have a specific reason why using `check_norain` here would be worse than what you are doing now?

FelixE91: I mean, I guess we can assume that a constant field exists if all precip is zero only. It woudl be very unlikely to have the same non-zero value at every pixel ;)
Since I need a list of length `ar_order+1` as output, I can simply do `const_precip = [check_norain(obs) for obs in precip]` that should work and address @RubenImhoff 's comment, too.

pysteps/blending/steps.py
```diff
@@ -3625,6 +3653,20 @@ def forecast(
...
+    # zero-precip/constant field in previous time steps has to be removed
+    # (constant field causes autoregression to fail)
+    precip, ar_order = check_previous_radar_obs(
+        precip,
+        ar_order,
+        {
+            "precip_thr": precip_thr,
+            "norain_thr": norain_thr,
+            # Set default window function to "tukey" to match the default in check_inputs() below
+            "win_fun": "tukey" if noise_kwargs is None else noise_kwargs["win_fun"],
```
FelixE91, line 3666: I was not totally sure whether this check should be for the entire domain or for the window function that is used in later noise generation.
Here I set it to match this window function ("tukey" by default).
If you think it is better to check the full domain at this point of the code, we simply delete lines 3665/3666.

pysteps/nowcasts/steps.py
```diff
@@ -1495,6 +1503,20 @@ def forecast(
...
+    # zero-precip/constant field in previous time steps has to be removed
+    # (constant field causes autoregression to fail)
+    precip, ar_order = check_previous_radar_obs(
+        precip,
+        ar_order,
+        {
+            "precip_thr": precip_thr,
+            "norain_thr": norain_thr,
+            # Set default window function to "tukey" to match the default in check_inputs() below
+            "win_fun": "tukey" if noise_kwargs is None else noise_kwargs["win_fun"],
```
FelixE91, line 1516: Same as in blending.steps.py.
If we want to check the entire domain here, we can delete lines 1515/1516.

RubenImhoff (approved):

pysteps/io/readers.py
```diff
@@ -61,7 +61,13 @@ def read_timeseries(inputfns, importer, **kwargs):
     timestamps = []
     for i, ifn in enumerate(inputfns[0]):
         if ifn is not None:
-            precip_, quality_, _ = importer(ifn, **kwargs)
+            precip_, quality_, meta_ = importer(ifn, **kwargs)
```
FelixE91, line 64: We noticed something slightly related to the no-precip/precip subject. Please let me know if these changes are clear or if you have any questions.
Basically we should check all input timesteps to set the precip_thr value, and not only one timestep as it has been currently.
Are there any other metadata entries that we have to check, too?

ladc: I think this is fine. Tangentially, I think the rain/norain thresholds are not correctly propagated in a number of places in pysteps, but that's a different issue.

PR #546, opened 2026-04-21
fix only zeros when using external nowcast

mats-knmi:
When using the new external nowcast solution I ran into an issue where the output would contain only 0 (or -15 dBZ). I debugged this issue and this change seems to fix it and also should be in line with what the "normal" steps blending does. The output also looks OK.

@Joep1999 Maybe you can also take a look at this? You do not seem to be a member so you cannot approve, but you can leave comments I believe.

### DIFF
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -450,6 +450,15 @@ class StepsBlendingNowcaster:
         # Store inputs
         self.__precip = precip
         self.__precip_nowcast = precip_nowcast
+        if precip_nowcast is not None:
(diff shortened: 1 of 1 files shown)


Joep1999:
Hey Mats, thanks for checking this out! In principle, I think this looks good! I also sometimes had this issue, but I had not yet committed the time to figuring this out. Thanks a lot!

RubenImhoff (approved):

RubenImhoff:
Nice work, @mats-knmi, this looks good to me!

PR #547, opened 2026-05-05
Add pysteps version to netcdf source metadata.

ladc:
Resolve this small todo.


RubenImhoff (approved):
Good addition! Looks good to me.

PR #549, opened 2026-05-11
fix: issue with example code

### LINKED ISSUE #548: Docs build fails in blended forecast example after #546
The Read the Docs build is currently failing when running `examples/steps_blended_forecast.py`.

Build log: https://app.readthedocs.org/projects/pysteps/builds/32583224/

The error is:

```python
IndexError: index 4 is out of bounds for axis 0 with size 1
```

Full traceback points to this line in pysteps/blending/steps.py:

```python
precip_extrapolated[self.__precip_nowcast_nan_mask[j][t]] = np.nan
```

This seems to be a regression introduced after #546. Maybe @mats-knmi you could have a quick look?

mats-knmi:
Fixes #548 

Looks like if you pass it a nowcast with only a single member the nan mask will not have the correct shape at the time it is used. This is because a single member gets repeated to match the number of members expected. Now the same happens for the nan mask.

### DIFF
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -457,15 +457,6 @@ class StepsBlendingNowcaster:
             # but probability matching needs the NaN domain mask to exclude
             # out-of-domain pixels from the CDF matching.
             self.__precip_nowcast_nan_mask = ~np.isfinite(self.__precip_nowcast)
-            if (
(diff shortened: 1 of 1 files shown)


dnerini (approved):
thanks @mats-knmi, I can confirm that the fix solves the regression described in #548

edit: I'm a bit concerned we didn't capture this with our tests, should we include this test case too?

PR #554, opened 2026-06-19
553 blending weights calculation overwrites weights of other ensemble members

### LINKED ISSUE #553: Blending weights calculation overwrites weights of other ensemble members
In the current blending weights implementation, the weights of other ensemble members can be overwritten, nor are weights calculated for each member individually. To fix this, it requires the worker states to explicitly account for the weights per member.

larsbonnefoy:
Previously NWP weights would be overwritten by each worker has they wrote to the same variable storing NWP weights.

The weights are now  computed per ensemble member j inside the worker(j) in  __blended_nowcast_main_loop, because the NWP skill can differ per member. They are kept on a per-member store (weights_per_member[j],  so that members do not overwrite one another. 

These changes passes the previous tests but I don't know what should be expected in this newer case and how to write a test for it. I could rely on AI to write one but would have very low confidence in it.

### DIFF
--- a/pysteps/blending/clim.py
+++ b/pysteps/blending/clim.py
@@ -193,8 +193,13 @@ def calc_clim_skill(
         past_skill = np.array(None)
     # check if there is enough data to compute the climatological skill
     if not past_skill.any():
+        print("WARNING: Past skill file is empty, using default BPS2006 skill")
         return get_default_skill(n_cascade_levels, n_models)
     elif past_skill.shape[0] < window_length:
+        print(
+            f"WARNING: Past skill file has less days ({past_skill.shape[0]})"
+            f"than expected ({window_length}). Using default BPS2006 skill"
+        )
         return get_default_skill(n_cascade_levels, n_models)
     # reduce window if necessary
     else:
--- a/pysteps/blending/clim.py
+++ b/pysteps/blending/clim.py
@@ -197,7 +197,7 @@ def calc_clim_skill(
         return get_default_skill(n_cascade_levels, n_models)
     elif past_skill.shape[0] < window_length:
         print(
-            f"WARNING: Past skill file has less days ({past_skill.shape[0]})"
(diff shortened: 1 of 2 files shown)


RubenImhoff:
Hi @larsbonnefoy, great addition! I will directly have a look at it. If tests are required, I can always add them (although I am on vacation coming week, so it may take a week longer).

pysteps/blending/clim.py
```diff
@@ -193,8 +193,13 @@ def calc_clim_skill(
         past_skill = np.array(None)
     # check if there is enough data to compute the climatological skill
     if not past_skill.any():
+        print("WARNING: Past skill file is empty, using default BPS2006 skill")
         return get_default_skill(n_cascade_levels, n_models)
     elif past_skill.shape[0] < window_length:
+        print(
+            f"WARNING: Past skill file has less days ({past_skill.shape[0]})"
```
RubenImhoff:
```suggestion
            f"WARNING: Past skill file has fewer days ({past_skill.shape[0]})"
```

RubenImhoff (changes requested):
This looks great, @larsbonnefoy, well done!

larsbonnefoy:
> Hi @larsbonnefoy, great addition! I will directly have a look at it. If tests are required, I can always add them (although I am on vacation coming week, so it may take a week longer).

If you have an idea of how we could test this I could also write it ! Otherwise do you have concrete example where this led previously to an issue in a pysteps run ? I could always try a rerun of a past case as our domains overlap. Might be able to spot already some differences.

RubenImhoff:
It seems challenging to test, because the models will run successfully without the fix. However, I notice a 'jumpy' behavior in the temporal consistency of the ensemble members. So not sure yet if defining a test is feasible here.

RubenImhoff (approved):

RubenImhoff:
@larsbonnefoy, I think this is good as is to commit. :)

PR #556, opened 2026-06-23
fix: removed hardcoded conditional=True for noise statistic computation

### LINKED ISSUE #555: Inconsitent noise computation in blending
`blending/steps.py` **hardcodes** `compute_noise_stddev_adjs(conditional=True)`, whereas the standalone `nowcasts/steps.py` uses  `conditional=self.__config.conditional`. Passing `conditional=False` to `steps.forecast()` means that the noise coefficients are sized over the rain area (because conditional is always true), while the field statistics use the whole domain. 

Same conditional=True for Kalman filter implementation but `Object of type EnKFCombinationConfig has no attribute conditional `

larsbonnefoy:
See #555 for description

### DIFF
--- a/pysteps/blending/steps.py
+++ b/pysteps/blending/steps.py
@@ -1437,7 +1437,7 @@ class StepsBlendingNowcaster:
                     self.__params.perturbation_generator,
                     self.__params.noise_generator,
                     20,
-                    conditional=True,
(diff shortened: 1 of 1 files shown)


RubenImhoff (approved):
Great work, this looks good to me!

PR #562, opened 2026-07-22
fix: extend knmi importer with support for other projections

mats-knmi:
A lot of metadata is currently hardcoded in the importer. To allow files in knmi hdf5 format that have different values here, we need to extend the importer.


RubenImhoff (approved):

RubenImhoff:
Hi @mats-knmi, this is looking great! The tests still fail on the assertion of the right x- and y-grid coordinates. Is there still something wrong in the test grid (km instead of m), or is the assertion just too strict (with e-10)?

PR #564, opened 2026-07-29
Fix issue with external ensemble nowcast for steps blending and no additional noise

mats-knmi:
When providing the STEPS blending with an external nowcast that already contains its own noise, you can set the noise method to None. This, however, resulted in the weights not being computed correctly. This PR fixes this and also adds regression tests to make sure this does not happen again.

Additionally this PR updates the docstring to indicate better to users how we recommend this feature to be used.


RubenImhoff (approved):
This looks great! Thanks for this addition, @mats-knmi . :)

RubenImhoff:
Looks like Black is still failing.

PR #566, opened 2026-08-04
Fix external nowcast only zeros

mats-knmi:
I thought I fixed this in https://github.com/pySTEPS/pysteps/pull/546, but I realized now that an external nowcast does not necessarily have to have nan's at the edge (especially in our usecase where we run the external nowcast on a larger domain to explicitly prevent this).

This should not be an issue if it were not for the bug where, if there are no nan's anywhere in the domain, the code that is supposed to fill this part with values from nwp will simply set the entire domain to 0.

The simple fix is to not try to fill in the nan's when there are no nan's anywhere in the domain.


RubenImhoff (approved):
Nice work, looks good to me!

PR #567, opened 2026-08-05
Fix boxplot labels for Matplotlib 3.11

dnerini:
See https://app.readthedocs.org/projects/pysteps/builds/33904697/

Fix the documentation build with matplotlib 3.11, where the deprecated labels argument was removed from `Axes.boxplot`. Set the x-axis labels separately to preserve compatibility with older matplotlib versions.


dnerini:
build successful in https://app.readthedocs.org/projects/pysteps/builds/33920845/

PR #568, opened 2026-08-10
bump version

mats-knmi:


mats-knmi:
@ladc If you can then also approve this one, I can create a new release and then the notebooks will fully work by installing pysteps through pip

RubenImhoff (approved):

PR #569, opened 2026-08-10
fix issue with linear blending

mats-knmi:
The pysteps short courses use the linear blending method, but it seems to be broken. The radar precip should just be passed to the nowcast method as is, since a nowcast method like steps expects all 3 timesteps to be there. If you want to run this with an extrapolation nowcast you should only pass the last timestep, just like you do when you call the extrapolation nowcast directly.

### DIFF
--- a/pysteps/tests/test_blending_linear_blending.py
+++ b/pysteps/tests/test_blending_linear_blending.py
@@ -185,21 +185,21 @@ def test_linear_blending(
     # entirely constant
 
     # Assert that the control time step is in the range of the forecasted time steps
-    assert controltime <= (n_timesteps * timestep), (
-        "Control time needs to be within reach of forecasts, controltime = {} and n_timesteps = {}".format(
-            controltime, n_timesteps
-        )
+    assert controltime <= (
+        n_timesteps * timestep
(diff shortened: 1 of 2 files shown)


pysteps/tests/test_blending_linear_blending.py
```diff
@@ -226,6 +227,10 @@ def test_linear_blending(
     r_input, _ = transformation.dB_transform(
         r_input, None, threshold=0.1, zerovalue=-15.0
     )
+    if len(r_input.shape) == 3:
+        r_input = (
+            r_input[-1] if nowcast_method in ["extrapolation", "eulerian"] else r_input
```
ladc: Are these the only methods requiring only the last frame? Maybe this could be made a bit less brittle, but it works for now :)

ladc (approved):
Looks good to me, I've added some regression tests (generated and checked, but feel free to double-check).

pysteps/tests/test_blending_linear_blending.py
mats-knmi: Actually it seems there are more that require only the last frame, but we don't have test cases for them so this should still be fine. I will add them for completeness here though

PR #572, opened 2026-08-17
Verification fixes

pulkkins:
This branch introduces fixes to the documentation of the **verification** and **utils** modules:
- **utils.spectral.rapsd**: Set `fft_method=np.fft` by default to avoid confusion of the input type. With `fft_method=None` the method previously assumed that the input is already Fourier-transformed, which is not the most common use case.
- **verification.probscores** / ROC curves: Explicitly specify the order of the x- and y-coordinates in the ROC curve object.
- **verification.probscores** / reliability diagrams: Expand the documentation of the methods.
- Miscellaneous fixes to coding and docstring style in the above modules

### DIFF
--- a/pysteps/utils/spectral.py
+++ b/pysteps/utils/spectral.py
@@ -98,37 +98,43 @@ def mean(X, shape):
 
 
 def rapsd(
-    field, fft_method=None, return_freq=False, d=1.0, normalize=False, **fft_kwargs
+    field, fft_method=np.fft, return_freq=False, d=1.0, normalize=False, **fft_kwargs
 ):
     """
-    Compute radially averaged power spectral density (RAPSD) from the given
-    2D input field.
+    Compute radially averaged power spectral density (RAPSD) of the given 2D
+    input field.
 
     Parameters
     ----------
-    field: array_like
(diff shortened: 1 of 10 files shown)


pysteps/utils/spectral.py
```diff
@@ -98,37 +98,43 @@ def mean(X, shape):
 
 
 def rapsd(
-    field, fft_method=None, return_freq=False, d=1.0, normalize=False, **fft_kwargs
+    field, fft_method=np.fft, return_freq=False, d=1.0, normalize=False, **fft_kwargs
```
dnerini, line 101: looks like this change broke some logic elsewhere, see failing tests.

e.g. now [fftgenerators.py#L150](https://github.com/pySTEPS/pysteps/blob/19e5f0df9eb65c1b0a5e03e2b0ec4e9223a945a7/pysteps/noise/fftgenerators.py#L150 )  should include `fft_method=None`, if I'm not wrong?

dnerini (changes requested):
Hi Seppo, thanks a lot for the nice contribution. I found only one issue that is causing the tests to fail.

